PR #10, opened 2023-08-18
initial oracle commit

otan:
This commit contains an initial version of oracle. It contains the bare
minimum for something to be demo'd.

It only works if:
* the schema and table names match case.
* the column names are lowercase in PG.
* The only tested types are integers and strings.

No escaping work has yet been done - but that will come - this makes the
PR a tad easier to read.

Unfortunately, CI w/ Oracle is not cooporating with me at the moment -
I'll tackle that later.

### DIFF
--- a/dbconn/dbconn.go
+++ b/dbconn/dbconn.go
@@ -69,6 +69,8 @@ func Connect(ctx context.Context, preferredID ID, connStr string) (Conn, error) 
 		return ConnectPG(ctx, id, connStr)
 	case strings.Contains(before[0], "mysql"):
 		return ConnectMySQL(ctx, id, before[len(before)-1])
+	case strings.Contains(before[0], "oracle"):
+		return ConnectOracle(ctx, id, connStr)
 	}
 	return nil, errors.Newf("unrecognised scheme %s from %s", before[0], connStr)
 }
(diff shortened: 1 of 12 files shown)


dbconn/dbconn.go
```diff
@@ -110,6 +112,24 @@ func TestOnlyCleanDatabase(ctx context.Context, id ID, url string, dbName string
 		}
 		cfgCopy.DBName = dbName
 		return ConnectMySQL(ctx, c.id, cfgCopy.FormatDSN())
+	case *OracleConn:
+		if _, err := c.ExecContext(ctx, "DROP USER "+dbName+" CASCADE"); err != nil && !strings.Contains(err.Error(), "ORA-01918") {
```
Jeremyyang920, line 116: Should we make the error code a constant? Or will this be the only place we use this.

oracleconv/datatype.go
```diff
@@ -0,0 +1,68 @@
...
+			if dataScale.Valid {
+				sc := dataScale.Int64
+				if sc != 0 {
+					return oid.T_numeric, true
+				}
+			}
+			// Precision can be up to 38, len('9223372036854776000') == 19.
+			// If we go higher, make it an int.
+			if prec >= 19 {
+				return oid.T_numeric, true
```
Jeremyyang920, line 31: Is this the right OID since it needs to be an int? Its different from the Integer case above.

bobvawter:
You might be able to crib some work from cdc-sink to make the case-alignment less awful, should you need to do more than just passing identifiers straight through.

https://github.com/cockroachdb/cdc-sink/pull/417 for a canonicalizing map.

https://github.com/cockroachdb/cdc-sink/pull/418 for a case-insensitive, but case-preserving maps in `internal/util/ident`

otan:
> You might be able to crib some work from cdc-sink to make the case-alignment less awful, should you need to do more than just passing identifiers straight through.

yep, that's next on my TODO list but i'll live with this for this commit

dbconn/dbconn.go
otan: i'm going to say i will do that in the future, but atm this is the only place i expect this (there is no `DROP USER IF EXISTS` in oracle)

oracleconv/datatype.go
otan: `NUMBER` in oracle can be an INT or DECIMAL. if you do `NUMBER(19)`, it's a number >= math.MaxInt64, so `T_numeric` is correct as that can represent things with arbitrary precision a la oracle.

Jeremyyang920 (approved):

PR #16, opened 2023-08-28
Add mysqlurl package

Jeremyyang920:
This commit adds the mysqlurl package that handles parsing of a mysql connection string that will be used both in here as well as the LMS.


mysqlurl/parse.go
```diff
@@ -0,0 +1,427 @@
...
+
+	"github.com/cockroachdb/errors"
+	mysqldriver "github.com/go-sql-driver/mysql"
+)
+
+const (
+	defaultMaxAllowedPacket = 64 << 20
+)
+
+func ParseMySQLDSN(connStr string) (*mysqldriver.Config, error) {
```
otan, line 22: nit: to avoid stuttering in package name, can you rename these functions this `ParseDSN` and `ParseConnStr`?

can you also make a `Parse` function that does the test one but defaults to another?

otan (approved):

PR #17, opened 2023-08-31
bin: introduced the bin directory and added the initial CLI build along with README

ryanluu12345:
This introduces the new molt-lms-cli binary so that it is available for customers to easily access. This is currently distributed as a binary since the source code is currently coupled to other packages in another private repository. The plan here is to decouple the CLI and bring it over to this repository once that work is completed. For now, there will be automation to sync changes to the CLI

Release Note: None

Companion PR: https://github.com/cockroachlabs/crdb-proxy/pull/529

### DIFF
--- a/bin/lms/README.md
+++ b/bin/lms/README.md
@@ -1,0 +1,47 @@
+# Live Migration Service
+
+## CLI
+
+The CLI is the tool that interacts with the Orchestrator to list connections, get status details, and initiate all cutover actions. It supports standard HTTP or HTTPS (provide your own certs) for the client. Overall, the tool simplifies interactions with the Orchestrator/LMS and has clear documentation via help commands for CLI usage.
+
+### To add to path
+
+In your `~/.zshrc`, `~/.bashrc`, or equivalent:
+
+```
+LMS_DIR="<path to your molt repo>/bin/lms"
+export PATH="$PATH:$LMS_DIR"
+```
+
+### To run
+
+Depending on your system's architecture, you'll probably be using a CLI with a different name. Substitute `molt-lms-cli` with the relevant name below to run.
+
+To get help with commands:
+
+```
+molt-lms-cli help
+```
+
+Without TLS:
+
+```
+molt-lms-cli <insert sub commands here>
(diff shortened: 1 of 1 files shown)


bin/lms/README.md
```diff
@@ -0,0 +1,49 @@
+# Live Migration Service
+
+## CLI
+
+The source code of the CLI is built alongside the LMS and the Orchestrator itself. After being built, the binary is moved over to this repository in the `./bin/lms` directory and named `molt-lms-cli`. In the future, we are considering moving the source code to this directory so that it can be built directly from source, once we decouple the CLI from other pieces we are reliant on in another private repository.
+
+There will be an automation piece that will make sure that any new builds of the `molt-lms-cli` in the private repository (`crdb-proxy`) will generate a pull request to update the version in this repository (`molt`).
```
Jeremyyang920, line 7: This section can probably be removed besides the description of what it is. The rest of the details like private repo and stuff can be removed. 

Only thing in here really should be how to use it.

bin/lms/README.md
```diff
@@ -0,0 +1,49 @@
...
+LMS_DIR="<path to your molt repo>/bin/lms"
+export PATH="$PATH:$LMS_DIR"
+```
+
+### To run
+
+To get help with commands:
+
+```
+./molt-lms-cli help
```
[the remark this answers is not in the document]
ryanluu12345: Ok good point. I updated.

bin/lms/README.md
ryanluu12345: Yep, I gave a brief description of the CLI and removed all of the repo description stuff.

otan (changes requested):
i'm going to advise against doing it this way for a few reasons:
* this makes the git repo big - at least 26MB per binary, per platform, per version
* coordinating the repo's versioning and molt lms cli's versioning will be tricky

instead, we should consider uploading versions onto an s3 bucket (or some alternate static mirror) for each platform that's easily downloadable. molt does it using the `artifacts` on each release which is valid too, but that involves making `molt lms` public which i think we want to avoid (for now).

PR #18, opened 2023-09-06
mysqlurl: register TLS config when parsing using connection string

otan:
Previously, when using a connection string to parse and reformatting it back as a DSN, the TLS config is lost as the field is not set nor the TLS config registered to the driver. Resolve this by registering any parsed config.

### DIFF
--- a/mysqlurl/parse.go
+++ b/mysqlurl/parse.go
@@ -1,8 +1,10 @@
 package mysqlurl
 
 import (
+	"crypto/sha1"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/hex"
 	"fmt"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


mysqlurl/parse_test.go
```diff
@@ -254,7 +256,7 @@ var testConnStr = []struct {
 			skipVerify: false,
 			address:    "address:3306",
 		},
-		outConnStr: "mysql://username:password@address:3306/dbname?columnsWithAlias=true&multiStatements=true&param=value&sslmode=verify-full&sslrootcert=.%2Ftestdata%2Frootcert.pem",
+		outConnStr: "mysql://username:password@address:3306/dbname?columnsWithAlias=true&multiStatements=true&param=value&sslmode=verify-full&sslrootcert=.%2Ftestdata%2Frootcert.pem&tls=parsed_d703e4191f4328a95d960e2df1231f7080c80be2",
```
otan, line 259: this is actually slightly broken; we should use `&sslrootcert=x&sslkey=y` etc., hmm

mysqlurl/parse.go
```diff
@@ -160,10 +170,6 @@ func CfgToConnStr(cfg *mysqldriver.Config, queryEscape bool) string {
 		urlValues["timeout"] = []string{cfg.Timeout.String()}
 	}
 
-	if len(cfg.TLSConfig) > 0 {
```
otan, line 163: this is the `RegisterTLSConfig` field, i don't think we should display it.

Jeremyyang920 (approved):

PR #20, opened 2023-09-11
LMS: Add Helm Chart

Jeremyyang920:
This commit adds the production helm chart for the LMS. It will deploy the LMS and Orchestrator by itself with full customization over how you want to configure it. There is an optional service monitor that can be installed if running with the prometheus operator.

### DIFF
--- a/charts/lms/Chart.yaml
+++ b/charts/lms/Chart.yaml
@@ -1,0 +1,7 @@
+apiVersion: v2
+name: molt-lms
+description: Helm chart for MOLT Live Migration Service
+type: application
+version: 0.1.0
+appVersion: 0.1.0
+icon: "https://raw.githubusercontent.com/cockroachdb/cockroach/master/docs/media/cockroach_db.png"
--- a/charts/lms/templates/_helpers.tpl
+++ b/charts/lms/templates/_helpers.tpl
@@ -1,0 +1,62 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "molt-lms.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "molt-lms.fullname" -}}
+{{- if .Values.fullnameOverride }}
+    {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+    {{- $name := default .Chart.Name .Values.nameOverride }}
+    {{- if contains $name .Release.Name }}
+        {{- .Release.Name | trunc 63 | trimSuffix "-" }}
+    {{- else }}
+        {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+    {{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "molt-lms.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "molt-lms.labels" -}}
+helm.sh/chart: {{ include "molt-lms.chart" . }}
+{{ include "molt-lms.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "molt-lms.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "molt-lms.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "molt-lms.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create }}
+    {{- default (include "molt-lms.fullname" .) .Values.serviceAccount.name }}
+{{- else }}
+    {{- default "default" .Values.serviceAccount.name }}
+{{- end }}
+{{- end }}
--- a/charts/lms/templates/lms-role.yaml
+++ b/charts/lms/templates/lms-role.yaml
@@ -5,7 +5,6 @@ metadata:
   namespace: {{ .Release.Namespace | quote }}
   labels:
     {{- include "molt-lms.labels" . | nindent 4 }}
-    release: {{ .Values.lms.releaseName }}
   {{- with .Values.labels }}
     {{- toYaml . | nindent 4 }}
   {{- end }}
@@ -25,7 +24,6 @@ metadata:
   namespace: {{ .Release.Namespace | quote }}
   labels:
     {{- include "molt-lms.labels" . | nindent 4 }}
-    release: {{ .Values.lms.releaseName }}
   {{- with .Values.labels }}
     {{- toYaml . | nindent 4 }}
   {{- end }}
--- a/charts/lms/templates/lms-role.yaml
+++ b/charts/lms/templates/lms-role.yaml
@@ -1,0 +1,39 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: lms-role
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "molt-lms.labels" . | nindent 4 }}
+    release: {{ .Values.lms.releaseName }}
+  {{- with .Values.labels }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+rules:
+  - apiGroups:
+        - ""
+    resources:
+      - endpoints
+    verbs: ["get"]
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: lms-rolebinding
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "molt-lms.labels" . | nindent 4 }}
+    release: {{ .Values.lms.releaseName }}
+  {{- with .Values.labels }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: lms-role 
+subjects:
+- kind: ServiceAccount
+  name: {{ include "molt-lms.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace | quote }}
--- a/charts/lms/templates/orchestrator-deployment.yaml
+++ b/charts/lms/templates/orchestrator-deployment.yaml
@@ -5,7 +5,6 @@ metadata:
   namespace : {{ .Release.namespace | quote }}
   labels:
     {{- include "molt-lms.labels" . | nindent 4 }}
-    release: {{ .Values.orchestrator.releaseName }}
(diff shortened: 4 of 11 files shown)


ryanluu12345:
Going to give it a deeper review this afternoon once I wrap up find/replace hotfixes. But overall question: how are we testing that this properly works? And do we have a serverless cluster and MySQL instance we can connect to to verify this is all working fine?

charts/lms/templates/orchestrator-deployment.yaml
```diff
@@ -0,0 +1,112 @@
...
+              value: {{ .Values.orchestrator.sourceDialect }}
+            - name: ORCH_TARGET_DIALECT
+              value: {{ .Values.orchestrator.targetDialect }}
+            - name: ORCH_ALLOW_ORIGIN
+              value: {{ .Values.orchestrator.allowOrigin }}
+          {{- with .Values.orchestrator.env }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+            {{- if .Values.orchestrator.tlsSecretName }}
+            - name: ORCH_CA_TLS_CERT
```
ryanluu12345, line 62: Do we want to give a commented out template that guardrails customers to use an external secret to populate a secret that will be mounted in as `/app/certs/*`. I think that it will be worthwhile for them to have an example in either YAML or the README, so it pushes them towards safer practices.

charts/lms/templates/orchestrator-deployment.yaml
```diff
@@ -0,0 +1,112 @@
...
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: orch-metrics
+              scheme: HTTP
+          resources:
+            {{- toYaml .Values.orchestrator.resources | nindent 12 }}
+      {{- if or .Values.orchestrator.tlsSecretName .Values.orchestrator.configSecretName }}
+          volumeMounts:
+            {{- if .Values.orchestrator.tlsSecretName }}
```
ryanluu12345, line 87: I guess this makes it so that the user has to put the TLS secrets in a K8s secret in order to mount it here. Yeah this does the enforcing.

ryanluu12345:
Overall, this looks very clean and straightforward to me. Great job on it. Like I mentioned will want to test and evaluate more deeply later.

charts/lms/templates/orchestrator-deployment.yaml
Jeremyyang920: Not in this file but for the general written up read me yes. There is a lot to document since we have so many possible variables. 
I actually need to update this section a bit more still.

Jeremyyang920:
> 

Yes, there is a mysql and serverless cluster im testing against pulled in via external secrets. Ill slack you the manifest to apply.

Jeremyyang920:
> Overall, this looks very clean and straightforward to me. Great job on it. Like I mentioned will want to test and evaluate more deeply later.

You don't have to fully install it to test some of the things. You can just use `helm template` to render the resulting yaml files and view that the right things were rendered.

charts/lms/values.yaml
```diff
@@ -0,0 +1,140 @@
...
+  nodeSelector: {}
+
+orchestrator:
+  releaseName: "molt-lms-orchestrator"
+  sourceDialect: 'mysql'
+  targetDialect: 'cockroach'
+  configSecretName: ""
+  # allowOrigin specifies the pattern or exact host(s) that should be
+  # allowed to make requests from a browser (CORS).
+  allowOrigin: "/.*localhost.*/"
```
otan, line 93: do we want this by default?

charts/lms/values.yaml
```diff
@@ -0,0 +1,140 @@
...
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+  resources: {}
+  nodeSelector: {}
+
+orchestrator:
+  releaseName: "molt-lms-orchestrator"
+  sourceDialect: 'mysql'
```
otan, line 88: can we comment this out / make users explicitly set this?

charts/lms/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+name: molt-lms
+description: Helm chart for MOLT-LMS
```
otan, line 3: nit: `MOLT Live Migration Service`

otan (commented):
looks good to me from a helm newb POV. small q's, will let ryan sign off.

charts/lms/Chart.yaml
```diff
@@ -0,0 +1,7 @@
...
+type: application
+version: 0.1.0
+appVersion: 0.1.0
+icon: "https://upload.wikimedia.org/wikipedia/en/3/31/Cockroach_Labs_Logo.png"
```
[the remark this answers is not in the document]
Jeremyyang920: Fixed! Took the one from the CRDB chart now. Though this icon really isint used anywhere haha.

charts/lms/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+name: molt-lms
+description: Helm chart for MOLT-LMS
+type: application
+version: 0.1.0
```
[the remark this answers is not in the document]
Jeremyyang920: Yes, thats the idea. Was just putting the release version in here so its ready to go already. I've already pushed a temp image with this tag so testing won't break with a image tag not found.

charts/lms/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
+{{/*
+Expand the name of the chart.
```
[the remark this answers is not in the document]
Jeremyyang920: This is auto generated and it kinda just look like this for all .tpl. But I can fix and manually indent the if statements to make it cleaner. But this file isin't usually read much at all.

charts/lms/values.yaml
Jeremyyang920: Will default to @ryanluu12345 for this answer as to what we might want to set this to for default.

charts/lms/templates/_helpers.tpl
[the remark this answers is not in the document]
otan: If it's autogenerated dw. May be worth adding it to git attributes tho so it doesn't get marked for review https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

charts/lms/values.yaml
ryanluu12345: This is really only relevant for browsers on different domains trying to access our API. If people are interacting via CLI, it's not a huge deal. Given that we don't intend to use the Grafana workflow, then I'd probably say not necessary and can leave default to localhost. Also, localhost necessitates that they are making requests from the machine, which I'm not too concerned about.

We should probably leave some guidance here that if they are trying to access the orchestrator from a browser, this value should be a regex for their host name or IP address (client).

charts/lms/templates/orchestrator-deployment.yaml
```diff
@@ -0,0 +1,100 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "molt-lms.fullname" . }}-orchestrator
```
ryanluu12345, line 4: What would this resulting name be? Do we have repetitive elements here anymore? Or is this just going to be `molt-lms-orchestrator`?

charts/lms/values.yaml
```diff
@@ -0,0 +1,140 @@
...
+  #   drop:
+  #   - ALL
+  # readOnlyRootFilesystem: true
+  # runAsNonRoot: true
+  # runAsUser: 1000
+
+lms:
+  replicaCount: 3
+  releaseName: "molt-lms"
+  sourceDialect: ""
```
ryanluu12345, line 38: Do we need to include a target dialect here too?

charts/lms/templates/orchestrator-deployment.yaml
Jeremyyang920: By default its molt-lms-orchestrator, but if the override fullname its whatever full name they choose.

charts/lms/values.yaml
```diff
@@ -0,0 +1,140 @@
...
+  #   cpu: 100m
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+  resources: {}
+  nodeSelector: {}
+
+orchestrator:
+  releaseName: "molt-lms-orchestrator"
```
ryanluu12345, line 87: Why not use this for the `name` for the deployment/service. Why are we using the variable for the `molt-lms` appended with `orchestrator`

charts/lms/values.yaml
```diff
@@ -0,0 +1,140 @@
...
+  sourceDialect: ""
+  targetDialect: 'cockroach'
+  configSecretName: ""
+  # allowOrigin specifies the pattern or exact host(s) that should be
+  # allowed to make requests from a browser (CORS).
+  allowOrigin: "/.*localhost.*/"
+
+  # Used for orchestrator ssl certs.
```
ryanluu12345, line 95: Clean, I like this!

charts/lms/values.yaml
Jeremyyang920: No. This is really only used to set the mode in the deployment. 
The rest should be set via init-target and source-target.

ryanluu12345 (approved):
LGTM! Clean. Just had a few clarifying questions.

charts/lms/values.yaml
Jeremyyang920: Answered above

Jeremyyang920: But i also might remove this from the chart. Don't really need it to do the match labels.

charts/lms/templates/_helpers.tpl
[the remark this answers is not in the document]
otan: as in, it's fine, but id rather tag autogenerated docs with `.gitattributes` `linguist-generated` so it doesn't show up on reviews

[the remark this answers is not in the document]
Jeremyyang920: Yea I know what you mean. Thing is you can also add extra template helpers to this file down the line if we need it for something so I dont want to fully ignore it.

charts/lms/values.yaml
```diff
@@ -0,0 +1,137 @@
...
+lms:
+  replicaCount: 3
+  sourceDialect: ""
+  shadowMode: none
+
+  # Name of the secret containing the JSON format of 
+  # environment variables.
+  configSecretName: ""
+
+  # Used for mounting MySQL and/or LMS server side certs.
```
ryanluu12345, line 44: Let's explicitly recommend using SSL for this and the orchestrator.

charts/lms/templates/lms-role.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
ryanluu12345, line 1: Typically want to isolate different services into their own namespace. In this case since it is tightly coupled, this is fine - @mtjabz

charts/lms/values.yaml
```diff
@@ -0,0 +1,137 @@
...
+  # where the value is the path to the file such as '/app/certs/lms-cert.pem'.
+  # sslVolumes:
+  #   - name: mysqlca
+  #     secret:
+  #       secretName: mysqlca
+  # sslVolumeMounts:
+  #   - mountPath: "/app/certs"
+  #     name: mysqlca
+  #     readOnly: true
+  sslVolumes: {}
```
ryanluu12345, line 58: @Jeremyyang920 I think we may also need SSL certs for the target DB too right? For example in the case where we connect to serverless, which would need certs.

charts/lms/values.yaml
```diff
@@ -0,0 +1,137 @@
...
+  sslVolumeMounts: {}
+
+  env: []
+  labels: {}
+  annotations: {}
+
+  service:
```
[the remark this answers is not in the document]
Jeremyyang920: No, because I think becomes too much extra stuff. People that already have it should already have it.

charts/lms/values.yaml
Jeremyyang920: Ah yea... We're gonna need to make a code change to allow for that crdb cert.

Jeremyyang920: Actually need it for both source and target if you are running in crdb mode....

PR #25, opened 2023-09-19
.github/workflows: productionize Helm chart publishing

ryanluu12345:
- Add environment references so that bucket names and URLs can be conditional for environment
- Add in basic versioning
- Pull old index and merge with existing
- Upload to bucket after authing with Google

Release Note: None

### DIFF
--- a/.github/workflows/release-helm.yaml
+++ b/.github/workflows/release-helm.yaml
@@ -7,12 +7,16 @@ on:
 
 jobs:
   publish-chart:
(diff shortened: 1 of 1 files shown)


otan (approved):

rail (changes requested):
Can you add the following block to allow the job to use tokens:
https://github.com/cockroachlabs/crdb-proxy/blob/137cb7bc6b49379c25fba88c66c176edb4ad751b/.github/workflows/release.yaml#L19-L21

rail (approved):

PR #27, opened 2023-09-22
.github/workflows: updated so that we remove potential erroneous index file

ryanluu12345:
Test on empty index.yaml: https://github.com/cockroachdb/molt/actions/runs/6279858063
Test on existing index.yaml: https://github.com/cockroachdb/molt/actions/runs/6279864587/job/17056238231


sjbarag (approved):
Thanks for the comment!

Issue #14, opened 2023-08-24
fetch: easy performance improvements when uploading to cloud stores


otan:
We currently are missing the following performance recommendations for fetch which I think would be useful:

* [x] We currently only break files down after it reaches a certain size. We should also give the option to "split" files by number of rows as well. That would be around here: https://github.com/cockroachdb/molt/blob/53f0b01025f56c8060620fb2d2c6291b47df2b5f/fetch/csv_pipe.go#L72
* [x] We should zip files whilst piping to the data store by writing the file through a `gzip` Writer object in `CSVPipe`: https://pkg.go.dev/compress/gzip - this should be optional (--compression gzip/bzip/none), defaulting to gzip. When calling `IMPORT` on CockroachDB, use `decompress` as well: https://www.cockroachlabs.com/docs/stable/import#import-options

ryanluu12345:
Ideas for breaking down files after num rows: 
- Keep track of the current rows in the csvPipe struct: `curNumRows`. If this number exceeds a certain number, we flush.
- We should add a flag like: `--flush-after-num-rows`

Ideas for gzip:
- Write the CSV items to a separate gzip or compressed writer
- Use that writer to write the record instead
- Update the import to use the `decompress` option
- These are both turned on conditionally based on the flag being passed in

ryanluu12345:
PR to address item 1: https://github.com/cockroachdb/molt/pull/31/files

ryanluu12345:
@otan agreed. It was auto-closed because I referenced the issue number in the PR.

PR #31, opened 2023-09-25
fetch: added in the config flag/option for configuring a number of rows to flush after

ryanluu12345:
Added in an optimization parameter so that users can control the number of rows after for the exportTable job to flush.

Resolves: #14
Release Note: None

### DIFF
--- a/fetch/csv_pipe.go
+++ b/fetch/csv_pipe.go
@@ -18,7 +18,9 @@ type csvPipe struct {
 	logger    zerolog.Logger
 
 	flushSize int
+	flushRows int
 	currSize  int
+	currRows  int
 	numRows   int
 	newWriter func() io.WriteCloser
 }
@@ -33,20 +35,22 @@ var (
 )
 
 func newCSVPipe(
-	in io.Reader, logger zerolog.Logger, flushSize int, newWriter func() io.WriteCloser,
+	in io.Reader,
+	logger zerolog.Logger,
+	flushSize int,
+	flushRows int,
+	newWriter func() io.WriteCloser,
 ) *csvPipe {
 	return &csvPipe{
 		in:        in,
 		logger:    logger,
 		flushSize: flushSize,
+		flushRows: flushRows,
 		newWriter: newWriter,
 	}
 }
 
 func (p *csvPipe) Pipe(tn dbtable.Name) error {
-	if err := p.flush(true); err != nil {
-		return err
-	}
 	r := csv.NewReader(p.in)
 	r.ReuseRecord = true
 	m := importedRows.WithLabelValues(tn.SafeString())
@@ -54,10 +58,12 @@ func (p *csvPipe) Pipe(tn dbtable.Name) error {
 		record, err := r.Read()
 		if err != nil {
 			if err == io.EOF {
-				return p.flush(false)
+				return p.flush()
 			}
 			return err
 		}
+		p.maybeInitWriter()
+		p.currRows++
 		p.numRows++
 		m.Inc()
 		if p.numRows%100000 == 0 {
@@ -69,15 +75,16 @@ func (p *csvPipe) Pipe(tn dbtable.Name) error {
 		if err := p.csvWriter.Write(record); err != nil {
 			return err
 		}
-		if p.currSize > p.flushSize {
-			if err := p.flush(true); err != nil {
+
+		if p.currSize > p.flushSize || (p.flushRows > 0 && p.currRows >= p.flushRows) {
+			if err := p.flush(); err != nil {
 				return err
 			}
 		}
 	}
 }
 
-func (p *csvPipe) flush(recreate bool) error {
+func (p *csvPipe) flush() error {
 	if p.csvWriter != nil {
 		p.csvWriter.Flush()
 		if err := p.out.Close(); err != nil {
@@ -85,12 +92,15 @@ func (p *csvPipe) flush(recreate bool) error {
 		}
 	}
 	p.currSize = 0
-	if recreate {
+	p.currRows = 0
+	p.out = nil
+	p.csvWriter = nil
+	return nil
+}
+
+func (p *csvPipe) maybeInitWriter() {
+	if p.csvWriter == nil {
 		p.out = p.newWriter()
 		p.csvWriter = csv.NewWriter(p.out)
-	} else {
-		p.out = nil
-		p.csvWriter = nil
 	}
-	return nil
 }
--- a/fetch/csv_pipe_test.go
+++ b/fetch/csv_pipe_test.go
@@ -17,6 +17,7 @@ func TestCSVPipe(t *testing.T) {
 		toWrite   string
 		files     []string
 		flushSize int
+		flushRows int
 	}{
 		{
 			desc: "one big file",
@@ -63,9 +64,68 @@ multiline part"
 `,
 				`2,a,c
 `,
-				``,
 			},
 			flushSize: 4,
+		},
+		{
+			desc: "flush after 1 row",
+			toWrite: `1,abcd,efgh
+2,efgh,""""
+3,%,g`,
+			files: []string{
+				"1,abcd,efgh\n",
+				`2,efgh,""""
+`,
+				`3,%,g
+`,
+			},
+			flushSize: 1024,
+			flushRows: 1,
+		},
+		{
+			desc: "flush after two rows",
+			toWrite: `1,abcd,efgh
+2,efgh,""""
+3,%,g`,
+			files: []string{
+				"1,abcd,efgh\n2,efgh,\"\"\"\"\n",
+				"3,%,g\n",
+			},
+			flushSize: 1024,
+			flushRows: 2,
+		},
+		{
+			desc: "flush after multiple rows",
+			toWrite: `1,abcd,efgh
+2,efgh,""""
+3,%,g`,
+			files: []string{
+				`1,abcd,efgh
+2,efgh,""""
+3,%,g
+`,
+			},
+			flushSize: 1024,
+			flushRows: 4,
+		},
+		{
+			desc: "flush after mix of flush size and flush rows",
+			toWrite: `1,abcd,efgh
+2,efgh,""""
+3,%,g
+4,a,b
+`,
+			files: []string{
+				`1,abcd,efgh
+`,
+				`2,efgh,""""
+3,%,g
+`,
+				`4,a,b
+`,
+			},
+			flushSize: 10,
+			flushRows: 2,
 		},
 	} {
 		t.Run(tc.desc, func(t *testing.T) {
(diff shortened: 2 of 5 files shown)


ryanluu12345:
@otan overall for Molt fetch, what types of testing do you prioritize and which do you prefer are added for these?

otan (commented):
lgtm pending tests

fetch/csv_pipe.go
```diff
@@ -69,7 +75,9 @@ func (p *csvPipe) Pipe(tn dbtable.Name) error {
 		if err := p.csvWriter.Write(record); err != nil {
 			return err
 		}
-		if p.currSize > p.flushSize {
+		// Flush size takes precedence since that is constrained by Blob storage limits/defaults.
+		// Then if the user set a number of rows to flush after, then we check if we should flush.
+		if p.currSize > p.flushSize || (p.flushRows != 0 && p.numRows%p.flushRows == 0) {
```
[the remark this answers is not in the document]
ryanluu12345: You're right, nothing to flush if there are zero rows:
```
(p.numRows != 0 && p.flushRows != 0 && p.numRows%p.flushRows == 0)
```
This should fix it ^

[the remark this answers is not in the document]
otan: oh no, i mean on flush you should reset the row count back to 0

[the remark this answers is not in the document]
otan: i just realised you use `%`, i think you can simplify it if you use `p.numRows > p.rowSize` and reset `p.rowSize = 0`, similar to how we do `flushSize`.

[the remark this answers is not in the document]
ryanluu12345: Ok makes sense. Given this feedback, it probably makes the most sense to wrap this into the `flush` behavior so that the counter resets both each time flush happens.

Then we can simplify to `p.numRows > p.flushRows` as you suggested.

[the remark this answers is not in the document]
ryanluu12345: @otan after trying this, I don't think it's as simple as this conditional statement. We do need the validity check since we want to make sure that we don't flush when numRows == 0 or p.flushRows == 0. In those two cases, we don't want to flush, otherwise, this could flush everytime, which would harm performance.

I went back to my original conditional statement.

fetch/csv_pipe_test.go
```diff
@@ -67,13 +68,56 @@ multiline part"
...
+			toWrite: `1,abcd,efgh
+2,efgh,""""
+3,%,g`,
+			files: []string{
+				"1,abcd,efgh\n",
+				`2,efgh,""""
+`,
+				`3,%,g
+`,
+				"",
```
ryanluu12345, line 82: This empty row seems logical to me since we have a new line that gets read and that is an empty string. So we would fully expect that it would show up in the output, based on the logic right now. However, the question is, do we want this logic? Or should we ignore empty lines?

I previously tried this:
```
for {
		record, err := r.Read()
                 if len(record) == 0 { continue }
```

However this led to an infinite loop, which I don't think is expected.... Still trying to think about this one.

otan: can you import an empty file? if so this is ok.

fetch/csv_pipe.go
```diff
@@ -69,7 +75,9 @@ func (p *csvPipe) Pipe(tn dbtable.Name) error {
 		if err := p.csvWriter.Write(record); err != nil {
 			return err
 		}
-		if p.currSize > p.flushSize {
+		// Flush size takes precedence since that is constrained by Blob storage limits/defaults.
+		// Then if the user set a number of rows to flush after, then we check if we should flush.
+		if p.currSize > p.flushSize || (p.numRows > 0 && p.flushRows > 0 && p.numRows%p.flushRows == 0) {
```
otan, line 80: i still think this should be p.numRows >= p.flushRows instead of `%`.

ryanluu12345: We still need to make sure that `p.numRows > 0`. Because in the case that we default to 0, we will always be true here and flush every time we are in the loop. So I guess we change the modulo part only and keep the gate for `p.numRows > 0 && p.flushRows > 0`.

ryanluu12345:
@otan not exactly sure if we can import an empty file. However, I can't seem to figure out why the empty string keeps getting written in the single line flush case...I put some `fmt.Println` statements and I can see that it spits out 4 records with the last being an empty record. It then goes to the `flush` call.

I have a theory that when we flush the last time, the empty string is in the buffer, which is why it gets written to its own file. Maybe this is a property of our test writeCloser?
```
func() io.WriteCloser {
					bufs = append(bufs, testStringBuf{})
					return &bufs[len(bufs)-1]
				},
```
Since it appears here that we put testStringBuf and append to the buffer regardless.

otan:
in cockroachdb, try run `IMPORT INTO` with an empty file. if that works, then we'll allow it.

otan:
anyway, the reason why a new file gets created is because `flush` too eagerly creates a new file. to fix that (and a few comments), apply this patch:

```diff
diff --git a/fetch/csv_pipe.go b/fetch/csv_pipe.go
index e4b87b4..4b35838 100644
--- a/fetch/csv_pipe.go
+++ b/fetch/csv_pipe.go
@@ -50,9 +50,6 @@ func newCSVPipe(
 }
 
 func (p *csvPipe) Pipe(tn dbtable.Name) error {
-	if err := p.flush(true); err != nil {
-		return err
-	}
 	r := csv.NewReader(p.in)
 	r.ReuseRecord = true
 	m := importedRows.WithLabelValues(tn.SafeString())
@@ -60,10 +57,11 @@ func (p *csvPipe) Pipe(tn dbtable.Name) error {
 		record, err := r.Read()
 		if err != nil {
 			if err == io.EOF {
-				return p.flush(false)
+				return p.flush()
 			}
 			return err
 		}
+		p.maybeInitWriter()
 		p.numRows++
 		m.Inc()
 		if p.numRows%100000 == 0 {
@@ -75,17 +73,16 @@ func (p *csvPipe) Pipe(tn dbtable.Name) error {
 		if err := p.csvWriter.Write(record); err != nil {
 			return err
 		}
-		// Flush size takes precedence since that is constrained by Blob storage limits/defaults.
-		// Then if the user set a number of rows to flush after, then we check if we should flush.
-		if p.currSize > p.flushSize || (p.numRows > 0 && p.flushRows > 0 && p.numRows >= p.flushRows) {
-			if err := p.flush(true); err != nil {
+		// We flush if we exceed target file or row size (if a row size exists).
+		if p.currSize > p.flushSize || (p.flushRows > 0 && p.numRows >= p.flushRows) {
+			if err := p.flush(); err != nil {
 				return err
 			}
 		}
 	}
 }
 
-func (p *csvPipe) flush(recreate bool) error {
+func (p *csvPipe) flush() error {
 	if p.csvWriter != nil {
 		p.csvWriter.Flush()
 		if err := p.out.Close(); err != nil {
@@ -94,12 +91,14 @@ func (p *csvPipe) flush(recreate bool) error {
 	}
 	p.currSize = 0
 	p.numRows = 0
-	if recreate {
+	p.out = nil
+	p.csvWriter = nil
+	return nil
+}
+
+func (p *csvPipe) maybeInitWriter() {
+	if p.csvWriter == nil {
 		p.out = p.newWriter()
 		p.csvWriter = csv.NewWriter(p.out)
-	} else {
-		p.out = nil
-		p.csvWriter = nil
 	}
-	return nil
 }
diff --git a/fetch/csv_pipe_test.go b/fetch/csv_pipe_test.go
index 550d3b8..8bbd88b 100644
--- a/fetch/csv_pipe_test.go
+++ b/fetch/csv_pipe_test.go
@@ -64,7 +64,6 @@ multiline part"
 `,
 				`2,a,c
 `,
-				``,
 			},
 			flushSize: 4,
 		},
@@ -79,7 +78,6 @@ multiline part"
 `,
 				`3,%,g
 `,
-				"",
 			},
 			flushSize: 1024,
 			flushRows: 1,

```

fetch/csv_pipe.go
```diff
@@ -85,6 +93,7 @@ func (p *csvPipe) flush(recreate bool) error {
 		}
 	}
 	p.currSize = 0
+	p.numRows = 0
```
otan, line 96: i just discovered we use `p.numRows` for stats as well (see p.numRows%100000).
do you mind adding a `currRows` variable which tracks flushing instead? we need both as `currRows` may not be moddable if we flush due to `flushSize` earlier

fetch/csv_pipe.go
```diff
@@ -69,7 +75,9 @@ func (p *csvPipe) Pipe(tn dbtable.Name) error {
 		if err := p.csvWriter.Write(record); err != nil {
 			return err
 		}
-		if p.currSize > p.flushSize {
+		// Flush size takes precedence since that is constrained by Blob storage limits/defaults.
+		// Then if the user set a number of rows to flush after, then we check if we should flush.
```
otan, line 79: just realised this comment really map here - there's no precedence; if we exceed flush or row size we flush.
this was fixed up in the diff

ryanluu12345:
> in cockroachdb, try run `IMPORT INTO` with an empty file. if that works, then we'll allow it.

Done!
```
root@:26257/defaultdb> IMPORT INTO users (id, city, name, address, credit_card)                                                                                                                                   CSV DATA (                                                                                                                                                                                                'userfile:///empty.csv'                                                                                                                                                                                       );
        job_id       |  status   | fraction_completed | rows | index_entries | bytes
---------------------+-----------+--------------------+------+---------------+--------
  903412610293465089 | succeeded |                  1 |    0 |             0 |     0
(1 row)


Time: 3.105s total (execution 3.103s / network 0.001s)

root@:26257/defaultdb> IMPORT INTO users (id, city, name, address, credit_card)                                                                                                                                   CSV DATA (                                                                                                                                                                                                'userfile:///empty.csv'                                                                                                                                                                                       );
        job_id       |  status   | fraction_completed | rows | index_entries | bytes
---------------------+-----------+--------------------+------+---------------+--------
  903412626888982529 | succeeded |                  1 |    0 |             0 |     0
(1 row)


Time: 3.085s total (execution 3.084s / network 0.000s)
```
It works!

fetch/csv_pipe.go
ryanluu12345: Done!

fetch/csv_pipe.go
ryanluu12345: Done!

ryanluu12345: Using cur rows. We no longer reset `numRows` since that's used for stats. We do currRows instead.

fetch/csv_pipe_test.go
```diff
@@ -63,17 +64,58 @@ multiline part"
...
+3,%,g`,
+			files: []string{
+				"1,abcd,efgh\n2,efgh,\"\"\"\"\n",
+				"3,%,g\n",
+			},
+			flushSize: 1024,
+			flushRows: 2,
+		},
+		{
+			desc: "flush after multiple",
```
otan, line 98: nit: `flush after multiple rows`

can you add 1 more test where it flushes in a mix of flush size and flush rows?

ryanluu12345: All done!

otan (approved):

PR #39, opened 2023-09-28
fetch: added feature to compress while writing and decompress while importing the compressed gzip files

ryanluu12345:
- Added cmd/fetch option for this functionality, defaulting to GZIP for `IMPORT INTO` and no compression for `COPY`
- Added in interface changes to `ImportInto` helper method so that it now takes KV options


ryanluu12345:
Ok did some testing here. There are a few bugs I noticed:
1. Incorrect query string in the case where there are no options. The `WITH` is appended, which unfortunately means that we need to update this so that options don't get added unless they're needed.
2. We also see that we are hanging at the `data extraction phase` for `public.bins3` when we use `compression` mode.
```
<nil> ERR table public.bins2 do not have matching primary keys, cannot migrate table=public.bins2
<nil> INF data extraction phase starting table=public.bins3
^Csignal: interrupt
```
The only changes made here were wrapping the forwardWriter with a gzip compression. I'm wondering if this runs into issues because we have two writers we must close?

cmd/fetch/fetch.go
```diff
@@ -37,6 +39,20 @@ func Command() *cobra.Command {
 			}
 			cmdutil.RunMetricsServer(logger)
 
+			if cfg.Live && cfg.Compression != compression.Default {
+				return errors.New("cannot run live mode with compression")
+			}
+
+			if cfg.Live && cfg.Compression == compression.Default {
+				logger.Info().Msgf("Default compression for live mode is none.")
```
otan, line 47: nit: lowercase, no full stop/period. rather you do something like
```go
if cfg.Compression == compression.Default {
   if cfg.Live {
      logger.Info().Msgf("default compression to none")
} else {

      logger.Info().Msgf("default compression to gzip")
}
}
```

cmd/fetch/fetch.go
```diff
@@ -184,6 +200,16 @@ func Command() *cobra.Command {
...
 	)
+	cmd.PersistentFlags().Var(
+		enumflag.New(
+			&cfg.Compression,
+			"compression",
+			compression.CompressionStringRepresentations,
+			enumflag.EnumCaseInsensitive,
+		),
+		"compression",
+		"compression to use (default/gzip/none)",
```
otan, line 211: you do not need `(default/gzip/none)` - i *think* gflags / enumflag gives that to you

compression/compression.go
```diff
@@ -0,0 +1,18 @@
+package compression
+
+import "github.com/thediveo/enumflag/v2"
+
+//go:generate go run github.com/alvaroloes/enumer -type=Compression -output compression_enumer.gen.go
+type Compression enumflag.Flag
```
otan, line 6: nit: name it `Flag`

fetch/export_table.go
```diff
@@ -21,14 +23,22 @@ type exportResult struct {
...
+
 func exportTable(
 	ctx context.Context,
 	logger zerolog.Logger,
 	sqlSrc dataexport.Source,
 	datasource datablobstorage.Store,
 	table dbtable.VerifiedTable,
-	flushSize int,
-	flushRows int,
+	cfg *Config,
```
otan, line 41: nit: do not take the pointer to the config (i.e. just `cfg Config`), make it the second argument instead in line with everything else. ideally in go, you pass by copy unless you have to or you're subject to side effects on the ptr

fetch/import_table.go
```diff
@@ -24,6 +26,7 @@ func importTable(
 	logger zerolog.Logger,
 	table dbtable.VerifiedTable,
 	resources []datablobstorage.Resource,
+	compressionOpt compression.Compression,
```
otan, line 29: might as well take in config too at this point

otan:
> <nil> ERR table public.bins2 do not have matching primary keys, cannot migrate table=public.bins2

this seems like the actual problem - your primary keys don't match. did you change anything?
that said we shouldn't hang when it happens - we should error out early

otan:
oh right, you're saying it hangs on `bins3`. it may be because of what you suggest, in which case you'd need to make something that implements `io.WriteCloser` that closers up and down the stack

fetch/export_table.go
```diff
@@ -82,7 +93,7 @@ func exportTable(
 				writerErrCh <- err
 			}
 		}()
-		return forwardWrite
+		return wrappedWriter
```
otan, line 96: the files that are uploaded to s3/gcp should probably also have the correct extension (csv/tar.gz)

fetch/fetch_test.go
```diff
@@ -83,7 +83,8 @@ func TestDataDriven(t *testing.T) {
 						if direct {
 							src = datablobstorage.NewCopyCRDBDirect(logger, conns[1].(*dbconn.PGConn).Conn)
 						} else {
-							src, err = datablobstorage.NewLocalStore(logger, dir, "localhost:4040", "localhost:4040")
+							// TODO: update the extension so that it's dynamic based on what it needs to be.
```
otan, line 86: yeah you'll want to add in more `CmdArgs` which take compression, parses as flags and calls `Fetch` with the right config.

fetch/datablobstorage/s3.go
```diff
@@ -80,20 +81,25 @@ func (r s3Reader) Close() error {
...
 		bucket:  bucket,
 		session: session,
 		logger:  logger,
 		creds:   creds,
+		fileExt: fileExt,
 	}
 }
 
 func (s *s3Store) CreateFromReader(
 	ctx context.Context, r io.Reader, table dbtable.VerifiedTable, iteration int,
```
otan, line 100: i'd make the extension part of this interface instead of `NewS3Store`/GCPStore/etc

cmd/fetch/fetch.go
ryanluu12345: It doesn't give it in the `help` command. Only when you provide an erroneous value. For clarity, I'll just include this.

fetch/export_table.go
ryanluu12345: Fixed this by passing this and storing it in the struct for concrete implementations.

fetch/fetch_test.go
ryanluu12345: Yep sounds good. I'll add this functionality so that the test can use input that relies on `compression`.

cmd/fetch/fetch.go
ryanluu12345: Yep addressed!

compression/compression.go
ryanluu12345: Yep addressed

fetch/export_table.go
ryanluu12345: Ok doing `cfg Config` without pointer.

fetch/import_table.go
ryanluu12345: Yep addressed!

fetch/datablobstorage/s3.go
ryanluu12345: Ok I fixed this. Made it part of `CreateFromReader` now.

otan:
I'll look again after tests are added

ryanluu12345:
> > ERR table public.bins2 do not have matching primary keys, cannot migrate table=public.bins2
> 
> this seems like the actual problem - your primary keys don't match. did you change anything? that said we shouldn't hang when it happens - we should error out early

We don't hang because of this. I understand why. My table on the target side is wrong. I'll delete these tables haha.

ryanluu12345:
> oh right, you're saying it hangs on `bins3`. it may be because of what you suggest, in which case you'd need to make something that implements `io.WriteCloser` that closes both io.Writers.
> 
> that said, are you sure it's hanging on the import phase or export phase?

Ended up confirming that this is a matter of leaking `io.Pipe`s. Fixed.

ryanluu12345:
@otan oddly enough, the compressed mode seems to take longer for imports. This is a bit odd, because I fully expect the import to be much faster with compression. Just a note that I run the test for no compression, immediately delete the data, then run the molt fetch with compression right after. I'm wondering if this affects the latency I see here since it might still be processing the deletes under the hood? Here is a comparison between the two:
**No Compression**
```
(⎈|docker-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % go run . fetch \
--source 'postgres://postgres@localhost:5432/postgres' \
--local-path /tmp/basic \
--local-path-listen-addr '0.0.0.0:9005' --compression none
# github.com/cockroachdb/molt
ld: warning: '/private/var/folders/kz/ttm0chqn0r30bn75tvq4pnr80000gq/T/go-link-1822718239/go.o' has malformed LC_DYSYMTAB, expected 129 undefined symbols to start at index 71186, found 138 undefined symbols starting at index 220
<nil> INF checking database details
<nil> INF starting file server crdb-access-addr=192.168.0.48:9005 listen-addr=0.0.0.0:9005
<nil> WRN ignoring table as it is missing a definition on the source table=public._crdb_proxy_settings
<nil> WRN ignoring table as it is missing a definition on the source table=public.bins
<nil> WRN ignoring table as it is missing a definition on the source table=public.bins2
<nil> WRN ignoring table as it is missing a definition on the source table=public.bob
<nil> WRN ignoring table as it is missing a definition on the source table=public.timestamp_tbl
<nil> WRN ignoring table as it is missing a definition on the source table=public.userfiles_root_upload_files
<nil> WRN ignoring table as it is missing a definition on the source table=public.userfiles_root_upload_payload
<nil> WRN ignoring table as it is missing a definition on the source table=public.users
<nil> WRN ignoring table as it is missing a definition on the source table=public.workload_table
<nil> WRN ignoring table as it is missing a definition on the source table=public.workload_table2
<nil> WRN ignoring table as it is missing a definition on the source table=public.yo
<nil> INF found matching table source_table=public.bins3 target_table=public.bins3
<nil> INF verifying common tables
<nil> INF establishing snapshot
<nil> INF starting fetch cdc_cursor=0/310DBA8 num_tables=1
<nil> INF data extraction phase starting table=public.bins3
<nil> INF row import status num_rows=100000 table=public.bins3
<nil> INF row import status num_rows=200000 table=public.bins3
<nil> INF data extraction from source complete export_duration=161.255292 num_rows=200000 table=public.bins3
<nil> INF starting data import on target table=public.bins3
<nil> INF data import on target for table complete cdc_cursor=0/310DBA8 import_duration=1698.234083 net_duration=1862.002292 table=public.bins3
<nil> INF fetch complete cdc_cursor=0/310DBA8 num_tables=1 tables=["public.bins3"]
```

**Compression**
```
(⎈|docker-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % go run . fetch \
--source 'postgres://postgres@localhost:5432/postgres' \
--local-path /tmp/basic \
--local-path-listen-addr '0.0.0.0:9005' --compression gzip
# github.com/cockroachdb/molt
ld: warning: '/private/var/folders/kz/ttm0chqn0r30bn75tvq4pnr80000gq/T/go-link-1735573185/go.o' has malformed LC_DYSYMTAB, expected 129 undefined symbols to start at index 71186, found 138 undefined symbols starting at index 220
<nil> INF checking database details
<nil> INF starting file server crdb-access-addr=192.168.0.48:9005 listen-addr=0.0.0.0:9005
<nil> WRN ignoring table as it is missing a definition on the source table=public._crdb_proxy_settings
<nil> WRN ignoring table as it is missing a definition on the source table=public.bins
<nil> WRN ignoring table as it is missing a definition on the source table=public.bins2
<nil> WRN ignoring table as it is missing a definition on the source table=public.bob
<nil> WRN ignoring table as it is missing a definition on the source table=public.timestamp_tbl
<nil> WRN ignoring table as it is missing a definition on the source table=public.userfiles_root_upload_files
<nil> WRN ignoring table as it is missing a definition on the source table=public.userfiles_root_upload_payload
<nil> WRN ignoring table as it is missing a definition on the source table=public.users
<nil> WRN ignoring table as it is missing a definition on the source table=public.workload_table
<nil> WRN ignoring table as it is missing a definition on the source table=public.workload_table2
<nil> WRN ignoring table as it is missing a definition on the source table=public.yo
<nil> INF found matching table source_table=public.bins3 target_table=public.bins3
<nil> INF verifying common tables
<nil> INF establishing snapshot
<nil> INF starting fetch cdc_cursor=0/310DBA8 num_tables=1
<nil> INF data extraction phase starting table=public.bins3
<nil> INF row import status num_rows=100000 table=public.bins3
<nil> INF row import status num_rows=200000 table=public.bins3
<nil> INF data extraction from source complete export_duration=194.987375 num_rows=200000 table=public.bins3
<nil> INF starting data import on target table=public.bins3
<nil> INF data import on target for table complete cdc_cursor=0/310DBA8 import_duration=1848.908625 net_duration=2046.582417 table=public.bins3
<nil> INF fetch complete cdc_cursor=0/310DBA8 num_tables=1 tables=["public.bins3"]
```

ryanluu12345:
Added the test scaffolding and it integrates fine; however there are still remaining test failures to debug. It's mainly related to the compressed files not being able to be found:
```
{"level":"error","table":"public.compresstbl2","error":"error importing data: ERROR: error response from server: 404 Not Found \"404 page not found\\n\": http storage file does not exist: external_storage: file doesn't exist (SQLSTATE XXUUU)","message":"error importing data, retrying"}
{"level":"error","table":"public.compresstbl","error":"error importing data: ERROR: error response from server: 404 Not Found \"404 page not found\\n\": http storage file does not exist: external_storage: file doesn't exist (SQLSTATE XXUUU)","message":"error importing data, retrying"}
{"level":"error","table":"public.compresstbl2","error":"error importing data: ERROR: error response from server: 404 Not Found \"404 page not found\\n\": http storage file does not exist: external_storage: file doesn't exist (SQLSTATE XXUUU)","message":"error importing data, retrying"}
{"level":"error","table":"public.compresstbl","error":"error importing data: ERROR: error response from server: 404 Not Found \"404 page not found\\n\": http storage file does not exist: external_storage: file doesn't exist (SQLSTATE XXUUU)","message":"error importing data, retrying"}
{"level":"error","table":"public.compresstbl2","error":"error importing data: ERROR: error response from server: 404 Not Found \"404 page not found\\n\": http storage file does not exist: external_storage: file doesn't exist (SQLSTATE XXUUU)","message":"error importing data, retrying"}
{"level":"error","table":"public.compresstbl","error":"error importing data: ERROR: error response from server: 404 Not Found \"404 page not found\\n\": http storage file does not exist: external_storage: file doesn't exist (SQLSTATE XXUUU)","message":"error importing data, retrying"}
``` 
This is odd to me because I tested this whole flow e2e locally, which should use the local server, and it doesn't appear to have any issues there.

fetch/internal/dataquery/dataquery.go
```diff
@@ -25,13 +25,20 @@ func NewPGCopyTo(table dbtable.VerifiedTable) string {
 	return f.CloseAndGetString()
 }
 
-func ImportInto(table dbtable.VerifiedTable, locs []string) string {
+func ImportInto(table dbtable.VerifiedTable, locs []string, opts tree.KVOptions) string {
 	importInto := &tree.Import{
 		Into:       true,
 		Table:      table.NewTableName(),
 		FileFormat: "CSV",
```
ryanluu12345, line 32: @otan open question here, does this file format have to be a compressed format like `tar.tgz` or is Cockroach smart enough to know that `CSV` + `decompress` means the underlying is CSV but .tar.tgz as the outer layer?

fetch/fetch_test.go
ryanluu12345: Added this!

cmd/fetch/fetch.go
```diff
@@ -37,6 +39,20 @@ func Command() *cobra.Command {
 			}
 			cmdutil.RunMetricsServer(logger)
 
+			if cfg.Live && cfg.Compression == compression.GZIP {
+				return errors.New("cannot run live mode with compression")
+			}
+
+			if cfg.Live && cfg.Compression <= compression.Default {
```
ryanluu12345, line 46: The first value in the enum is actually iota + 1, which is 1. But in the struct, it gets initialized to 0. So we check below or equal to the default value because those are guaranteed to be default values.

otan: i still rather the `if cfg.Live { ... } else { ... }` statement. it's less to logically reason about.

fetch/writer.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"compress/gzip"
+	"io"
+)
+
+// compressedPipeWriter wraps around the underlying pipe writer and
+// gzip writer so that we can access both to close them.
+type compressedPipeWriter struct {
+	pipeWriter *io.PipeWriter
+	gzipWriter *gzip.Writer
```
otan, line 12: nit: let's make this a generic `io.WriterCloser`, in case we decide to support more compressions later

fetch/fetch_test.go
```diff
@@ -83,10 +89,17 @@ func TestDataDriven(t *testing.T) {
 						if direct {
 							src = datablobstorage.NewCopyCRDBDirect(logger, conns[1].(*dbconn.PGConn).Conn)
 						} else {
-							src, err = datablobstorage.NewLocalStore(logger, dir, "localhost:4040", "localhost:4040")
+							addr := fmt.Sprintf("localhost:%d", port)
+							src, err = datablobstorage.NewLocalStore(logger, dir, addr, addr)
```
otan, line 93: we shouldn't have to increment the port - if we do, it means something was closed incorrectly and we should add a method to close it.

fetch/internal/dataquery/dataquery.go
otan: the latter - this is fine.

fetch/writer.go
ryanluu12345: Will do!

fetch/fetch_test.go
ryanluu12345: Yep I agree. I ended up storing the server in the `localStorage` struct. That way the caller can call the `Shutdown`. Originally I wanted to put it into the `Cleanup` method of the interface, but I was thinking there may be cases where a shutdown runs but the program expects the HTTP server to be up still. I also didn't want to add another method for Shutdown because I'm not sure we need these behavior for each concrete implementation. Wondering your thoughts here? Can standardize more as needed.

fetch/writer.go
```diff
@@ -0,0 +1,36 @@
...
+	return c.pipeWriter.Close()
+}
+
+// The gzip writer is responsible for writing the data to the compressed file,
+// since it wraps around the pipeWriter.
+func (c *compressedPipeWriter) Write(p []byte) (n int, err error) {
+	return c.gzipWriter.Write(p)
+}
+
+func newCompressedPipeWriter(w *io.PipeWriter) *compressedPipeWriter {
```
otan, line 31: nit: `newGZIPPipeWriter`, rename `gzipWriter` to `compressionWriter`

fetch/fetch_test.go
otan: i think we can safely put it in cleanup. once that's called, i dont expect the same data store to be used again.

fetch/writer.go
ryanluu12345: Done

fetch/datablobstorage/local.go
```diff
@@ -42,12 +44,16 @@ func NewLocalStore(
 			port := splat[1]
 			crdbAccessAddr = ip + ":" + port
 		}
+		// We want to instantiate a server so that we can return it
```
otan, line 47: nit: i tihnk you can drop this comment

fetch/datablobstorage/local.go
```diff
@@ -29,6 +30,7 @@ func NewLocalStore(
 	if err := os.MkdirAll(basePath, os.ModePerm); err != nil {
 		return nil, err
 	}
+	server := http.Server{}
```
otan, line 33: this should be nil if not setup, i.e. do `var server *http.Server` instead

otan (approved):

fetch/datablobstorage/local.go
ryanluu12345: Ok got it!

fetch/datablobstorage/local.go
ryanluu12345: Done!

fetch/datablobstorage/local.go
```diff
@@ -116,7 +123,8 @@ func (l *localStore) Cleanup(ctx context.Context) error {
 			return err
 		}
 	}
-	return nil
+
+	return l.server.Shutdown(ctx)
```
otan, line 127: ... and only do this `if l.server != nil` (sorry missed that). forgot why i did the `listenaddr != ""` and what that supports tho.

ryanluu12345: done

PR #40, opened 2023-09-28
.github/workflows: fix duplicated runs because of pull request and push actions

ryanluu12345:


ryanluu12345:
The PR shows only 3 runs instead of 6.

otan (approved):

PR #42, opened 2023-10-04
fetch: added in progress indicator for the import table jobs

ryanluu12345:
Added in a Go routine that repeatedly queries the database for the IMPORT job and reports progress to the STDOUT.

Release Note: None

Action items:
- [x] Don't print 0 if that's the percentage
- [x] Don't print the same number multiple times
- [x] Ask #sql to know how the fraction_completed is calculated

### DIFF
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -219,6 +219,18 @@ func fetchTable(
 				Msgf("starting data import on target")
 
 			if !cfg.Live {
+				go func() {
+					err := reportImportTableProgress(ctx,
+						targetConn,
+						logger,
+						table.VerifiedTable,
+						time.Now(),
+						false /*testing*/)
+					if err != nil {
+						logger.Err(err).Msg("failed to report import table progress")
+					}
+				}()
+
 				r, err := importTable(ctx, cfg, targetConn, logger, table.VerifiedTable, e.Resources)
 				if err != nil {
 					return err
--- a/fetch/import_table.go
+++ b/fetch/import_table.go
@@ -2,6 +2,10 @@ package fetch
 
 import (
 	"context"
+	"fmt"
+	"math"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/cockroachdb/cockroachdb-parser/pkg/sql/sem/tree"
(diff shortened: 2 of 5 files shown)


fetch/fetch.go
```diff
@@ -219,6 +219,13 @@ func fetchTable(
 				Msgf("starting data import on target")
 
 			if !cfg.Live {
+				go func() {
+					err := reportImportTableProgress(ctx, targetConn, logger, table.VerifiedTable)
```
otan, line 223: this routine should exit when `importTable` is complete. ideally do this after `importTable` returns - otherwise a stray 100% may be returned.

fetch/import_table.go
```diff
@@ -20,6 +23,66 @@ type importResult struct {
...
+		}
+		defer rows.Close()
+
+		p, err := pgx.CollectRows(rows, pgx.RowToStructByName[importProgress])
+		if err != nil {
+			return err
+		} else if len(p) == 0 {
+			return errors.New("retrying because no rows found")
+		} else if p[0].FractionCompleted != 1 {
+			logger.Info().Msgf("%.2f%%", p[0].FractionCompleted*100)
```
otan, line 74: for now, let's make message the constant but variable fields change here
```
logger.Info().String("completion", fmt.Sprintf("%s%", frac*100)).Msgf("progress")
```

as discussed, we should also make it such that we don't print `0%`, and we don't repeat the same `%` as before as otherwise we look like we're stuck.

fetch/fetch.go
otan: oh the routine exits when 100% is reached. we should probably not output 100%

fetch/import_table.go
```diff
@@ -20,6 +23,66 @@ type importResult struct {
...
+	if err := r.Do(func() error {
+		query := fmt.Sprintf(`
+		WITH x as (SHOW JOBS) 
+		SELECT description, started, fraction_completed 
+		FROM x 
+		WHERE job_type='IMPORT' 
+			AND description LIKE '%%.%s(id)%%'
+			AND started > '%s'
+		ORDER BY created DESC
+		`, table.Table.String(), curTime)
```
otan, line 61: you need to include the schema name as well or else multiple tables in diff schemas with the same name may be incorrectly returned

also what is the `(id)` for?

fetch/fetch.go
ryanluu12345: Yep, made it so that it does not report 100%, 0%, or the same repeated values.

fetch/import_table.go
ryanluu12345: Let me look more into the escape character aspect.

ryanluu12345:
@otan , got some insight from @cucaroach and he pointed me to this limitation of `fraction_completed` for IMPORT:
https://cockroachlabs.atlassian.net/browse/CRDB-19266

fetch/import_table.go
```diff
@@ -20,6 +23,73 @@ type importResult struct {
...
+	r, err := retry.NewRetry(retry.Settings{
+		InitialBackoff: time.Second,
+		Multiplier:     1,
+		MaxRetries:     math.MaxInt64,
+	})
+	if err != nil {
+		return err
+	}
+
+	pgConn := baseConn.(*dbconn.PGConn)
```
otan, line 45: let's make this a little safer 
```go
pgConn, ok := ....
if !ok {
   return errors.Newf("expected pgx conn, got %T", baseConn)
}
```

otan (commented):
code lgtm. are you able to unit test `reportImportTableProgress` by e.g. manually triggering an import job?

someone else will have to approve once the tests are added.

fetch/import_table.go
ryanluu12345: Added this change!

fetch/import_table.go
ryanluu12345: Yep, just added escaping for the `%` case. I also verified that the generated SQL works.

ryanluu12345:
> code lgtm. are you able to unit test `reportImportTableProgress` by e.g. manually triggering an import job?
> 
> someone else will have to approve once the tests are added.

Tests all added. Injected logger to capture output. Having @Jeremyyang920 review.

fetch/import_table_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+		err = reportImportTableProgress(ctx, conn, logger, dbtable.VerifiedTable{
+			Name: dbtable.Name{
+				Schema: tree.Name("public"),
+				Table:  tree.Name("teams"),
+			},
+		}, curTime, true /*testing*/)
+		require.NoError(t, err)
+		require.Equal(t, `{"level":"info","message":"100.00% completed (public.teams)"}
+`, b.String())
```
Jeremyyang920, line 111: Is this always going to be true? some reason import ran slow. Can you get another percentage log here? I can see this potentially flaking if you had a larger test file.

ryanluu12345: It is, because the testing line is at the logical path where it hits 100% completion. I decided to put the test logging after it hits 100% because this line is always guaranteed to be hit eventually.

ryanluu12345: As you can see below, in every other case, we return and retry. Then eventually we will hit that line.
```
if err != nil {
			return err
		} else if len(p) == 0 {
			return errors.New("retrying because no rows found")
		} else if p[0].FractionCompleted != 1 {
			frac := p[0].FractionCompleted
			if frac != 0.0 && prevVal != frac {
				logger.Info().Str("completion", fmt.Sprintf("%.2f%%", frac*100)).Msgf("progress")
			}

			prevVal = p[0].FractionCompleted
			return errors.New("retrying because job not finished yet")
		}

		if testing {
			logger.Info().Msgf("%.2f%% completed (%s.%s)", p[0].FractionCompleted*100, table.Schema.String(), table.Table.String())
		}
```

Jeremyyang920 (approved):

Issue #59, opened 2023-11-08
fetch/verify: add in a description for each tool and flag in help output


ryanluu12345:
- Description: fetch is missing from the molt help output
- Repro Steps: run `molt` command with no other subcommands
- Expected Behavior: i’d see a line about fetch

PR #68, opened 2023-11-14
molt: made the command flags and descriptions consistent with other migrations tooling

ryanluu12345:
Made the command descriptions and flags have consistent formatting. Prioritizing full sentences and better clarity about what each flag does for both fetch and verify. Also made the help command and autocomplete hooks so that it has behavior we expect. Removed the autocomplete command because that is extraneous.

Resolves: #59
Release Note: None

### DIFF
--- a/cmd/fetch/fetch.go
+++ b/cmd/fetch/fetch.go
@@ -28,9 +28,9 @@ func Command() *cobra.Command {
 		cfg                     fetch.Config
 	)
 	cmd := &cobra.Command{
-		Use:  "fetch",
-		Long: `Fetches data from source to directly import into target.`,
-
+		Use:   "fetch",
+		Short: "Moves data from source to target.",
+		Long:  `Imports data from source directly into target tables.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := context.Background()
 
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
 		&directCRDBCopy,
 		"direct-copy",
 		false,
-		"whether to use direct copy mode",
+		"Enables direct copy mode, which copies data directly from source to target without using an intermediate store.",
 	)
 	cmd.PersistentFlags().BoolVar(
 		&cfg.Cleanup,
 		"cleanup",
 		false,
-		"whether any file resources created should be deleted",
+		"Whether any created resources should be deleted.",
 	)
 	cmd.PersistentFlags().BoolVar(
 		&cfg.Live,
 		"live",
 		false,
-		"whether the table must be queriable during load import",
+		"Whether the table must be queryable during load import.",
 	)
 	cmd.PersistentFlags().IntVar(
 		&cfg.FlushSize,
 		"flush-size",
 		0,
-		"if set, size (in bytes) before the data source is flushed",
+		"If set, size (in bytes) before the source data is flushed to intermediate files.",
 	)
 	cmd.PersistentFlags().IntVar(
 		&cfg.FlushRows,
 		"flush-rows",
 		0,
-		"if set, number of rows before the data source is flushed",
+		"If set, number of rows before the source data is flushed to intermediate files.",
 	)
 	cmd.PersistentFlags().IntVar(
 		&cfg.Concurrency,
 		"concurrency",
 		4,
-		"number of tables to move data with at a time",
+		"Number of tables to move at a time.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&s3Bucket,
 		"s3-bucket",
 		"",
-		"s3 bucket",
+		"Name of the S3 bucket.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&gcpBucket,
 		"gcp-bucket",
 		"",
-		"gcp bucket",
+		"Name of the GCP bucket.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&bucketPath,
 		"bucket-path",
 		"",
-		"path within the bucket to write intermediate files (i.e. test-migrations/v1 with no ending slash)",
+		"Path within the bucket where intermediate files are written (e.g., bucket-name/folder-name).",
 	)
 	cmd.PersistentFlags().StringVar(
 		&localPath,
 		"local-path",
 		"",
-		"path to upload files to locally",
+		"Path to upload files to locally.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&localPathListenAddr,
 		"local-path-listen-addr",
 		"",
-		"local address to listen to for traffic",
+		"Address of a local store server to listen to for traffic.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&localPathCRDBAccessAddr,
 		"local-path-crdb-access-addr",
 		"",
-		"address CockroachDB can access to connect to the --local-path-listen-addr",
+		"Address of data that CockroachDB can access to import from a local store (defaults to local-path-listen-addr).",
 	)
 	cmd.PersistentFlags().BoolVar(
 		&cfg.Truncate,
 		"truncate",
 		false,
-		"whether to truncate the table being imported to",
+		"Whether to truncate the target tables before source data is imported.",
 	)
 	cmd.PersistentFlags().IntVar(
 		&cfg.ExportSettings.RowBatchSize,
 		"row-batch-size",
 		100_000,
-		"how many rows to select at a time for the database",
+		"Number of rows to select at a time for export from the source database.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&cfg.ExportSettings.PG.SlotName,
 		"pg-logical-replication-slot-name",
 		"",
-		"if set, the name of a replication slot that should be created before taking a snapshot of data",
+		"If set, the name of a replication slot that will be created before taking a snapshot of data.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&cfg.ExportSettings.PG.Plugin,
 		"pg-logical-replication-slot-plugin",
 		"pgoutput",
-		"if set, the output plugin used for logical replication under pg-logical-replication-slot-name",
+		"If set, the output plugin used for logical replication under pg-logical-replication-slot-name.",
 	)
 	cmd.PersistentFlags().BoolVar(
 		&cfg.ExportSettings.PG.DropIfExists,
 		"pg-logical-replication-slot-drop-if-exists",
 		false,
-		"if set, drops the replication slot if it exists",
+		"If set, drops the replication slot if it exists.",
 	)
 	cmd.PersistentFlags().Var(
 		enumflag.New(
@@ -214,7 +214,7 @@ func Command() *cobra.Command {
 			enumflag.EnumCaseInsensitive,
 		),
 		"compression",
-		"compression to use (default/gzip/none)",
+		"Compression type (default/gzip/none) to use (IMPORT INTO mode only).",
 	)
 	cmdutil.RegisterDBConnFlags(cmd)
 	cmdutil.RegisterLoggerFlags(cmd)
--- a/cmd/verify/verify.go
+++ b/cmd/verify/verify.go
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
 		&verifyConcurrency,
 		"concurrency",
 		0,
-		"number of tables to process at a time (defaults to number of CPUs)",
+		"Number of tables to process at a time (defaults to number of CPUs).",
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyTableSplits,
 		"table-splits",
 		1,
-		"number of shards to break down each table into whilst doing row-based verification",
+		"Number of shards to break down each table into while doing row-based verification.",
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyRowBatchSize,
 		"row-batch-size",
 		20000,
-		"number of rows to get from a table at a time",
+		"Number of source/target rows to scan at a time.",
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyLimitRowsPerSecond,
 		"rows-per-second",
 		0,
-		"if set, maximum number of rows to read per second during scanning per shard",
+		"If set, maximum number of rows to read per second on each shard.",
 	)
 	cmd.PersistentFlags().BoolVar(
 		&verifyFixup,
 		"fixup",
 		false,
-		"whether to fix up any rows",
+		"Whether to fix up inconsistencies found during row verification.",
 	)
 	cmd.PersistentFlags().DurationVar(
 		&verifyContinuousPause,
 		"continuous-pause-between-runs",
 		0,
-		"time to pause between continuous runs",
+		"Amount of time to pause between continuous runs (e.g. 1h, 2m).",
 	)
 	cmd.PersistentFlags().BoolVar(
 		&verifyContinuous,
 		"continuous",
 		false,
-		"whether verification should continuously run on each shard",
+		"Whether verification should continuously run on each shard.",
 	)
 	cmd.PersistentFlags().BoolVar(
 		&verifyLive,
 		"live",
 		false,
-		"enables live mode, which attempts to account for rows that can change in value by retrying them before marking them as an inconsistency",
+		"Enable live mode, which attempts to account for rows that can change in value by retrying them before marking them as inconsistent.",
 	)
 	cmd.PersistentFlags().BoolVar(
 		&verifyRows,
 		"rows",
 		true,
-		"whether rows should be verified (otherwise, performs a more basic schema check)",
+		"If true, verify both the schema (columns, types) and row data. If false, verify only the schema.",
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyLiveVerificationSettings.RunsPerSecond,
 		"live-runs-per-second",
 		verifyLiveVerificationSettings.RunsPerSecond,
-		"if set, sets a maximum number of attempts to reverify per second",
+		"Maximum number of retry attempts per second (live mode only).",
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyLiveVerificationSettings.MaxBatchSize,
 		"live-max-batch-size",
 		verifyLiveVerificationSettings.MaxBatchSize,
-		"maximum number of rows to retry at a time in live verification mode",
+		"Maximum number of rows to retry at a time (live mode only).",
 	)
 	cmd.PersistentFlags().DurationVar(
 		&verifyLiveVerificationSettings.FlushInterval,
 		"live-flush-interval",
 		verifyLiveVerificationSettings.FlushInterval,
-		"maximum amount of time to wait before rows can start being retried",
+		"Maximum amount of time to wait before retrying rows (live mode only).",
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyLiveVerificationSettings.RetrySettings.MaxRetries,
 		"live-retries-max-iterations",
 		verifyLiveVerificationSettings.RetrySettings.MaxRetries,
-		"maximum number of retries live reverification should run before marking rows as inconsistent",
+		"Maximum number of retries before marking rows as inconsistent (live mode only).",
 	)
 	cmd.PersistentFlags().DurationVar(
 		&verifyLiveVerificationSettings.RetrySettings.MaxBackoff,
 		"live-retry-max-backoff",
 		verifyLiveVerificationSettings.RetrySettings.MaxBackoff,
-		"maximum amount of time live reverification should take before retrying",
+		"Maximum amount of time a retry attempt should take before retrying again (live mode only).",
 	)
 	cmd.PersistentFlags().DurationVar(
 		&verifyLiveVerificationSettings.RetrySettings.InitialBackoff,
 		"live-retry-initial-backoff",
 		verifyLiveVerificationSettings.RetrySettings.InitialBackoff,
-		"amount of time live verification should initially backoff for before retrying",
+		"Amount of time live verification should initially backoff for before retrying.",
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyLiveVerificationSettings.RetrySettings.Multiplier,
 		"live-retry-multiplier",
 		verifyLiveVerificationSettings.RetrySettings.Multiplier,
-		"multiplier applied to retry after each unsuccessful live reverification run",
+		"Multiplier to apply to backoff duration after each failed row verification (live mode only).",
 	)
 	for _, hidden := range []string{"fixup", "table-splits"} {
 		if err := cmd.PersistentFlags().MarkHidden(hidden); err != nil {
--- a/cmd/internal/cmdutil/dbconn.go
+++ b/cmd/internal/cmdutil/dbconn.go
@@ -14,13 +14,13 @@ func RegisterDBConnFlags(cmd *cobra.Command) {
 		&DBConnConfig.Source,
 		"source",
 		"",
-		"URL of the source database",
+		"Connection string of the source database.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&DBConnConfig.Target,
 		"target",
 		"",
-		"URL of the target database",
+		"Connection string of the target database.",
 	)
 
 	for _, required := range []string{"source", "target"} {
--- a/cmd/internal/cmdutil/logger.go
+++ b/cmd/internal/cmdutil/logger.go
@@ -18,7 +18,7 @@ func RegisterLoggerFlags(cmd *cobra.Command) {
 		&loggerConfigInst.level,
 		"logging",
 		loggerConfigInst.level,
-		"what level to log at - maps to zerolog.Level",
+		"Level to log at (maps to zerolog.Level).",
 	)
 }
 
--- a/cmd/internal/cmdutil/name_filter.go
+++ b/cmd/internal/cmdutil/name_filter.go
@@ -12,13 +12,13 @@ func RegisterNameFilterFlags(cmd *cobra.Command) {
 		&tableFilter.TableFilter,
 		"table-filter",
 		tableFilter.TableFilter,
-		"POSIX regexp filter for tables to action on",
+		"POSIX regexp filter for tables to action on.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&tableFilter.SchemaFilter,
 		"schema-filter",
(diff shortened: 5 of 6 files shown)


cmd/fetch/fetch.go
```diff
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
...
 		"",
-		"path to upload files to locally",
+		"Path to upload files to locally.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&localPathListenAddr,
 		"local-path-listen-addr",
 		"",
-		"local address to listen to for traffic",
+		"Local address to listen to for traffic.",
```
ZhouXing19, line 171: nit: Wondering if it would be more accurate to say "Address of a local store server to listen for traffic."

cmd/fetch/fetch.go
```diff
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
...
 		false,
-		"whether to truncate the table being imported to",
+		"Whether to truncate the table being imported to.",
 	)
 	cmd.PersistentFlags().IntVar(
 		&cfg.ExportSettings.RowBatchSize,
 		"row-batch-size",
 		100_000,
-		"how many rows to select at a time for the database",
+		"Number of rows to select at a time for the source database.",
```
ZhouXing19, line 189: out of curiosity, does it mean `number of rows to import` as well?

cmd/fetch/fetch.go
```diff
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
...
 		"",
-		"local address to listen to for traffic",
+		"Local address to listen to for traffic.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&localPathCRDBAccessAddr,
 		"local-path-crdb-access-addr",
 		"",
-		"address CockroachDB can access to connect to the --local-path-listen-addr",
+		"Address CockroachDB can access to connect to the --local-path-listen-addr.",
```
ZhouXing19, line 177: nit: personally i found it a bit confusing, but maybe it's because I'm not that familiar with the context. Does it mean `Address of data that CockroachDB can access to import from a local store`?

ZhouXing19 (approved):
LGTM but I'll leave Jeremy and Ryan to finalize the review.

cmd/fetch/fetch.go
```diff
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
...
 		"",
-		"s3 bucket",
+		"Path to the S3 bucket.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&gcpBucket,
 		"gcp-bucket",
 		"",
-		"gcp bucket",
+		"Path to the GCP bucket.",
```
Jeremyyang920, line 153: Same as AWS, but not sure if gcp has a prefix in front.

cmd/fetch/fetch.go
Jeremyyang920: Should it also be `Number of rows to select at a time FROM the source db`?

taroface: If Jane's suggestion is true, then I suggest ```Number of source database rows to import at a time.```

cmd/fetch/fetch.go
```diff
@@ -214,7 +214,7 @@ func Command() *cobra.Command {
 			enumflag.EnumCaseInsensitive,
 		),
 		"compression",
-		"compression to use (default/gzip/none)",
+		"Compression type to use (default/gzip/none).",
```
taroface, line 217: Is this accurate?

```suggestion
		"Compression type to use for data import (default/gzip/none).",
```

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 		0,
-		"if set, maximum number of rows to read per second during scanning per shard",
+		"If set, maximum number of rows to read per second during scanning per shard.",
 	)
 	cmd.PersistentFlags().BoolVar(
 		&verifyFixup,
 		"fixup",
 		false,
-		"whether to fix up any rows",
+		"Whether to fix up any rows.",
```
taroface, line 122: We should probably add more detail on what "fix up" does.

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().DurationVar(
 		&verifyContinuousPause,
 		"continuous-pause-between-runs",
 		0,
-		"time to pause between continuous runs",
+		"Time to pause between continuous runs.",
```
taroface, line 128: What are the units of time?

```suggestion
		"Amount of time to pause between continuous runs.",
```

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().BoolVar(
 		&verifyContinuous,
 		"continuous",
 		false,
-		"whether verification should continuously run on each shard",
+		"Whether verification should continuously run on each shard.",
```
taroface, line 134: Ditto re: shard

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().BoolVar(
 		&verifyLive,
 		"live",
 		false,
-		"enables live mode, which attempts to account for rows that can change in value by retrying them before marking them as an inconsistency",
+		"Enables live mode, which attempts to account for rows that can change in value by retrying them before marking them as an inconsistency.",
```
taroface, line 140:
```suggestion
		"Enable live mode, which attempts to account for rows that can change in value by retrying them before marking them as inconsistent.",
```

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().BoolVar(
 		&verifyRows,
 		"rows",
 		true,
-		"whether rows should be verified (otherwise, performs a more basic schema check)",
+		"Whether rows should be verified (otherwise, performs a more basic schema check).",
```
taroface, line 146: I'm not clear on what the distinction between "verified" and "basic schema check" is.

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyLiveVerificationSettings.RunsPerSecond,
 		"live-runs-per-second",
 		verifyLiveVerificationSettings.RunsPerSecond,
-		"if set, sets a maximum number of attempts to reverify per second",
+		"If set, sets a maximum number of attempts to re-verify per second.",
```
taroface, line 152: I assume this is live mode only.

```suggestion
		"Maximum number of row retry attempts per second (live mode only).",
```

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyLiveVerificationSettings.MaxBatchSize,
 		"live-max-batch-size",
 		verifyLiveVerificationSettings.MaxBatchSize,
-		"maximum number of rows to retry at a time in live verification mode",
+		"Maximum number of rows to retry at a time in live verification mode.",
```
taroface, line 158:
```suggestion
		"Maximum number of rows to retry at a time (live mode only).",
```

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().DurationVar(
 		&verifyLiveVerificationSettings.FlushInterval,
 		"live-flush-interval",
 		verifyLiveVerificationSettings.FlushInterval,
-		"maximum amount of time to wait before rows can start being retried",
+		"Maximum amount of time to wait before rows can start being retried.",
```
taroface, line 164:
```suggestion
		"Maximum amount of time to wait before retrying rows (live mode only).",
```

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyLiveVerificationSettings.RetrySettings.MaxRetries,
 		"live-retries-max-iterations",
 		verifyLiveVerificationSettings.RetrySettings.MaxRetries,
-		"maximum number of retries live reverification should run before marking rows as inconsistent",
+		"Maximum number of retries live re-verification should run before marking rows as inconsistent.",
```
taroface, line 170:
```suggestion
		"Maximum number of retries before marking rows as inconsistent (live mode only).",
```

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().DurationVar(
 		&verifyLiveVerificationSettings.RetrySettings.MaxBackoff,
 		"live-retry-max-backoff",
 		verifyLiveVerificationSettings.RetrySettings.MaxBackoff,
-		"maximum amount of time live reverification should take before retrying",
+		"Maximum amount of time live reverification should take before retrying.",
```
taroface, line 176: Is this still accurate?

```suggestion
		"Maximum amount of time a row retry attempt should take before retrying again (live mode only).",
```

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().DurationVar(
 		&verifyLiveVerificationSettings.RetrySettings.InitialBackoff,
 		"live-retry-initial-backoff",
 		verifyLiveVerificationSettings.RetrySettings.InitialBackoff,
-		"amount of time live verification should initially backoff for before retrying",
+		"Amount of time live verification should initially backoff for before retrying.",
```
taroface, line 182: This probably needs a rephrase, but I don't know what "backoff" means in this context.

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyLiveVerificationSettings.RetrySettings.Multiplier,
 		"live-retry-multiplier",
 		verifyLiveVerificationSettings.RetrySettings.Multiplier,
-		"multiplier applied to retry after each unsuccessful live reverification run",
+		"Multiplier applied to retry after each unsuccessful live re-verification run.",
```
taroface, line 188: What is the multiplier for?

taroface (changes requested):
Left a bunch of copy changes and some questions -- hope it's ok if we take a bit more time with this.

cmd/fetch/fetch.go
ryanluu12345: Got it!

cmd/fetch/fetch.go
ryanluu12345: This is just the export. I'm going to do @Jeremyyang920 's suggestion since I think it captures it more accurately. It's the number of source database rows we are fetching.

cmd/fetch/fetch.go
ryanluu12345: Yeah, pretty much. I'll go with this.

cmd/fetch/fetch.go
```diff
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
...
 		4,
-		"number of tables to move data with at a time",
+		"Number of tables to move at a time.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&s3Bucket,
 		"s3-bucket",
 		"",
-		"s3 bucket",
+		"Path to the S3 bucket.",
```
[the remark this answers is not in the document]
ryanluu12345: Actually now that I think about it, this is the name. I'll make this `Name of the S3 bucket`

cmd/fetch/fetch.go
ryanluu12345: See comment above. But I think it should be `Name of the bucket`

cmd/fetch/fetch.go
```diff
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
 		&directCRDBCopy,
 		"direct-copy",
 		false,
-		"whether to use direct copy mode",
+		"Whether to use direct copy mode.",
```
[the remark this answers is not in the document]
ryanluu12345: Yeah, so direct copy mode copies the data directly from the source database to the target database without using an intermediate store.

cmd/fetch/fetch.go
```diff
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
...
 		false,
-		"whether the table must be queriable during load import",
+		"Whether the table must be query-able during load import.",
 	)
 	cmd.PersistentFlags().IntVar(
 		&cfg.FlushSize,
 		"flush-size",
 		0,
-		"if set, size (in bytes) before the data source is flushed",
+		"If set, size (in bytes) before the data source is flushed.",
```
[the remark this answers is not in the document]
ryanluu12345: Ok this should clarify: `"If set, size (in bytes) before the source data is flushed to intermediate files."`

cmd/fetch/fetch.go
```diff
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().IntVar(
 		&cfg.FlushRows,
 		"flush-rows",
 		0,
-		"if set, number of rows before the data source is flushed",
+		"If set, number of rows before the data source is flushed.",
```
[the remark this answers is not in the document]
ryanluu12345: Look at comment above

cmd/fetch/fetch.go
```diff
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
...
 		"",
-		"gcp bucket",
+		"Path to the GCP bucket.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&bucketPath,
 		"bucket-path",
 		"",
-		"path within the bucket to write intermediate files (i.e. test-migrations/v1 with no ending slash)",
+		"Path within the bucket to write intermediate files (i.e. test-migrations/v1 with no ending slash).",
```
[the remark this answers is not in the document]
ryanluu12345: I'll take your suggestion. It should be clear if people first read the README. The intermediate files are the ones that are written to the local store or S3/GCP

cmd/fetch/fetch.go
```diff
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
...
 		"",
-		"address CockroachDB can access to connect to the --local-path-listen-addr",
+		"Address CockroachDB can access to connect to the --local-path-listen-addr.",
 	)
 	cmd.PersistentFlags().BoolVar(
 		&cfg.Truncate,
 		"truncate",
 		false,
-		"whether to truncate the table being imported to",
+		"Whether to truncate the table being imported to.",
```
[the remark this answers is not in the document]
ryanluu12345: Correct. I'll resolve this.

cmd/fetch/fetch.go
```diff
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
...
 		100_000,
-		"how many rows to select at a time for the database",
+		"Number of rows to select at a time for the source database.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&cfg.ExportSettings.PG.SlotName,
 		"pg-logical-replication-slot-name",
 		"",
-		"if set, the name of a replication slot that should be created before taking a snapshot of data",
+		"If set, the name of a replication slot that should be created before taking a snapshot of data.",
```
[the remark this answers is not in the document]
ryanluu12345: Done

cmd/fetch/fetch.go
ryanluu12345: I think I"ll keep this one as is. This compression happens on the export. The decompression happens on import.

ryanluu12345: Compression is used throughout at different steps.

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 		0,
-		"number of tables to process at a time (defaults to number of CPUs)",
+		"Number of tables to process at a time (defaults to number of CPUs).",
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyTableSplits,
 		"table-splits",
 		1,
-		"number of shards to break down each table into whilst doing row-based verification",
+		"Number of shards to break down each table into while doing row-based verification.",
```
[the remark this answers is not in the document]
ryanluu12345: Looking at the code I think it means how many parts to break a certain table into. So let's say there are 10000 rows and we want to break it down to 10 shards, each shard should be 1000 rows. Not sure how to succinctly explain that though.

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyRowBatchSize,
 		"row-batch-size",
 		20000,
-		"number of rows to get from a table at a time",
+		"Number of rows to select from a table at a time.",
```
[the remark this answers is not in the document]
ryanluu12345: Ok so this one ends up scanning both of the tables. Let me know your thoughts on wording this? Maybe `Number of rows to select from row + target databases at a time.

cmd/verify/verify.go
ryanluu12345: Correct any inconsistencies found from the `verification` run.

cmd/verify/verify.go
ryanluu12345: This can actually be specified as a string, so like `1h`, `2m`, etc.

cmd/verify/verify.go
ryanluu12345: Look at above comments.

cmd/verify/verify.go
ryanluu12345: So basically this option would verify the schema **AND** the data in the rows if it's enabled. In the case that this is set to `false` then only the basic schema checks (matching columns and types for schema) but not the data itself.

cmd/verify/verify.go
ryanluu12345: Good! Done.

cmd/verify/verify.go
ryanluu12345: `Backoff` means how much time it should wait until it does a retry after the initial row verification failure.

cmd/verify/verify.go
ryanluu12345: Based on these lines of code:
```
func (rm *Retry) Next() {
	nextDuration := rm.settings.InitialBackoff * time.Duration(math.Pow(float64(rm.settings.Multiplier), float64(rm.Iteration)))
	if rm.settings.MaxBackoff > 0 && nextDuration > rm.settings.MaxBackoff {
		nextDuration = rm.settings.MaxBackoff
	}
	rm.Iteration++
	rm.NextRetry = rm.NextRetry.Add(nextDuration)
}
```
It appears that the multiplier is for the backoff duration. So basically the amount of time to wait before next retry.

cmd/fetch/fetch.go
```diff
@@ -108,103 +108,103 @@ func Command() *cobra.Command {
...
 		"",
-		"local address to listen to for traffic",
+		"Address of a local store server to listen to for traffic.",
 	)
 	cmd.PersistentFlags().StringVar(
 		&localPathCRDBAccessAddr,
 		"local-path-crdb-access-addr",
 		"",
-		"address CockroachDB can access to connect to the --local-path-listen-addr",
+		"Address of data that CockroachDB can access to import from a local store.",
```
ryanluu12345, line 177: @taroface this could use more workshopping

ryanluu12345: Also not sure if this is used much. I don't see it in the README Oliver wrote. Also the `local-path-listen-addr` is the main one to set up the server to listen.

cmd/verify/verify.go
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 		true,
-		"whether rows should be verified (otherwise, performs a more basic schema check)",
+		"Whether rows should be verified (otherwise, performs a more basic schema check).",
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyLiveVerificationSettings.RunsPerSecond,
 		"live-runs-per-second",
 		verifyLiveVerificationSettings.RunsPerSecond,
-		"if set, sets a maximum number of attempts to reverify per second",
+		"Maximum number of row retry attempts per second (live mode only).",
```
ryanluu12345, line 152: Need to revisit if this actually means `row retry attempts`.

cmd/fetch/fetch.go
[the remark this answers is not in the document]
taroface: ```Enable direct copy mode, which copies data directly from source to target without using an intermediate store.```

cmd/fetch/fetch.go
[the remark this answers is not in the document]
taroface: SGTM!

cmd/fetch/fetch.go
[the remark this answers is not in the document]
taroface: ```If set, number of rows before the source data is flushed to intermediate files.```

cmd/fetch/fetch.go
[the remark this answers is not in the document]
taroface: ```Path within the bucket where intermediate files are written (e.g., bucket-name/v1).```

cmd/verify/verify.go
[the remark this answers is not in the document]
taroface: I'm not sure if this is mixing in/up too much terminology, but this seems to read most clearly to me (feel free to adjust):

```Number of row-based shards into which tables are partitioned for row verification.```

File: cmd/verify/verify.go (hunk 26 above)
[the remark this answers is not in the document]
Comment (taroface) [reply cmd/verify/verify.go:110]: It depends on it's the combined number of source + target rows, or the number of pairs. Neither of these options sounds that elegant to me so feel free to adjust.

```Combined number of source and target rows to scan at a time.```
```Number of pairs of source and target rows to scan at a time.```

File: cmd/verify/verify.go [hunk 29]
```diff
@@ -95,97 +95,97 @@ func Command() *cobra.Command {
...
 		20000,
-		"number of rows to get from a table at a time",
+		"Number of rows to select from a table at a time.",
 	)
 	cmd.PersistentFlags().IntVar(
 		&verifyLimitRowsPerSecond,
 		"rows-per-second",
 		0,
-		"if set, maximum number of rows to read per second during scanning per shard",
+		"If set, maximum number of rows to read per second during scanning per shard.",
```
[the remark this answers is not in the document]
taroface: ```If set, maximum number of rows to read per second on each shard.```

cmd/verify/verify.go
taroface: ```Whether to fix up inconsistencies found during row verification.```

cmd/verify/verify.go
taroface: ```Amount of time to pause between continuous runs (e.g., 1h, 2m).```

cmd/verify/verify.go
taroface: This one LGTM

cmd/verify/verify.go
taroface: ```If true, verify both the schema (columns, types) and row data. If false, verify only the schema.```

cmd/fetch/fetch.go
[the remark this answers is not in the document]
ryanluu12345: Got it!

cmd/fetch/fetch.go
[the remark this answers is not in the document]
ryanluu12345: Opted for this instead: `"Path within the bucket where intermediate files are written (e.g., bucket-name/folder-name)."`

cmd/verify/verify.go
taroface: I am assuming `backoff` is a technical term, so maybe this is enough:

```Amount of time to backoff before retrying row verification (live mode only).```

cmd/verify/verify.go
taroface: ```Multiplier to apply to backoff duration after each failed row verification (live mode only).```

cmd/verify/verify.go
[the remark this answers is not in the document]
ryanluu12345: Yeah this is clear.

cmd/verify/verify.go
ryanluu12345: Done!

cmd/verify/verify.go
taroface: Ah I see what you mean.

```Maximum number of retry attempts per second (live mode only).```

cmd/verify/verify.go
ryanluu12345: Good let's do it this way!

cmd/verify/verify.go
ryanluu12345: Done!

cmd/verify/verify.go
ryanluu12345: Done!

cmd/verify/verify.go
ryanluu12345: Done!

cmd/fetch/fetch.go
taroface: What is the relation of the address to `--local-path-listen-addr`?

cmd/fetch/fetch.go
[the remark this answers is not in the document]
ryanluu12345: Done!

cmd/verify/verify.go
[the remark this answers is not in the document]
ryanluu12345: I'm going to stick with this: `Number of shards to break down each table into while doing row-based verification.`

I believe the edits you gave might be more confusing. Shards should be pretty self explanatory.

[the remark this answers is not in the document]
taroface: NP! I would suggest cutting "down" from that description, but not a big deal.

cmd/fetch/fetch.go
ryanluu12345: It's defaulted to this if we don't specify. So basically we will:
1. Serve at this address
2. Cockroach will try to retrieve from this
3.

taroface (approved):
LGTM!!

Issue #53, opened 2023-11-08
fetch: when GTID table not present, return more helpful error message


ryanluu12345:
**Description**
On fresh run from an AWS RDS instance, the GTID table returns empty set, which leads to this error

**Repro Steps**
Create a new MySQL instance without GTID enabled

**Expected Behavior**
- The absence of this returns a helpful error that tells you that GTID must be enabled.
- Call out in the docs that GTID must be enabled on a MySQL instance before starting.
- For AWS specifically, customers could create a DB parameter group, modify the DB to use the parameter group, and reboot the instance to make it take effect: 
https://repost.aws/questions/QUDLICa_LyR66ELmKxdVEVag/cannot-enable-gtid-mode-in-parameter-group
enforce_gtid_consistency to ON
gtid_mode to ON

PR #70, opened 2023-11-14
fetch: added clearer help text when MySQL gtid_executed table not set up

ryanluu12345:
Now prompts the user the check that GTID-based replication is enabled before running with MOLT fetch.

Resolves: #53 
Release Note: None

### DIFF
--- a/fetch/dataexport/mysql.go
+++ b/fetch/dataexport/mysql.go
@@ -12,6 +12,8 @@ import (
 	"github.com/cockroachdb/molt/rowiterator"
 )
 
+const GTIDHelpInstructions = `please ensure that you have GTID-based replication enabled`
+
 type mysqlSource struct {
 	gtid     string
 	settings Settings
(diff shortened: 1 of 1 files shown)


Jeremyyang920 (approved):

ZhouXing19 (approved):

fetch/dataexport/mysql.go
```diff
@@ -27,7 +29,7 @@ func NewMySQLSource(
 		if err := conn.QueryRowContext(ctx, "select source_uuid, min(interval_start), max(interval_end) from mysql.gtid_executed group by source_uuid").Scan(
 			&source, &start, &end,
 		); err != nil {
-			return errors.Wrap(err, "failed to export snapshot")
+			return errors.Wrapf(err, "failed to export snapshot: %s", GTIDHelpInstructions)
```
ZhouXing19, line 32: nit: could use a more standard `errors.WithHint(err, GTIDHelpInstructions)`

ryanluu12345: I tried: `return errors.WithHintf(err, "failed to export snapshot: %s", GTIDHelpInstructions)` but it returns this:
```
Error: done
```
I'll just keep this.

PR #72, opened 2023-11-14
demo: LMS docker demo

Jeremyyang920:
This commit adds the start to what could be the mega demo. It spins up the LMS with 3 nodes and the respective databases to do what the k8s demo does.

### DIFF
--- a/demo/docker-compose.yml
+++ b/demo/docker-compose.yml
@@ -1,0 +1,141 @@
+version: "3"
+
+services:
+  lms:
+    image: cockroachdb/molt-lms:latest
+    restart: always
+    working_dir: /app
+    ports:
+      - "9046-9048:9043"
+      - "9044"
+    command: 
+      - mysql 
+      - --init-source=mysql://root:password@mysql:3306/defaultdb
+      - --init-target=postgresql://root@crdb:26257/defaultdb?sslmode=disable
+      - --shadow-mode=none 
+    depends_on:
+      - mysql
+      - postgres
+      - crdb
+      - lms-orchestrator
+  
+  lms-orchestrator:
+    image: cockroachdb/molt-lms:latest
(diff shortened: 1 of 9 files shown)


ZhouXing19 (approved):

ryanluu12345 (approved):
Very solid work! Let's get this in!

demo/docker-compose.yml
```diff
@@ -0,0 +1,141 @@
...
+    ports:
+      - "4200:4200"
+      - "4201"
+    entrypoint: ./molt-lms-orchestrator
+    environment:
+      - ORCH_ALLOW_ORIGIN=/.*/
+    command: 
+     --source-dialect=mysql
+     --target-dialect=crdb
+     --lms-url=mysql://root:password@demo-lms-1:9043/defaultdb,mysql://root:password@demo-lms-2:9043/defaultdb,mysql://root:password@demo-lms-3:9043/defaultdb
```
ryanluu12345, line 35: Quick clarifying question: does these comma separated values store a slice of values for `LMS URLs`? So does this mean when we initiate cutover or list connections we don't need to specify them from the command line, but rather it will just use these instances?

demo/docker-compose.yml
```diff
@@ -0,0 +1,141 @@
...
+     --crdb-url=postgresql://root@crdb:26257/defaultdb?sslmode=disable
+  
+  workload:
+    image: jyangcrl/workload:latest
```
ryanluu12345, line 39: Nit: we should most likely move this over to a Cockroach owned image soon.

demo/docker-compose.yml
Jeremyyang920: Didn't you write this section of the code? :p 
Yea you don't need to specify any urls when using the cli.

PR #73, opened 2023-11-15
docs/RFCS: added template for RFCs and initial MOLT fetch RFC

ryanluu12345:
Resolves: #71
Release Note: None

### DIFF
--- a/docs/RFCS/20231113_molt_fetch.md
+++ b/docs/RFCS/20231113_molt_fetch.md
@@ -1,0 +1,223 @@
+- Feature Name: MOLT Fetch
+- Status: in-progress
+- Start Date: 2023-11-13
+- Authors: Ryan Luu
+- RFC PR: 73
+- Ticket: https://github.com/cockroachdb/molt/issues/71
+
+# Summary
+
+This RFC presents the case for a MOLT tool (`fetch`) which can fetch data from a source database and import the data into Cockroach. Compared to third-party migration tools, `fetch` will provide better observability, debugability, and control over performance optimizations. The initial RFC is scoped to MySQL and PostgreSQL source databases and the initial data load action. Later on, we plan to support streaming updates by integrating with tools such as `cdc-sink`. However, that will be scoped to a later RFC.
+
+The high level plan is to build out this functionality as part of the existing `molt` tool as a subcommand. The two key features that need to be built out are the ability to export/dump data from the source, and run the appropriate `COPY` or `IMPORT` command on Cockroach.
+
+# Motivation
+
+The data movement portion of a migration is a crucial step that every customer goes through when migrating to Cockroach. Although there are third party tools currently available for this task, there is currently limited observability into these tools and how exactly they interact with Cockroach. By building the tooling in-house, we control the logging, metrics, and way we surface errors and anomalies. Additionally, we can control how `fetch` interacts with Cockroach, which will help to address compatibility and performance issues.
+
+# Detailed design
+
+`molt fetch` enables an initial dump of bulk data from source databases to target databases. This is supported in two general modes: one that takes the target table offline so that an `IMPORT INTO` can be performed. And another, where the target table stays online and a `COPY FROM` is used.
+
+`fetch` will output a `cdc_cursor` which can be fed to third-party tools to migrate live data without taking your database offline.
+
+`fetch` will currently support the following:
+
+- Pulling a table, uploading CSVs to S3/GCP/local machine (`--listen-addr` must be set) and running IMPORT on Cockroach for you.
+- Pulling a table, uploading CSVs to S3/GCP/local machine and running COPY TO on Cockroach from that CSV.
+- Pulling a table and running COPY TO directly onto the CRDB table without an intermediate store.
+
+By default, data will be imported using `IMPORT INTO`. You can use the `live` mode if you need target data to be query-able during loading, which will use `COPY FROM` instead.
+
+For now, schemas must be manually created and identical on both sides. This is verified upfront - tables with mismatching columns may only be partially migrated. Partially migrated means that only matching columns will be migrated; mismatching columns will be skipped. Longer term, we will enable automatic schema creation for the target. This feature will be detailed in a follow-up RFC.
+
+```mermaid
+flowchart LR
+    LegacyDB[legacy database<br/>i.e. PG, MySQL]
+    S3[(Amazon S3)]
+    GCP[(Google Cloud<br/>Bucket)]
+    Local[(Local File Server)]
+    CRDB[CockroachDB]
+
+    LegacyDB -- CSV Dump --> S3
+    LegacyDB -- CSV Dump --> GCP
+    LegacyDB -- CSV Dump --> Local
+    S3 -- IMPORT INTO<br/>or COPY FROM --> CRDB
+    GCP -- IMPORT INTO<br/>or COPY FROM --> CRDB
+    Local -- "IMPORT INTO (exposes HTTP server)"<br/>or COPY FROM --> CRDB
+
+    LegacyDB -- COPY FROM --> CRDB
+```
+
+## Data export
+
+`fetch` will initiate a CSV dump by going through each table in the source schema. This initial dump phase can support both compressed (.gzip, .bzip) and uncompressed types.
+
+The targets of this dump includes:
+
+- AWS S3 buckets
+- GCP GCS buckets
+- Local file server (on the executor host)
+
+In order to target the buckets, the user can pass in the relevant bucket and sub-path details and provide their cloud identification details via the environment.
+
+We have two options in how we implement the data export:
+
+- Running a direct COPY with a target of CSV or compressed CSV files: easier to implement, but slower
+- Selecting all tables and paginating through all of the data: more complexity, fine grained control, more performant
+
+For the initial version, using `COPY` to export data works fine, however we want to move to a model where we can paginate through all the data by selecting data from each of the tables because it allows more flexibility in how we filter through data.
+
+In the case that a `COPY FROM` or `IMPORT INTO` is run and dumps to intermediary files in a store (i.e. AWS S3 or GCP GCS), the data travels from the source database to the machine which is running `fetch`, then to the intermediary store, and then finally to the target.
+
+In the case that a `COPY` is issued from the source to the target directly, the data is fetched from the source database to the machine that is running `fetch` in memory, formatted in a manner suitable for `COPY` to Cockroach by `fetch`, and copied over to the target.
+
+### Commands used for dumping for each dialect
+
+For the initial scope, only `Cockroach`, `PostgreSQL` and `MySQL` are supported.
+
+Cockroach: uses `SELECT` to scan and pagniate through all the data in each table, and write them out to a CSV file.
+
+MySQL: uses `SELECT` to scan and pagniate through all the data in each table, and write them out to a CSV file.
+
+PostgreSQL: uses `Copy To` to dump the data into a CSV file; later on, use the `SELECT` scan/pagination method to improve performance and provide table-granularity parallelization.
+
+### Limitations
+
+The data export process is primarily limited by the dialects supported (mentioned above) and the machine that `fetch` is running on. The number of CPUs on the machine will affect the speed of the export for schemas with many tables, because it constrains the amount of concurrency `fetch` can do. The amount of disk storage will affect the case where intermediate files are used on the machine that `fetch` is running on. This limitation is mitigated if cloud storage (GCS or S3) are used. The amount of RAM will affect the case where `direct copy` mode is used since the data is kept in memory before it's copied over to the target database.
+
+## Data import
+
+### Options
+
+Data can be imported into Cockroach in one of three ways:
+
+1. `IMPORT INTO` from intermediary store to Cockroach:
+
+- Target table is taken offline
+- Supports compression (recommended option because `IMPORT INTO` is much faster than copy, especially with compression).
+
+Source -> Fetch -> Intermediary store -> Target
+
+2. `COPY FROM` (`live` mode) from intermediary store to Cockroach:
+
+- Table is still online and queryable; the intermediary store is used, instead of directly from the database
+- Compression can't be supported here because `COPY FROM` currently doesn't have a `decompress` feature
+
+Source -> Fetch -> Intermediary store -> Target
+
+3. Direct `COPY FROM` source database to Cockroach without intermediary store
+
+- Table is still online and query-able
+- Not recommended for very large table migrations
+- Data is copied from the source to target without intermediary store or files
+- Data still passes from the source to `fetch`, but in memory; this means that the RAM specs of the execution machine must be appropriately sized
+- Fetch is coordinating the `COPY FROM` and holds access to the DB connections; this is the most direct path `fetch` can support
+
+Source -> Fetch (in memory) -> Target
+
+### Testing for import into vs. copy
+
+For Cockroach, `IMPORT INTO` is much faster, as noted above.
+
+Previously, our team performed baseline testing on `IMPORT INTO` vs. `COPY`:
+For 10M rows:
+
+- `IMPORT INTO` (Postgres -> S3 CSVs -> Cockroach) took 90s; 30s to upload the files to S3 and 60s to ingest it.
+- `COPY` (Postgres output directly into Cockroach input) took around 350s (~6 mins) using 4MB COPY chunks.
+
+Follow up parameter tuning notes:
+
+- Using 1MB `COPY` chunks tooks 5mins instead of 6mins from PG->CRDB
+- Using `COPY` to migrate from Postgres -> Postgres took 120s using 1MB COPY chunks and 95s using 4MB chunks
+
+Suffice to say, `IMPORT INTO` in Cockroach is comparable to Postgres `COPY`.
+
+For more baseline testing of these modes of operation, check out the scale testing [section](#scale-baseline-testing).
+
+## Integrating with replication tools
+
+As part of the initial scope, a `cdc_cursor` will be outputted at the end of the `fetch` task so that the user can feed that into the replication tool of their choice. The limitiation here is that this will be a manual process that the user will have to take.
+
+As part of a future RFC, automatic integration with other replication tools will be considered and evaluated, especially `cdc-sink`.
+
+## Handling failures
+
+The two primary types of failures that will show up are:
+
+1. Row level data failures: inconsistent or missing data or unsupported conversions
+
+- Log out to a separate log file with clear explanation of what failed and why it failed
+- Persist these failures in an exceptions log table
+- Output a recommendation to run `molt verify` after the fact to see which data is inconsistent
+
+2. Task level failures: job failed to start, dies midway through, or runs into stability issues while running
+
+- To handle these failures, we can output a structured log line the task log to let the future run know where the previous run failed. We can then pick up from that point if it's possible, or retry from the start.
+- We can also issue automatic retries if there are recoverable failures during the task.
+- If the task itself dies or fails to start, we should log a fatal error that gives clear, actionable steps about how to resolves.
+
+Generally for known error types in the system, there will be a flag to configure how `fetch` should behave. For example, being able to specify that for data errors, `fetch` should just log and continue; or for `fetch` to not tolerate any errors and abort execution if any error is hit.
+
+## Drawbacks
+
+By bringing tooling development in house, we will need to invest time to build out feature support that is on par with the existing tools on the market. Then after that, we have the maintenance cost for our tools, while also providing support for existing third party tools.
+
+## Rationale and alternatives
+
+### Alternatives considered
+
+1. Private labeling existing tools
+   This was strongly considered, but private labeling would leave us with the same lack of visibility into tooling. Additionally it would make it harder to iterate and fix gaps for customers because the source is controlled by third-parties.
+
+2. Supporting third party tools
+   This is something the team already supports and will continue to support. However, this is the approach that gives us the least visibility.
+
+# Testing plan
+
+## Scale baseline testing
+
+- Uses the 200M row workload in AWS RDS that we test DMS with
+- Will run the `fetch` process on a Roachprod AWS machine
+- Will measure the amount of time the export and import takes
+- Will baseline against all three modes of operation (see above)
+
+1. AWS DMS test (uses `COPY`): 33 minutes in nightly tests; 18 minutes in preliminary manual testing with parallel individual table loads (future optimization `fetch` can make as well)
+
+2. `IMPORT INTO` with intermediate store and compression: 26 minutes; 14.5 minutes for export, 11.5 minutes for import
+
+3. `COPY FROM` with intermediate store: 2 hours 23 minutes; 7 minutes 29 seconds for export, 2 hrs 15 minutes 52 seconds for import
+
+4. `COPY FROM` directly without intermediate store: 47 minutes
+
+There are still areas of improvement to the testing process and how we run `fetch` during our baselining tests. Some future areas of exploration are to test with:
+
+- Different number of vCPUs
+- Different specs for storage (RAM and disk)
+- Using a serverless cluster as target instead of Roachprod AWS instance
+- Running with multiple nodes
+- And more to come...
+
+## Randomized schema testing
+
+- Generates random schemas and data of varying types on the source
+- Will run `fetch` on the data to migrate the data to the target
+- Will check the correctness of the data
+
+## E2E testing
+
+- Automates the scale testing at varying scales (small/medium/large workloads)
+- A variation of the tests will run on each: PR, post-merge to main, release, nightly
+
+## Replication enabled testing
+
+- Verify that streaming updates from replication make their way properly to the target
+- Can automate starting up a replication tool after initial data load and running for a period of time
+
+For all of the aforementioned tests, `molt verify` will be used to ensure correctness of the schema and rows.
+
+# Unresolved questions
+
+- How do we plan on handling matching schema creation on the target DB?
+- What does extending dialect support look like?
+- How can we integrate this into other surface areas of our product (i.e. cloud managed-service)?
--- a/docs/RFCS/00000000_template.md
+++ b/docs/RFCS/00000000_template.md
@@ -1,0 +1,46 @@
+- Feature Name:
+- Status: draft/in-progress/completed/rejected/obsolete/postponed
+- Start Date: YYYY-MM-DD
+- Authors:
+- RFC PR: (PR # after acceptance of initial draft)
+- Ticket:
+
+**Remember, you can submit a PR with your RFC before the text is
+complete. Refer to the [README](README.md#rfc-process) for details.**
+
+**Remember, you can either fill in this template from scratch, for
+example if you prefer working from a blank slate, or you can follow
+the writing prompts in the [GUIDE](GUIDE.md). In any case, please ensure
+at the end that you have all relevant topics from the guide covered in
+your prose.**
+
+# Summary
+
+- What is being proposed
+- Why (short reason)
+- How (short plan)
+- Impact
+
+# Motivation
+
+Audience: PMs, end-users, CockroachDB team members.
+
+# Technical design
+
+Audience: CockroachDB team members, expert users.
+
+## Drawbacks
+
+...
+
+## Rationale and Alternatives
+
+...
+
+# Explain it to folk outside of your team
+
+Audience: PMs, doc writers, end-users, CockroachDB team members in other areas of the project.
+
+# Unresolved questions
+
+Audience: all participants to the RFC review.


docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,112 @@
...
+# Motivation
+
+The data movement portion of a migration is a crucial step that every customer goes through when migrating to Cockroach. Although there are third party tools currently available for this task, there is currently limited observability into these tools and how exactly they interact with Cockroach. By building the tooling in-house, we can control how the tool interacts with Cockroach to address compatibility and/or performance issues.
+
+# Detailed Design
+
+The tool will migrate data from PG or MySQL tables to CockroachDB
+without taking your PG/MySQL tables offline.
+
+It will outputa a `cdc_cursor` which can be fed to CDC programs (e.g. cdc-sink, AWS DMS)
```
ZhouXing19, line 23: nit: outputa -> output

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,112 @@
...
+
+- Running a direct COPY with a target of CSV or compressed CSV files
+- Selecting all tables and paginating through all of the data
+
+## Data import
+
+Data can be imported into Cockroach in one of three ways:
+
+1. `COPY FROM` source database to Cockroach directly: table is still online and query-able; data is copied directly instead of dumped to an intermediary location.
+2. `IMPORT INTO` from intermediary store to Cockroach: table is offline; can improve performance by using compression.
```
ZhouXing19, line 77: Do we want to mention the possible benefit of `IMPORT INTO`? From this paragraph it seems an inferior option compared to COPY. There might have been docs comparing IMPORT and COPY and I think it's helpful if we can link them here.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,112 @@
...
+3. Build our own tooling
+   This gives us control on how the data movement tooling interacts with the source and target databases, which will allow us to more easily respond to customers needs for moving data into Cockroach.
+
+# Explain it to folk outside of your team
+
+This tool helps customers to move data from their own databases to Cockroach.
+
+# Unresolved questions
+
+- How do we plan on handling matching schema creation on the target DB?
```
ZhouXing19, line 109: Should we include plan for benchmarking / performance / correctness tests?

docs/RFCS/20231113_molt_fetch.md
ZhouXing19, line 23: Would be beneficial if we can explain or link docs about the cdc_cursor.

ZhouXing19, line 23: I found this sentence a bit disconnected from the context. Would it be more accurate to put something like: `molt fetch enables an initial dump of bulk data and streaming of updates happening on the source database after the dump is initiated.`

ZhouXing19 (approved):
LGTM but I'll leave Adam and Jeremy to finalize the review.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,112 @@
+- Feature Name: MOLT Fetch
+- Status: draft/in-progress/completed/rejected/obsolete/postponed
+- Start Date: 2023-11-13
+- Authors: Ryan Luu
+- RFC PR: 73
+- Ticket: https://github.com/cockroachdb/molt/issues/71
+
+# Summary
+
+This RFC presents the case for a MOLT tool which can fetch data from a source database and import the data into Cockroach. This allows for better observability and control over performance optimizations into Cockroach, compared to third-party migrations tools. The initial RFC is scoped to MySQL and PostgreSQL source databases and the initial data load action.
```
Jeremyyang920, line 10: I know what "This allows for better observability and control over performance optimizations into Cockroach, compared to third-party migrations tools." is trying to say, but it's a bit difficult to read as is.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,112 @@
...
+
+The high level plan is to build out this functionality as part of the existing `molt` tool as a subcommand. The two key features that need to be built out are the ability to export/dump data from the source, and run the appropriate `COPY` or `IMPORT` command on Cockroach.
+
+# Motivation
+
+The data movement portion of a migration is a crucial step that every customer goes through when migrating to Cockroach. Although there are third party tools currently available for this task, there is currently limited observability into these tools and how exactly they interact with Cockroach. By building the tooling in-house, we can control how the tool interacts with Cockroach to address compatibility and/or performance issues.
+
+# Detailed Design
+
+The tool will migrate data from PG or MySQL tables to CockroachDB
```
Jeremyyang920, line 20: Not a big deal, but we should maybe standardize on the naming when writing. There are instances where its referenced as "The tool" and others its "it". Maybe we should reference these all as Fetch to avoid any ambiguous interpretation.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Noted.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: I was looking at the docs and couldn't find a clear comparison between `IMPORT INTO` and `COPY FROM`. What makes `IMPORT INTO` so much more performant? I reordered so that `IMPORT INTO` is the first option and marked as a default.

ZhouXing19: Oh i don't think we need to go into tech details about why import is more performant than copy, but just wanted to note that we can mention IMPORT is fasrer then COPY: https://www.cockroachlabs.com/docs/stable/import-performance-best-practices#:~:text=IMPORT%20INTO%20is,COPY%20FROM%20instead.

ZhouXing19: nit: Also, iiuc, for `IMPORT INTO`, it's the _target_ table that is offline. Might worth noting that this is just the table on the crdb, while the table on the source db never goes offline.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Yes, including a new section for this.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Hmm I think that this would be a term we can assume our audience knows. Given that `CDC` cursor is a concept not specific to this tool, but for database changefeeds and tracking.

@taroface @ajstorm please let me know if this is a fair assumption to make though.

ryanluu12345: I merged the two sentences (original and your suggestion):
```
`molt fetch` enables an initial dump of bulk data from source databases to target databases, without taking source tables offline. The scope of the initial design will support PostgreSQL and MySQL source databases.

`fetch` will output a `cdc_cursor` which can be fed to CDC programs (e.g. cdc-sink, AWS DMS)
to migrate live data without taking your database offline.
```

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Did this instead:
```
Compared to third-party migration tools, `fetch` will provide better observability, debugability, and control over performance optimizations
```

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Yeah, much better to standardize like this!

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Ok think I addressed with this: 
```
1. `IMPORT INTO` from intermediary store to Cockroach: target table is taken offline; can improve performance by using compression (recommended option because `IMPORT INTO` is fast, especially with compression).
```

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: @ZhouXing19 let me know if this is enough detail? Look at #Testing Plan

ZhouXing19: LGTM! Thanks

docs/RFCS/20231113_molt_fetch.md
taroface: @ryanluu12345 Your comment makes sense to me. BTW, there's a typo on `outputa`.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
+- Feature Name: MOLT Fetch
+- Status: draft/in-progress/completed/rejected/obsolete/postponed
```
ajstorm, line 2: Nit: in-progress?

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+- Start Date: 2023-11-13
+- Authors: Ryan Luu
+- RFC PR: 73
+- Ticket: https://github.com/cockroachdb/molt/issues/71
+
+# Summary
+
+This RFC presents the case for a MOLT tool (`fetch`) which can fetch data from a source database and import the data into Cockroach. Compared to third-party migration tools, `fetch` will provide better observability, debugability, and control over performance optimizations. The initial RFC is scoped to MySQL and PostgreSQL source databases and the initial data load action.
+
+The high level plan is to build out this functionality as part of the existing `molt` tool as a subcommand. The two key features that need to be built out are the ability to export/dump data from the source, and run the appropriate `COPY` or `IMPORT` command on Cockroach.
```
ajstorm, line 12: Nit: Should we mention cdc-sink here?

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+
+# Motivation
+
+The data movement portion of a migration is a crucial step that every customer goes through when migrating to Cockroach. Although there are third party tools currently available for this task, there is currently limited observability into these tools and how exactly they interact with Cockroach. By building the tooling in-house, we can control how the `fetch` interacts with Cockroach to address compatibility and performance issues.
```
ajstorm, line 16: Nit: The second and third sentences here don't perfectly match up.  The second talks about observability problems, while the third talks about compatibility and performance.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+
+# Detailed Design
+
+`molt fetch` enables an initial dump of bulk data from source databases to target databases, without taking source tables offline.
```
ajstorm, line 20: Nit: It actually provides two mode, one of which _does_ take the table offline.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+
+`fetch` will output a `cdc_cursor` which can be fed to CDC programs (e.g. cdc-sink, AWS DMS)
```
ajstorm, line 22: Nit: I'd avoid mentioning third-party tools by name here.  We can just say "third-party tools".

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+to migrate live data without taking your database offline.
+
+`fetch` will currently support the following:
+
+- Pulling a table, uploading CSVs to S3/GCP/local machine (`--listen-addr` must be set) and running IMPORT on Cockroach for you.
+- Pulling a table, uploading CSVs to S3/GCP/local machine and running COPY TO on Cockroach from that CSV.
+- Pulling a table and running COPY TO directly onto the CRDB table without an intermediate store.
+
+By default, data will be imported using `IMPORT INTO`. You can use the `live` mode if you need target data to be query-able during loading, which will use `COPY FROM` instead.
```
ajstorm, line 31: Nit: Above we say the table is not taken offline and here we say that the default _is_ to take it offline.  We should try to be consistent.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+
+For now, schemas must be identical on both sides. This is verified upfront -
```
ajstorm, line 33: Should we talk about the fact that schemas must be manually created?  Also, that longer term we want them to be created manually?

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+    LegacyDB -- CSV Dump --> GCP
+    LegacyDB -- CSV Dump --> Local
+    S3 -- IMPORT INTO<br/>or COPY FROM --> CRDB
+    GCP -- IMPORT INTO<br/>or COPY FROM --> CRDB
+    Local -- "IMPORT INTO (exposes HTTP server)"<br/>or COPY FROM --> CRDB
+
+    LegacyDB -- COPY FROM --> CRDB
+```
+
+## Data export
```
ajstorm, line 54: In this section it would be good to discuss exactly what happens under the covers.  Specifially, that the data always flows to the machine which is running `fetch` before making it to its final destination.  This is an important detail.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+The dump of the data can be achieved by two main methods:
+
+- Running a direct COPY with a target of CSV or compressed CSV files
+- Selecting all tables and paginating through all of the data
+
+## Data import
+
+Data can be imported into Cockroach in one of three ways:
+
+1. `IMPORT INTO` from intermediary store to Cockroach: target table is taken offline; supports compression (recommended option because `IMPORT INTO` is much faster than copy, especially with compression).
```
ajstorm, line 75: It might be good if we had some performance numbers in here.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+2. `COPY FROM` (`live` mode) from intermediary store to Cockroach: table is still online and queryable; the intermediary store is used, instead of directly from the database; compression can't be supported here because `COPY FROM` currently doesn't have a `decompress` feature
+3. `COPY FROM` source database to Cockroach directly: table is still online and query-able; data is copied directly instead of dumped to an intermediary location.
```
ajstorm, line 77: We should be clear about what we mean by "copied directly".  The data is flowing from the source to the `fetch` machine, and then onto the target.  It's only "direct" from the perspective of disk.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Drawbacks
+
+By bringing tooling development in house, we will need to invest time to build out feature support that is on par with the existing tools on the market. Then after that, we have the maintenance cost for our tools, while also providing support for existing third party tools.
+
+`fetch` will also need a strong set of initial users to give the proper customer-facing evaluation necessary to GA. This may also not be as trusted from the outset because `fetch` is newer and not as well tested as existing solutions, such as AWS DMS, Qlik, or Striim.
```
ajstorm, line 83: I'd remove this part.  I don't think that it's necessary.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Rationale and Alternatives
+
+### Decision
+
+Decided to build out or own tooling because of aforementioned reasons: better visibility/observability, performance improvements, and compatibility.
```
ajstorm, line 89: I'd also remove this section

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Alternatives Considered
+
+1. Private labeling existing tools
+   This was strongly considered, but private labeling would leave us with the same lack of visibility into tooling and supporting changes to source code to provide better compatibility/perf for Cockroach.
+
+2. Supporting third party tools
+   This is something the team already supports and will continue to support. However, this is the approach that gives us the least visibility.
+
+3. Build our own tooling
```
ajstorm, line 99: Is this really an alternative?  I thought that this is what we were proposing here.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+   This gives us control on how the data movement tooling interacts with the source and target databases, which will allow us to more easily respond to customers needs for moving data into Cockroach.
+
+# Testing Plan
```
ajstorm, line 102: We should have some replication-enabled testing too.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,128 @@
...
+
+- Automates the scale testing at varying scales (small/medium/large workloads)
+- A variation of the tests will run on each: PR, post-merge to main, release, nightly
+
+For all of the aforementioned tests, `molt verify` will be used to ensure correctness of the schema and rows.
+
+# Unresolved questions
+
+- How do we plan on handling matching schema creation on the target DB?
+- How do we support other replication tools?
```
ajstorm, line 126: I'd remove this one, as it's resolved.

ajstorm (changes requested):
This is a great start.  My main comment is that the "Detailed Design" section is pretty light on internal info.  For example:

- What commands are we using for dumping from each of the dialects?
- What limitations are there on dumping files (if any)?
- What failure modes do we need to consider (i.e. what if it dies mid-dump?  mid-load?)
- How does the integration with replication work?  Are there any limitations there?
- ...

We should have enough detail in here that someone unfamiliar with the tool will know how it was built, the tradeoffs we made, and what remains to be done.

docs/RFCS/20231113_molt_fetch.md
Jeremyyang920: @ajstorm You mean longer term we want them automatically right?

ajstorm: Oops, yeah, automatically.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Done (will push in a bit)

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Thoughts?
```
Later on, we plan to support streaming updates by integrating with tools such as `cdc-sink`. However, that will be scoped to a later RFC.
```

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: This is actually just confusing wording on my end. Since the source tables technically do stay online. The target tables are the ones that don't necessarily. Changed to this. Longer but more explicit:
```
`molt fetch` enables an initial dump of bulk data from source databases to target databases. This is supported in two general modes: one that takes the target table offline so that an `IMPORT INTO` can be performed. And another, where the target table stays online and a `COPY FROM` is used.
```

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Done!

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Done.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Good point. Removing this from the section.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Done!

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Done!

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Yep, agreed. I added a parallel sentence after and then made sure to mention concerns about compatibility and performance as a follow up point.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: I think this is resolved now that I don't even mention the bit about not taking things offline. I think I more clearly mention that we support different modes where the target table can be taken offline or remain online.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Understood. Made a point about this. As discussed offline, referencing this feature as something that will be explored in a future RFC.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: I understand what you're saying here. For the case where we do a fetch that dumps to intermediary files using `COPY FROM` or `IMPORT INTO` this is true. However, we also support direct copy from the source to the target.

ryanluu12345: I noted this!

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Hmm but the `COPY FROM` source to Cockroach directly should be direct right? Since we are issuing a copy from the source to target directly. Unless I'm misunderstanding how this works?

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Noted.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: This is an outdated diff. Resolved now.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: @ajstorm how are these numbers? https://cockroachlabs.slack.com/archives/C02TLESRWCB/p1690516707529709

Oliver already had done some testing. So we could probably use this. Additionally, I can also note how I can do this testing by benchmarking with the latest scale testing.

docs/RFCS/20231113_molt_fetch.md
ajstorm: SGMT

docs/RFCS/20231113_molt_fetch.md
ajstorm: I think it's fine to lean on those results, but we may also want to mention some end-to-end Fetch results which do the pulling and the IMPORT/COPY.

docs/RFCS/20231113_molt_fetch.md
ajstorm: That may be the case.  I thought that when we met in person you said that flowing directly from the source to target was not possible?  Or maybe it's only possible when there's a "direct" fetch?  Would be worth confirming.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,125 @@
...
+
+The targets of this dump includes:
+
+- AWS S3 buckets
+- GCP GCS buckets
+- Local file server (on the executor host)
+
+In order to target the buckets, the user can pass in the relevant bucket and sub-path details and provide their cloud identification details via the environment.
+
+The dump of the data can be achieved by two main methods:
```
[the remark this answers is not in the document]
ryanluu12345: `The user has the option to dump the data in two different ways:`

Basically, one takes the **target** table offline (`IMPORT INTO`). The other keeps the target table online (`COPY FROM`)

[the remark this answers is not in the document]
ryanluu12345: Oh sorry, scratch that. This is a different context. Basically, these are the two methods we are considering under the hood.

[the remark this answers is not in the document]
ryanluu12345:
```
We have two options in how we implement the data export:

- Running a direct COPY with a target of CSV or compressed CSV files: easier to implement, but slower
- Selecting all tables and paginating through all of the data: more complexity, fine grained control, more performant

For the initial version, using `COPY` directly works fine, however we want to move to a model where we can paginate through all the data by selecting data from each of the tables.
```

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,125 @@
...
+    S3 -- IMPORT INTO<br/>or COPY FROM --> CRDB
+    GCP -- IMPORT INTO<br/>or COPY FROM --> CRDB
+    Local -- "IMPORT INTO (exposes HTTP server)"<br/>or COPY FROM --> CRDB
+
+    LegacyDB -- COPY FROM --> CRDB
+```
+
+## Data export
+
+`fetch` will initiate a CSV dump by going through each table in the source table. This initial dump phase can support both compressed (.gzip, .bzip) and uncompressed types.
```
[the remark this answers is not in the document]
ryanluu12345: Exactly, good catch. Will update to `source schema`.

ryanluu12345:
Addressed this too: https://github.com/cockroachdb/molt/pull/73#discussion_r1402381950

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,125 @@
...
+In order to target the buckets, the user can pass in the relevant bucket and sub-path details and provide their cloud identification details via the environment.
+
+The dump of the data can be achieved by two main methods:
+
+- Running a direct COPY with a target of CSV or compressed CSV files
+- Selecting all tables and paginating through all of the data
+
+In the case that a `COPY FROM` or `IMPORT INTO` is run and dumps to intermediary files, the data travels from the source database to the machine which is running `fetch` and then finally to its destination.
+
+In the case that a `COPY` is issued from the source to the target directly, the data goes directly from source to target.
```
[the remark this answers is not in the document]
ryanluu12345: Tracing the code, I see the following:
1. Source gets initialized. Note that `conns[1]`  is passed in. This relates to the target conn.
```
case directCRDBCopy:
				src = datablobstorage.NewCopyCRDBDirect(logger, conns[1].(*dbconn.PGConn).Conn)
```
2. This source's reader seems to run a `CopyFrom` the source table:
```
func (c *copyCRDBDirect) CreateFromReader(
	ctx context.Context, r io.Reader, table dbtable.VerifiedTable, iteration int, fileExt string,
) (Resource, error) {
	conn, err := pgx.ConnectConfig(ctx, c.target.Config())
	if err != nil {
		return nil, err
	}
	c.logger.Debug().Int("batch", iteration).Msgf("csv batch starting")
	if _, err := conn.PgConn().CopyFrom(ctx, r, dataquery.CopyFrom(table)); err != nil {
		return nil, errors.CombineErrors(err, conn.Close(ctx))
	}
	c.logger.Debug().Int("batch", iteration).Msgf("csv batch complete")
	return nil, conn.Close(ctx)
}
```
3. In the action `export` fetch command, this is the logic that executes that uses `CreateFromReader`
```
pipe := newCSVPipe(sqlRead, logger, cfg.FlushSize, cfg.FlushRows, func() io.WriteCloser {
		resourceWG.Wait()
		forwardRead, forwardWrite := io.Pipe()
		wrappedWriter := getWriter(forwardWrite, cfg.Compression)
		resourceWG.Add(1)
		go func() {
			defer resourceWG.Done()
			itNum++
			if err := func() error {
				resource, err := datasource.CreateFromReader(ctx, forwardRead, table, itNum, importFileExt)
				if err != nil {
					return err
				}
				ret.Resources = append(ret.Resources, resource)
				return nil
			}(); err != nil {
				logger.Err(err).Msgf("error during data store write")
				if err := forwardRead.CloseWithError(err); err != nil {
					logger.Err(err).Msgf("error closing write goroutine")
				}
				writerErrCh <- err
			}
		}()
		return wrappedWriter
	})
```

Based on the logic above, it looks like the data goes directly from the source to the target, using the `COPY FROM` command. This means data is moved directly.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,127 @@
...
+- Will measure the amount of time the export and import takes
+- Will baseline against all three modes of operation (see above)
+
+2. [Randomized Schema Testing](https://github.com/cockroachdb/molt/issues/37)
+
+- Generates random schemas and data of varying types on the source
+- Will run `fetch` on the data to migrate the data to the target
+- Will check the correctness of the data
+
+3. [E2E testing](https://github.com/cockroachdb/molt/issues/36)
```
Jeremyyang920, line 111: I think we can unlink these github issues as they won't be tracked here anymore.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,127 @@
...
+## Drawbacks
+
+By bringing tooling development in house, we will need to invest time to build out feature support that is on par with the existing tools on the market. Then after that, we have the maintenance cost for our tools, while also providing support for existing third party tools.
+
+## Rationale and Alternatives
+
+### Alternatives Considered
+
+1. Private labeling existing tools
+   This was strongly considered, but private labeling would leave us with the same lack of visibility into tooling and supporting changes to source code to provide better compatibility/perf for Cockroach.
```
Jeremyyang920, line 92: The second half of this sentence doesn't flow that well especially after the and. I think its missing a word, or can be rephrased.

docs/RFCS/20231113_molt_fetch.md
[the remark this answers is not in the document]
ryanluu12345: Zooming on this particular line, I do see a nuance though. Even though the data is copied from source to target, it looks like it passes through the `fetch` tool since `fetch` is facilitating this `COPY` from source to target and it passes through the `io.Reader`
```
if _, err := conn.PgConn().CopyFrom(ctx, r, dataquery.CopyFrom(table)); err != nil {
		return nil, errors.CombineErrors(err, conn.Close(ctx))
	}
```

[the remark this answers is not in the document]
ryanluu12345: Ok so I'll point out these key points:
- Direct copy mode allows `fetch` to `COPY FROM` the source to the target
- The data will still pass through `fetch` because it's in charge of coordinating `COPY FROM` and holds onto both DB connections during the process; so compute is still important
- No additional data is stored in an intermediary store or files

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Cool, thanks!

ryanluu12345: Done

docs/RFCS/20231113_molt_fetch.md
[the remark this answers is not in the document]
ryanluu12345: Reviewed this with @Jeremyyang920 offline and I think we have a good understanding here. I'll make sure to call out these constraints. 

TLDR; direct-copy is the most direct way we can copy from source to target since we coordinate the source to target `COPY FROM`

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Done

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: See here: https://github.com/cockroachdb/molt/pull/73#discussion_r1402646616

I need to point out this nuance that data still passes through the machine that `fetch` is running on.

ryanluu12345: Refactored this section to:
```
1. `IMPORT INTO` from intermediary store to Cockroach:

- Target table is taken offline
- Supports compression (recommended option because `IMPORT INTO` is much faster than copy, especially with compression).

2. `COPY FROM` (`live` mode) from intermediary store to Cockroach:

- Table is still online and queryable; the intermediary store is used, instead of directly from the database
- Compression can't be supported here because `COPY FROM` currently doesn't have a `decompress` feature

3. Direct `COPY FROM` source database to Cockroach without intermediary store

- Table is still online and query-able
- Data is copied from the source to target without intermediary store or files
- Data still passes from the source to `fetch`, but in memory; this means that the RAM specs of the execution machine must be appropriately sized
```

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: Added this section:
```
### Testing for Import Into vs. Copy

For Cockroach, `IMPORT INTO` is much faster, as noted above.

Previously, our team performed baseline testing on `IMPORT INTO` vs. `COPY`:
For 10M rows:

- `IMPORT INTO` (PG -> s3 CSVs -> CRDB) took 90s, 30s to upload the files to S3 and 60s to ingest it
- `COPY` (PG output directly into CRDB input) took around 350s (~6 mins) using 4MB COPY chunks.

Follow up parameter tuning notes:

- Using 1MB `COPY` chunks tooks 5mins instead of 6mins from PG->CRDB
- Using `COPY` to migrate from Postgres -> Postgres took 120s using 1MB COPY chunks and 95s using 4MB chunks. Suffice to say, IMPORT is comparable to PG COPY.
```

ryanluu12345: Ok also added the fetch scale testing results so far. Two more tests to run for the `COPY` cases.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,215 @@
...
+# Testing plan
+
+## Scale baseline testing
+
+- Uses the 200M row workload in AWS RDS that we test DMS with
+- Will run the `fetch` process on a Roachprod AWS machine
+- Will measure the amount of time the export and import takes
+- Will baseline against all three modes of operation (see above)
+
+1. AWS DMS test (uses `COPY`): 33 minutes
```
Jeremyyang920, line 185: Is this line entry for comparing how long our DMS roachprod test takes? Since we know that this in theory is way faster when set up correctly so not sure if this is fair to DMS for this benchmark.

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,215 @@
...
+
+2. `IMPORT INTO` with intermediate store and compression: 26 minutes; 14.5 minutes for export, 11.5 minutes for import
+
+3. `COPY FROM` with intermediate store: 2 hours 23 minutes; 7 minutes 29 seconds for export, 2 hrs 15 minutes 52 seconds for import
+
+4. `COPY FROM` directly without intermediate store: 47 minutes
```
Jeremyyang920, line 191: This is just a side comment, but I brought it up briefly last week that we know there are definitely still areas of improvement here in terms of how the process is run so we may want to investigate those and retest. But not a must for this first draft of the RFC

docs/RFCS/20231113_molt_fetch.md
```diff
@@ -0,0 +1,215 @@
...
+
+Cockroach: uses `SELECT` to scan and pagniate through all the data in each table, and write them out to a CSV file.
+
+MySQL: uses `SELECT` to scan and pagniate through all the data in each table, and write them out to a CSV file.
+
+PostgreSQL: uses `Copy To` to dump the data into a CSV file; later on, use the `SELECT` scan/pagination method to improve performance and provide table-granularity parallelization.
+
+### Limitations
+
+The data export process is primarily limited by the dialects supported (mentioned above) and the machine that `fetch` is running on. The number of CPUs on the machine will affect the speed of the export for schemas with many tables. The amount of disk storage will affect the case where intermediate files are used on the machine that `fetch` is running on. This limitation is mitigated if cloud storage (GCS or S3) are used. The amount of RAM will affect the case where `direct copy` mode is used since the data is kept in memory before it's copied over to the target database.
```
Jeremyyang920, line 87: Have we confirmed that the number of CPU's affects speed of export? Im curious the difference in numbers and this would also be something worth documenting if we know what specs could result in what performance.

docs/RFCS/20231113_molt_fetch.md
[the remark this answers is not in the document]
ajstorm: This all SGTM, but I think the above text needs some changing as a result.  Specifically, it would be good if you added all of the detail you uncovered during this investigation and were very precise about the flow of data (including to any involved object storage).

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: I'll note that it can take 18 minutes with the correct parallel table load settings.

ryanluu12345:
```
1. AWS DMS test (uses `COPY`): 33 minutes in nightly tests; 18 minutes in preliminary manual testing with parallel individual table loads (future optimization `fetch` can make as well)
```

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: I'll make a note that there are some testing improvements and testing steps we'll take on in the future.

docs/RFCS/20231113_molt_fetch.md
ryanluu12345: This is what I outlined as `future testing` in the section I added in response to your feedback above. I don't think this is immediately necessary to test for this initial RFC. But with how `fetch` is built, we currently default concurrency for tables to how many CPUs the machine has.

docs/RFCS/20231113_molt_fetch.md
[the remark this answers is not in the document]
ryanluu12345: I actually already updated this in a section below:
```
### Options

Data can be imported into Cockroach in one of three ways:

1. `IMPORT INTO` from intermediary store to Cockroach:

- Target table is taken offline
- Supports compression (recommended option because `IMPORT INTO` is much faster than copy, especially with compression).

Source -> Fetch -> Intermediary store -> Target

2. `COPY FROM` (`live` mode) from intermediary store to Cockroach:

- Table is still online and queryable; the intermediary store is used, instead of directly from the database
- Compression can't be supported here because `COPY FROM` currently doesn't have a `decompress` feature

Source -> Fetch -> Intermediary store -> Target

3. Direct `COPY FROM` source database to Cockroach without intermediary store

- Table is still online and query-able
- Not recommended for very large table migrations
- Data is copied from the source to target without intermediary store or files
- Data still passes from the source to `fetch`, but in memory; this means that the RAM specs of the execution machine must be appropriately sized
- Fetch is coordinating the `COPY FROM` and holds access to the DB connections; this is the most direct path `fetch` can support

Source -> Fetch (in memory) -> Target
```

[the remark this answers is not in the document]
ryanluu12345: I also updated the section you request above:
```
For the initial version, using `COPY` to export data works fine, however we want to move to a model where we can paginate through all the data by selecting data from each of the tables because it allows more flexibility in how we filter through data.

In the case that a `COPY FROM` or `IMPORT INTO` is run and dumps to intermediary files in a store (i.e. AWS S3 or GCP GCS), the data travels from the source database to the machine which is running `fetch`, then to the intermediary store, and then finally to its destination.

In the case that a `COPY` is issued from the source to the target directly, the data is fetched from the source database to the machine that is running `fetch` in memory, formatted in a manner suitable for `COPY` to Cockroach, and copied over to Cockroach.
```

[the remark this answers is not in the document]
ajstorm: Weird.  For some reason I couldn't see those edits until now.  I'm probably doing something wrong on my end though, as I'm used to Reviewable reviews.

:lgtm:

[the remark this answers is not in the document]
ajstorm: > I'm probably doing something wrong on my end though, as I'm used to Reviewable reviews.
As evidenced by the fact that GH doesn't even have an LGTM emoji 🤦

ajstorm (approved):
Thanks for persisting through the review.  This looks good to me.

PR #77, opened 2023-11-15
README: added in instructions for encoding passwords

ryanluu12345:
Release Note: None

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,38 @@ go build -o artifacts/molt .
 GOOS=linux GOARCH=amd64 go build -v -o artifacts/molt .
 ```
 
+## Encoding passwords
+
+If your password contains [special characters](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding), the MOLT tooling may not be able to parse the password. In the event that this happens, you should:
+
+1. Percent/URL-encode the password.
+2. Use the encoded password in your connection string.
+
+```
+# Original connection string
+postgres://postgres:a$52&@localhost:5432/replicationload
+
+# Percent-encoded password
+%3Aa%2452%26
+
+# Percent-encoded connection string
+postgres://postgres:%3Aa%2452%26@localhost:5432/replicationload
+```
+
+To simplify this process, you can use the `escape-password` command:
+
+```
+# Command template
+molt escape-password "<your password string>"
+
+...
+
+# Example output
+molt run . escape-password ";@;"
+Substitute the following encoded password in your original connection url string:
+%3B%40%3B
+```
+
 ## MOLT Verify
 
 `molt verify` does the following:
@@ -44,7 +76,7 @@ molt verify \
   --target 'postgresql://root@127.0.0.1:26257/defaultdb?sslmode=disable'
 ```
 
-See `molt verify --help` for all available parameters.
+See `molt verify --help` for all available parameters. Make sure that your connection strings are [properly encoded](#encoding-passwords).
 
 ### Filters
 
@@ -111,6 +143,8 @@ For now, schemas must be identical on both sides. This is verified upfront -
 tables with mismatching columns may only be partially migrated.
 
 ### Example invocations
+
+Make sure that your connection strings are [properly encoded](#encoding-passwords).
 
 S3 usage:
 
--- a/cmd/escape_password.go
+++ b/cmd/escape_password.go
@@ -1,0 +1,24 @@
+package cmd
+
+import (
+	"errors"
+	"net/url"
+
+	"github.com/spf13/cobra"
+)
+
+func EscapePasswordCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "escape-password",
+		Short: "Utility to help escape passwords with special characters.",
+		Long:  `Utility to help percent-encode passwords that have special characters.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				return errors.New("received empty password string")
+			}
+			passwordStr := args[0]
+			cmd.Printf("Substitute the following encoded password in your original connection url string:\n%s\n", url.QueryEscape(passwordStr))
+			return nil
+		},
+	}
+}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -32,6 +32,7 @@ func Execute() {
 }
 
 func init() {
+	rootCmd.AddCommand(EscapePasswordCommand())
 	rootCmd.AddCommand(verify.Command())
 	rootCmd.AddCommand(fetch.Command())
 


ryanluu12345:
@ZhouXing19 @Jeremyyang920 @ajstorm this will be the topic we're discussing more in detail tomorrow on how to handle on the code. For now, making clarity changes to the docs.

ryanluu12345:
@taroface heads up that I made a DOCS ticket for this companion change: https://cockroachlabs.atlassian.net/browse/DOC-9223

README.md
```diff
@@ -215,3 +217,23 @@ CREATE DATABASE defaultdb;
 
 - Run the tests: `go test ./...`.
   - Data-driven tests can be run with `-rewrite`, e.g. `go test ./verification -rewrite`.
+
+## Notes
+
+### Encoding Passwords
+
+If your password contains special characters, there is chance that the MOLT tooling will not be able to parse. In the event that this happens, here is what you should do:
```
ZhouXing19, line 225: Is this only for the case where the password is embedded in the url? IIUC, if we have plan to let verify tool be accessed via some other interfaces (such as some UI console), this might not be a blocker?

Jeremyyang920: Correct, this is a problem when passed in via a full url. If we do it like we do for the SCT credentials flow and build out the conn string, we can safely escape things.

ryanluu12345: Exactly as Jeremy mentioned. However, I think that escaping would be at the layer where we receive the HTTP or gRPC request, before we even pass it over to the `molt` tooling. 

Given that we don't have a web UI just yet for this, we should give a heads up to customers who are using the CLI version with a conn string.

Jeremyyang920: I feel like we should move this section up before the sample invocations bit.

README.md
```diff
@@ -44,7 +44,7 @@ molt verify \
   --target 'postgresql://root@127.0.0.1:26257/defaultdb?sslmode=disable'
 ```
 
-See `molt verify --help` for all available parameters.
+See `molt verify --help` for all available parameters. Make sure that your connection strings are properly encoded [(instructions)](#encoding-passwords).
```
taroface, line 47:
```suggestion
See `molt verify --help` for all available parameters. Make sure that your connection strings are [properly encoded](#encoding-passwords).
```

README.md
```diff
@@ -112,6 +112,8 @@ tables with mismatching columns may only be partially migrated.
 
 ### Example invocations
 
+Make sure that your connection strings are properly encoded [(instructions)](#encoding-passwords).
```
taroface, line 115:
```suggestion
Make sure that your connection strings are [properly encoded](#encoding-passwords).
```

README.md
taroface, line 225: Can we specify exactly which special characters are problematic?

```suggestion
If your password contains special characters, the MOLT tooling may not be able to parse the password. In the event that this happens, here is what you should do:
```

README.md
```diff
@@ -215,3 +217,23 @@ CREATE DATABASE defaultdb;
...
 - Run the tests: `go test ./...`.
   - Data-driven tests can be run with `-rewrite`, e.g. `go test ./verification -rewrite`.
+
+## Notes
+
+### Encoding Passwords
+
+If your password contains special characters, there is chance that the MOLT tooling will not be able to parse. In the event that this happens, here is what you should do:
+
+1. Percent/URL encode the password.
```
taroface, line 227:
```suggestion
1. Percent/URL-encode the password.
```

README.md
```diff
@@ -215,3 +217,23 @@ CREATE DATABASE defaultdb;
...
+2. Substitute the password back into your connection string.
```
taroface, line 228:
```suggestion
2. Use the encoded password in your connection string.
```

Jeremyyang920 (changes requested):
See the comment about moving the encoding passwords section up so its nearer to where we call it out.

README.md
```diff
@@ -215,3 +217,23 @@ CREATE DATABASE defaultdb;
...
+
+```
+# Original String
```
taroface, line 231:
```suggestion
# Original connection string
```

README.md
```diff
@@ -215,3 +217,23 @@ CREATE DATABASE defaultdb;
...
+postgres://postgres:a$52&@localhost:5432/replicationload
+
+# Percent Encoded Password
```
taroface, line 234:
```suggestion
# Percent-encoded password
```

README.md
```diff
@@ -215,3 +217,23 @@ CREATE DATABASE defaultdb;
...
+%3Aa%2452%26
+
+# Percent Encoded String
```
taroface, line 237:
```suggestion
# Percent-encoded connection string
```

taroface (commented):
Left suggestions and a question on special characters.

README.md
```diff
@@ -215,3 +217,23 @@ CREATE DATABASE defaultdb;
 
 - Run the tests: `go test ./...`.
   - Data-driven tests can be run with `-rewrite`, e.g. `go test ./verification -rewrite`.
+
+## Notes
+
+### Encoding Passwords
```
taroface, line 223:
```suggestion
### Encoding passwords
```

README.md
ryanluu12345: Done!

README.md
ryanluu12345: Done!

README.md
ryanluu12345: Done

README.md
ryanluu12345: Did this instead: 
```
If your password contains [special characters](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding), the MOLT tooling may not be able to parse the password. In the event that this happens, here is what you should do:
```

ryanluu12345: I moved it up to the top!

README.md
```diff
@@ -20,6 +20,24 @@ go build -o artifacts/molt .
 GOOS=linux GOARCH=amd64 go build -v -o artifacts/molt .
 ```
 
+## Encoding passwords
+
+If your password contains [special characters](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding), the MOLT tooling may not be able to parse the password. In the event that this happens, here is what you should do:
```
taroface, line 25:
```suggestion
If your password contains [special characters](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding), the MOLT tooling may not be able to parse the password. In the event that this happens, you should:
```

taroface (approved):
Just one suggestion; otherwise LGTM.

README.md
ryanluu12345: Done!

Jeremyyang920 (approved):
LGTM pending ryans review of the tools copy.

cmd/escape_password.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func EscapePasswordCommand() *cobra.Command {
+	var inputPassword string
+	cmd := &cobra.Command{
+		Use:   "escape-password",
+		Short: "Utility to help escape passwords with special characters.",
+		Long:  `Utility to help percent-encode passwords that have special characters.`,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmd.Printf("Substitute the following string in your original connection string:\n%s\n", url.QueryEscape(inputPassword))
```
ZhouXing19, line 16: nit: personally i found the `--input` flag a bit unnecessary. Would it be easier to use if the password is just the first arg? I.e. something like:

```go
func EscapePasswordCommand() *cobra.Command {
	cmd := &cobra.Command{
		Use:   "escape-password",
		Short: "Utility to help escape passwords with special characters.",
		Long:  `Utility to help percent-encode passwords that have special characters.`,
		Run: func(cmd *cobra.Command, args []string) {
			if len(args) == 0 {
				cmd.Println("received empty password string")
				return
			}
			passwordStr := args[0]
			cmd.Printf("Substitute the following string in your original connection string:\n%s\n", url.QueryEscape(passwordStr))
		},
	}
	return cmd
}
```

With ux:

```
jane in molt on  main [!?⇡] via 🐹 v1.21.3 on ☁️  (us-east-1) on ☁️  jane@cockroachlabs.com(us-east4) took 10s 
❯ ./molt escape-password 
received empty password string
(base) 
jane in molt on  main [!?⇡] via 🐹 v1.21.3 on ☁️  (us-east-1) on ☁️  jane@cockroachlabs.com(us-east4) 
❯ ./molt escape-password wyuwe$6#dsfjh
Substitute the following string in your original connection string:
wyuwe%23dsfjh
(base) 

```

ZhouXing19: Also, out of curiosity, do we want to add test for this command?

ryanluu12345: I'll take this suggestion! I think test wise, it's a very simple function, so it's ok.

cmd/escape_password.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"errors"
+	"net/url"
+
+	"github.com/spf13/cobra"
+)
+
+func EscapePasswordCommand() *cobra.Command {
+	var inputPassword string
```
ZhouXing19, line 11: The inputPassword var and the StringVarP() can be removed now.

cmd/escape_password.go
```diff
@@ -0,0 +1,28 @@
...
+	cmd := &cobra.Command{
+		Use:   "escape-password",
+		Short: "Utility to help escape passwords with special characters.",
+		Long:  `Utility to help percent-encode passwords that have special characters.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				return errors.New("received empty password string")
+			}
+			passwordStr := args[0]
+			cmd.Printf("Substitute the following string in your original connection url string:\n%s\n", url.QueryEscape(passwordStr))
```
taroface, line 21: I assume this is just returning the encoded password and not the whole connection string?

```suggestion
			cmd.Printf("Substitute the following encoded password in your original connection string:\n%s\n", url.QueryEscape(passwordStr))
```

taroface (commented):
One suggestion on the tool output.

cmd/escape_password.go
ryanluu12345: Yep, it's just the encoded password, not the string.

ryanluu12345: I'll take this suggestion.

cmd/escape_password.go
ryanluu12345: Addressed!

ryanluu12345:
```
(⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % go run . escape-password ";@;"             
Substitute the following encoded password in your original connection url string:
%3B%40%3B
```

ZhouXing19 (approved):

PR #79, opened 2023-11-16
molt: unify the release process

ryanluu12345:
**molt: added in script to build tool cross platform and updated Makefile entries**
Added in a helper script so that we build out the tooling in a more standardized way.

Also updated `go-ole` to the latest version in order to address build issues with Windows + Arm64.

**climanifest: added in manifest generation code that outputs an HTML manifest for versions**
This is the common package that will be used across MOLT tools to generate manifest for binaries.

**.github/workflows: added release-cli workflow steps to upload to GCS bucket**
Added logic to run cross platform builds, update the manifest, and upload the manifest and new builds to GCS. Invalidates the cache afterwards.

Removed legacy release steps.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ gen:
 clean_artifacts:
 	cd ./artifacts && rm *
 
-build_cli:
+build_molt_cli:
 	mkdir -p ./artifacts
 	if test "$(version)" = "" ; then \
         echo "tag is not set, try running this command with a tag like 'make build_cli version=1.0.0'"; \
--- a/Makefile
+++ b/Makefile
@@ -1,2 +1,13 @@
 gen:
 	go generate ./...
+
+clean_artifacts:
+	cd ./artifacts && rm *
+
+build_cli:
+	mkdir -p ./artifacts
+	if test "$(version)" = "" ; then \
+        echo "tag is not set, try running this command with a tag like 'make build_cli version=1.0.0'"; \
+        exit 1; \
+    fi
+	./scripts/build-cross-platform.sh ./ ./artifacts/molt $(version)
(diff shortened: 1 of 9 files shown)


climanifest/base-versions.txt
```diff
@@ -0,0 +1,6 @@
+https://molt.cockroachdb.com/molt/cli/molt-latest.darwin-amd64.tgz
```
ryanluu12345, line 1: @Jeremyyang920 @ZhouXing19 definitely want some feedback on the path: `molt/cli` may be a bit too general. Should this maybe be under `molt/tools`. Not exactly sure right now. For reference, LMS is at `lms/cli`.

Just a note that if we change this, we need to change the `release.yaml` to accommodate this.

ryanluu12345:
Note that the CI portions of this code was largely copied over from `crdb-proxy` with minimal changes. The plan is for `crdb-proxy` to depend on `go run github.com/cockroachdb/molt/climanifest` and `scripts/build-cross-platform.sh`. The pipeline will download the script from the public repository.

Makefile
```diff
@@ -1,2 +1,13 @@
 gen:
-	go generate ./...
\ No newline at end of file
+	go generate ./...
+
+clean_artifacts:
+	cd ./artifacts && rm *
+
+build_cli:
```
Jeremyyang920, line 7: Should we rename this to like `build_molt_cli`

climanifest/base-versions.txt
Jeremyyang920: I don't really have a preference on this since we are just storing the artifacts somewhere and as long as its versioned things will be fine.

Makefile
ryanluu12345: Yes, done.

Jeremyyang920 (approved):

ryanluu12345:
Final test run in dev: https://github.com/cockroachdb/molt/actions/runs/6896145990

PR #85, opened 2023-11-28
[wip]fetch: export with FmtExport

ZhouXing19:
Previously, with formating the result from crdb with FmtBareStrings, if the result string contains special char, it will be wrapped with single quotes. This will break the verification as the data imported is not the same as original.


.github/docker-compose.yml
```diff
@@ -3,6 +3,15 @@ services:
     image: cockroachdb/cockroach:latest-v23.1
     network_mode: host
     command: start-single-node --insecure --store type=mem,size=2G
+  cockroachdbtarget:
+    image: cockroachdb/cockroach:latest-v23.1
+    # We have to port forward rather than have it run on host, as the docker image
```
ryanluu12345, line 8: Thanks for leaving the comment; this is super helpful.

fetch/dataexport/dataexport.go
```diff
@@ -65,7 +65,7 @@ func scanWithRowIterator(
 		strings = strings[:0]
 		datums := it.Next(ctx)
 		for _, d := range datums {
-			f := tree.NewFmtCtx(tree.FmtBareStrings | tree.FmtParsableNumerics)
+			f := tree.NewFmtCtx(tree.FmtExport | tree.FmtParsableNumerics)
```
ryanluu12345, line 68: A question from another thread but: is there any consequences of using `FmtExport` over other options? Especially for direct-copy.

fetch/fetch_test.go
```diff
@@ -84,10 +94,62 @@ func TestDataDriven(t *testing.T) {
...
+							// which not necessarily share the network with the host.
+							// When importing the data to the target database,
+							// it requires the database reaches the local storage server
+							// (spun up on host network) from the container (i.e. from
+							// the container's network). According to the 2 links
+							// above, the `localhost` on the host network is accessible
+							// via different endpoint based on the operating system:
+							// - Linux, Windows: 172.17.0.1
+							// - MacOS: host.docker.internal
+							switch runtime.GOOS {
```
ryanluu12345, line 128: Clean, thanks!

fetch/fetch_test.go
```diff
@@ -116,14 +178,14 @@ func TestDataDriven(t *testing.T) {
...
 						require.NoError(t, err)
-						require.NoError(t, src.Cleanup(ctx))
 						return ""
 					default:
 						t.Errorf("unknown command: %s", d.Cmd)
 					}
 
 					return ""
 				})
+				fileIdx++
```
ryanluu12345, line 188: Why is this fileIdx needed?

ryanluu12345 (approved):
LGTM but just want the clarifying questions answered.

verify/inconsistency/reporter.go
```diff
@@ -94,7 +94,7 @@ func (l LogReporter) Report(obj ReportableObject) {
 }
 
 func reportableVal(d tree.Datum) string {
-	f := tree.NewFmtCtx(tree.FmtBareStrings | tree.FmtParsableNumerics)
+	f := tree.NewFmtCtx(tree.FmtExport | tree.FmtParsableNumerics)
```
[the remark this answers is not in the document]
Jeremyyang920: Single quotes are used as identifiers so I think thats correct behavior. If the string was double quoted then I believe it should be different as "item" is different from item

fetch/fetch_test.go
ZhouXing19: Ah, stale code! Removed.

PR #86, opened 2023-11-29
.github/workflows: automate the docs creation ticket after release

ryanluu12345:
Gets the log between the head of the branch and the latest release tag. Then creates a ticket with the relevant fields, using the official JIRA GH action helper.

Resolves: CC-26297, CC-26370
Release Note: Release notes are now more structured and describe customer implications more clearly. These notes are now picked up by automation and will be reflected in documentation pages.

Test run: https://github.com/cockroachdb/molt/actions/runs/7038865628
Created ticket: https://cockroachlabs.atlassian.net/browse/DOC-9302

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -87,3 +87,52 @@ jobs:
       if: ${{ github.event_name == 'release' }}
       run: |
         gcloud compute url-maps invalidate-cdn-cache molt-lms-release-artifacts-prod-default --path "/molt/cli/*" --async
+
+  create-docs-ticket:
+    if: ${{ github.event_name == 'release' }}
+    needs: [release-cli]
+    environment: ${{ github.event_name == 'release' && 'prod' || 'dev' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Extract short log, ref, and date
+        id: extract
+        run: | 
+          echo "Ref: ${GITHUB_REF_NAME#v}"
+          echo "tag=${GITHUB_REF_NAME#v}" >> $GITHUB_OUTPUT
+          echo "release_date=$(date '+%Y-%m-%d')" >> $GITHUB_OUTPUT
+          # We are sorting the tags by semver and getting the second highest because that is
+          # the latest released version. The highest semver is the one currently being released
+          # since the tagging happens right before this runs.
+          latest_tag=$(git tag -l | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -2 | head -1)
+          echo "$latest_tag is the latest tag"
+
+          # Filters all commits from latest tag to now, looks for release notes, excludes empty ones, and adds a dash at the front of each line.
+          output=$(git log $latest_tag..HEAD | grep -i "release note:" | grep -vi 'release note: none' | sed 's/^ *Release Note: *//I' | sed 's/^/- /')
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -87,3 +87,52 @@ jobs:
...
+          echo "tag=${GITHUB_REF_NAME#v}" >> $GITHUB_OUTPUT
+          echo "release_date=$(date '+%Y-%m-%d')" >> $GITHUB_OUTPUT
+          # We are sorting the tags by semver and getting the second highest because that is
+          # the latest released version. The highest semver is the one currently being released
+          # since the tagging happens right before this runs.
+          latest_tag=$(git tag -l | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -2 | head -1)
+          echo "$latest_tag is the latest tag"
+
+          # Filters all commits from latest tag to now, looks for release notes, excludes empty ones, and adds a dash at the front of each line.
+          output=$(git log $latest_tag..HEAD | grep -i "Release note:" | sed 's/^ *Release Note: *//I' | grep -vi 'None' | sed 's/^/- /')
```
ZhouXing19, line 113: I wonder if `grep -vi 'None'` here will exclude commit messages that contains `None` as a substring. IIUC, here we want to exclude the case with `Release Note: None`. So do you think we should do `grep -vi '^none$'`? 

FYI:

```bash
❯ echo "hello none" | grep -vi 'none'
(base)

❯ echo "hello none" | grep -vi '^none$'
hello none
```

ryanluu12345: Actually a really good catch, I'm going to do this.

.github/workflows/release.yaml
```diff
@@ -87,3 +87,52 @@ jobs:
...
+          echo "tag=${GITHUB_REF_NAME#v}" >> $GITHUB_OUTPUT
+          echo "release_date=$(date '+%Y-%m-%d')" >> $GITHUB_OUTPUT
+          # We are sorting the tags by semver and getting the second highest because that is
+          # the latest released version. The highest semver is the one currently being released
+          # since the tagging happens right before this runs.
+          latest_tag=$(git tag -l | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -2 | head -1)
+          echo "$latest_tag is the latest tag"
+
+          # Filters all commits from latest tag to now, looks for release notes, excludes empty ones, and adds a dash at the front of each line.
+          output=$(git log $latest_tag..HEAD | grep -i "Release note:" | sed 's/^ *Release Note: *//I' | grep -vi 'release note: none' | sed 's/^/- /')
```
ryanluu12345, line 113: Did this: `grep -vi 'release note: none'`. The reason is that we want to look for release notes having `None`. As long as we abide by the release note guidelines, this should be safe, and specific enough.

.github/workflows/release.yaml
```diff
@@ -87,3 +87,52 @@ jobs:
...
+          JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
+      - name: Create DOC JIRA ticket
+        id: create
+        uses: atlassian/gajira-create@v3
+        with:
+          project: DOC
+          issuetype: Docs
+          summary: MOLT Fetch/Verify (${{steps.extract.outputs.release_date}}) release ${{steps.extract.outputs.tag}}
+          description: ${{steps.extract.outputs.shortlog}}
+          # 10310 is product area - Migrations. 10175 is doc type - release notes. Assignee is Ryan Kuo.
```
ZhouXing19, line 135: Thanks for the detailed comment!

ZhouXing19 (approved):

PR #88, opened 2023-12-01
README: added specific MySQL setup instructions

ryanluu12345:
Added details about setting up replication using GTIDs so that people can have their MySQL instance ready for use with MOLT fetch. GTID is required because it is a necessary step for returning the cdc_cursor.

Resolves: CC-26359
Release Note: README for MOLT fetch is now updated with MySQL specific setup instructions that instruct users to set up replication using GTID, which is required for MOLT fetch to operate with MySQL correctly. This change will clarify confusion users had about MOLT fetch instructions not working because of a missing pre-condition step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,28 @@ Substitute the following encoded password in your original connection url string
 %3B%40%3B
 ```
 
+## Database-specific setup
+
+### MySQL
+
+A prerequisite of using `molt fetch` with MySQL is that GTID consistency must be enabled. This is necessary for returning the `cdc_cursor`. To enable GTID, pass two flags to the `mysql` start command or define them in `mysl.cnf`:
+
+```
+--gtid-mode=ON
+--enforce-gtid-consistency=ON
+```
+
+Additionally, disable `ONLY_FULL_GROUP_BY`:
+
+```
+// Inside the MySQL shell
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,6 +52,44 @@ Substitute the following encoded password in your original connection url string
...
+  -p 3306:3306 \
+  -e MYSQL_ROOT_PASSWORD=password \
+    mysql:8.2.0 \
+      --server-id=1 \
+      --log-bin=mysql-bin \
+      --binlog-format=ROW \
+      --gtid-mode=ON \
+      --enforce-gtid-consistency=ON \
+      --log-slave-updates
+```
```
Jeremyyang920, line 80: I don't know how much an example is needed for this. I think its fine to just say you must set gtid-mode and enforce-gtid-consistency and we have the links to the mysql doc.

Jeremyyang920: Also most people are not going to be running their mysql command with like with the command flags. Typically it would all be through the mysql.cnf file as well.

ryanluu12345: I'll call out that this will either be set in the start command or defined in a `mysql.cnf`.

ryanluu12345: Done, thoughts?

codingconcepts (approved):
LGTM!

Issue #93, opened 2023-12-06
fetch: strategy for AOST clause given possible time shift caused by docker


ZhouXing19:
On Macos, docker could bring time shift \[[1](https://www.docker.com/blog/addressing-time-drift-in-docker-desktop-for-mac/)\]\[[2](https://www.docker.com/blog/addressing-time-drift-in-docker-desktop-for-mac/)\], which means the server in docker sees inconsistent time with the host. 

This is a concern as we [use](https://github.com/cockroachdb/molt/blob/87cbfb55fa20970f8e0f2fe872c345de6288a7fa/rowiterator/scan_iterator.go#L206-L211) `AS OF SYSTEM TIME` clause (AOST) when fetch the data from a crdb source. If the crdb source database is in a docker container, such an AOST clause may bring the following error:
```
ERROR: AS OF SYSTEM TIME: cannot specify timestamp in the future (1701836988.000000000,0 > 1701836987.322737000,0) (SQLSTATE XXUUU)
```

## Reproduction Step

0. Make sure you're in MacOS.
1. Start the crdb server in docker, such as `docker run -d --name=roach5 --hostname=roach5 -p 26257:26257 -p 8080:8080 cockroachdb/cockroach:latest-v23.1 start-single-node --insecure`
2. Run the following test:
```go
func TestExportFromCrdbFailed(t *testing.T) {
	t.Run("fetch from crdb", func(t *testing.T) {
		connStr := "postgres://root@127.0.0.1:26257/defaultdb?sslmode=disable"

		ctx := context.Background()

		var err error

		cfg, err := pgx.ParseConfig(connStr)
		require.NoError(t, err)
		crdbConn, err := pgx.ConnectConfig(ctx, cfg)
		require.NoError(t, err)
		defer func() {
			require.NoError(t, crdbConn.Close(ctx))
		}()
		_, err = crdbConn.Exec(ctx, "DROP TABLE IF EXISTS tb1")
		require.NoError(t, err)
		_, err = crdbConn.Exec(ctx, "CREATE TABLE tb1 (x int)")
		require.NoError(t, err)

		aost := time.Now().UTC().Truncate(time.Second)
		asotExpr, err := tree.MakeDTimestamp(aost, time.Microsecond)
		require.NoError(t, err)

		tn := tree.MakeTableNameFromPrefix(
			tree.ObjectNamePrefix{SchemaName: "public", ExplicitSchema: true},
			"tb1",
		)

[...]

PR #96, opened 2023-12-06
fetch: add edge case explanation for time drift

ZhouXing19:
Based on:

1. https://github.com/cockroachdb/molt/issues/93
2. https://cockroachlabs.slack.com/archives/C04TQ67V1JL/p1701871210926509

Release note: Updated the readme for a OS + docker based edge case with fetch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,50 @@ molt fetch \
   --pg-logical-replication-slot-decoding 'pgoutput'
 ```
 
+### Edge case
+#### MacOS + CockroachDB as Source within Docker container
+
+If you encounter an error similar to the following, please contact the support team and we will try to provide a solution.
+
+```
+ERROR: AS OF SYSTEM TIME: cannot specify timestamp in the future (1701836988.000000000,0 > 1701836987.322737000,0) (SQLSTATE XXUUU)
+```
+
+This error is due to the fact that with MacOS as the runtime OS, Docker may have indeterministic time drift from the host machine.[[1](https://github.com/cockroachdb/molt/issues/93)]
+
+Example to reproduce the time drift:
+```bash
+#!/bin/bash
+
+set -e
+docker rm -f random-cont
+docker run -d --name random-cont alpine sh -c "apk add --no-cache coreutils && tail -f /dev/null"
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -245,6 +245,50 @@ molt fetch \
   --pg-logical-replication-slot-decoding 'pgoutput'
 ```
 
+### Edge case
+#### MscOS + CockroachDB as Source within Docker container
```
ryanluu12345, line 249: Typo: `MacOS` instead of `MscOS`.

ZhouXing19:
@taroface tagging you for review as this may needed to be mentioned in doc too. The tricky part is that we don't have a solution to this so far.

README.md
```diff
@@ -245,6 +245,50 @@ molt fetch \
...
+
+If you encountered an error similar to the following, please contact the support time and we will try to provide a solution.
+
+```
+ERROR: AS OF SYSTEM TIME: cannot specify timestamp in the future (1701836988.000000000,0 > 1701836987.322737000,0) (SQLSTATE XXUUU)
+```
+
+This error is due to the fact that with MacOS as the runtime OS, docker may have indeterministic time drift from the host machine.[[1](https://github.com/cockroachdb/molt/issues/93)]
```
ryanluu12345, line 257: Super nit: `Docker` instead of `docker`

ryanluu12345 (approved):
A few typos and wording changes. But otherwise LGTM. Merge after changes are in.

README.md
```diff
@@ -245,6 +245,50 @@ molt fetch \
...
+time_diff_seconds=$(bc <<< "scale=6; $time_diff / 1000")
+
+echo "Time difference: ${time_diff} milliseconds"
+echo "Time difference: ${time_diff_seconds} seconds"
+echo
+
+done
+```
+
+This will cause problem when using `molt fetch` to export data from a CockroachDB within a container, as we run a `SELECT ... AS OF SYSTEM TIME` query to iterate content from the target table.
```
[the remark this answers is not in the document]
ZhouXing19: Good point, fixed it in https://github.com/cockroachdb/molt/pull/97. Thanks!!

PR #99, opened 2023-12-09
molt: added in logger field for type to differentiate between task, summary, and data logging

ryanluu12345:
Added in logger types and Zerolog log context so that logs can be tagged with different contextual information. Also added utility methods that allow a user to retrieve a summary or data logger from their base logger.

Resolves: CC-26366
Release Note: Added tagging for summary and data logs so that customers can filter by "summary" or "data" in a variety of log processing tools and even grep in the command line. This solves the pain point of all the logs going to the same location without any differentiation. Now it's possible to view types of logs one at a time.

**Options Considered**
1. Making a base logger with the type already set => good solution and almost like the solution presented here, however, this led to duplicated `type` fields because zerolog doesn't de-duplicate repeated fields (https://github.com/rs/zerolog/issues/205)
2. New `MOLTLogger` type that contains multiple loggers inside of it => needs massive refactoring, worse DX, more complexity for little benefit
3. (CHOSEN) Making a base logger without the type set and helper methods to retrieve data and summary loggers => given that the minority logging case is summary or data logging, we have helpers that help us retrieve a tagged data/summary logger when we need it. An untagged, no `type` log is considered a task log. 
- We can grep for summary and data logs:
```
// Summary logs
(⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % cat task.log | grep '"summary"'
{"level":"info","type":"summary","num_tables":1,"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:24-08:00","message":"starting fetch"}
{"level":"info","table":"public.employees","type":"summary","num_rows":200000,"export_duration":2068.466084,"time":"2023-12-08T16:55:26-08:00","message":"data extraction from source complete"}
{"level":"info","table":"public.employees","type":"summary","net_duration":5464.5905,"import_duration":3328.998625,"export_duration":2068.466084,"num_rows":200000,"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:30-08:00","message":"data import on target for table complete"}
{"level":"info","type":"summary","num_tables":1,"tables":["public.employees"],"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:30-08:00","message":"fetch complete"}

// Data logs
(⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % cat task.log | grep '"data"'   
{"level":"info","table":"public.employees","type":"data","num_rows":100000,"time":"2023-12-08T16:55:25-08:00","message":"row import status"}
{"level":"info","table":"public.employees","type":"data","num_rows":200000,"time":"2023-12-08T16:55:25-08:00","message":"row import status"}

// Task logs
(⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % cat task.log | grep -v -e '"summary"' -e '"data"'
{"level":"info","time":"2023-12-08T16:55:24-08:00","message":"default compression to GZIP"}
{"level":"info","time":"2023-12-08T16:55:24-08:00","message":"checking database details"}
{"level":"info","source_table":"public.employees","target_table":"public.employees","time":"2023-12-08T16:55:24-08:00","message":"found matching table"}
{"level":"info","time":"2023-12-08T16:55:24-08:00","message":"verifying common tables"}
{"level":"info","time":"2023-12-08T16:55:24-08:00","message":"establishing snapshot"}
{"level":"info","table":"public.employees","time":"2023-12-08T16:55:24-08:00","message":"data extraction phase starting"}
{"level":"info","table":"public.employees","time":"2023-12-08T16:55:26-08:00","message":"truncating table"}
{"level":"info","table":"public.employees","time":"2023-12-08T16:55:26-08:00","message":"starting data import on target"}
```

Option 2 can be revisited if we ever need multiple output sources for our logger. For now, it's not worth the complexity.

### DIFF
--- a/cmd/fetch/fetch.go
+++ b/cmd/fetch/fetch.go
@@ -11,6 +11,7 @@ import (
 	"github.com/cockroachdb/molt/dbconn"
 	"github.com/cockroachdb/molt/fetch"
 	"github.com/cockroachdb/molt/fetch/datablobstorage"
+	"github.com/cockroachdb/molt/moltlogger"
 	"github.com/spf13/cobra"
 	"github.com/thediveo/enumflag/v2"
 	"golang.org/x/oauth2/google"
@@ -34,8 +35,7 @@ func Command() *cobra.Command {
 		Long:  `Imports data from source directly into target tables.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := context.Background()
-
-			logger, err := cmdutil.Logger(logFile)
+			logger, err := moltlogger.Logger(logFile)
 			if err != nil {
 				return err
 			}
@@ -228,8 +228,8 @@ func Command() *cobra.Command {
 		"compression",
 		"Compression type (default/gzip/none) to use (IMPORT INTO mode only).",
 	)
+	moltlogger.RegisterLoggerFlags(cmd)
 	cmdutil.RegisterDBConnFlags(cmd)
-	cmdutil.RegisterLoggerFlags(cmd)
 	cmdutil.RegisterNameFilterFlags(cmd)
 	cmdutil.RegisterMetricsFlags(cmd)
 	return cmd
--- a/fetch/copy.go
+++ b/fetch/copy.go
@@ -8,6 +8,7 @@ import (
 	"github.com/cockroachdb/molt/dbtable"
 	"github.com/cockroachdb/molt/fetch/datablobstorage"
 	"github.com/cockroachdb/molt/fetch/internal/dataquery"
+	"github.com/cockroachdb/molt/moltlogger"
 	"github.com/rs/zerolog"
 )
 
@@ -23,6 +24,7 @@ func Copy(
 	table dbtable.VerifiedTable,
 	resources []datablobstorage.Resource,
 ) (CopyResult, error) {
+	dataLogger := moltlogger.GetDataLogger(logger)
 	ret := CopyResult{
 		StartTime: time.Now(),
 	}
@@ -30,7 +32,7 @@ func Copy(
 	conn := baseConn.(*dbconn.PGConn).Conn
 
 	for i, resource := range resources {
-		logger.Debug().
+		dataLogger.Debug().
 			Int("idx", i+1).
 			Msgf("reading resource")
 		if err := func() error {
@@ -38,7 +40,7 @@ func Copy(
 			if err != nil {
 				return err
 			}
-			logger.Debug().
+			dataLogger.Debug().
 				Int("idx", i+1).
 				Msgf("running copy from resource")
 			if _, err := conn.PgConn().CopyFrom(
@@ -55,7 +57,7 @@ func Copy(
 	}
 
 	ret.EndTime = time.Now()
-	logger.Info().
+	dataLogger.Info().
 		Dur("duration", ret.EndTime.Sub(ret.StartTime)).
 		Msgf("table COPY complete")
 	return ret, nil
--- a/moltlogger/logger.go
+++ b/moltlogger/logger.go
@@ -1,4 +1,4 @@
-package cmdutil
+package moltlogger
 
 import (
 	"io"
@@ -8,6 +8,16 @@ import (
 
 	"github.com/rs/zerolog"
 	"github.com/spf13/cobra"
+)
+
+const (
+	LoggerTypeKey = "type"
+	// Summary logs are ones that report on either task or step summaries
+	// (i.e. 500 tables processed; task: "10 minutes")
+	LoggerTypeSummary = "summary"
+	// Data logs are ones that report on row or individual level operations
+	// and quantitiative measures (i.e. 10000 rows completed; failed to verify item with PK 1000)
+	LoggerTypeData = "data"
 )
 
 type loggerConfig struct {
@@ -35,6 +45,10 @@ func RegisterLoggerFlags(cmd *cobra.Command) {
 	)
 }
 
+// The default logger returned without a `type` attribute is considered the task logger.
+// We only mark a log differently if it relates to data and summary fields because
+// those are the minority case and we want people to know that this information
+// is relevant to the performance / operation of their data load.
 func Logger(fileName string) (zerolog.Logger, error) {
 	var writer io.Writer = os.Stdout
 	if loggerConfigInst.useConsoleWriter {
@@ -61,5 +75,14 @@ func Logger(fileName string) (zerolog.Logger, error) {
 	if err != nil {
 		return logger, err
 	}
+
 	return logger.Level(lvl).With().Timestamp().Logger(), err
+}
+
+func GetDataLogger(logger zerolog.Logger) zerolog.Logger {
+	return logger.With().Str(LoggerTypeKey, LoggerTypeData).Logger()
+}
+
+func GetSummaryLogger(logger zerolog.Logger) zerolog.Logger {
+	return logger.With().Str(LoggerTypeKey, LoggerTypeSummary).Logger()
 }
--- a/utils/time_test.go
+++ b/utils/time_test.go
@@ -1,0 +1,58 @@
+package utils
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFormatMillisecondsToTime(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    time.Duration
+		expected string
+	}{
+		{
+			name:     "seconds duration",
+			input:    20 * time.Second,
+			expected: "000h 00m 20s",
+		},
+		{
+			name:     "minutes + seconds duration",
+			input:    80 * time.Second,
+			expected: "000h 01m 20s",
+		},
+		{
+			name:     "minutes duration",
+			input:    20 * time.Minute,
+			expected: "000h 20m 00s",
+		},
+		{
+			name:     "hours + minutes duration",
+			input:    80 * time.Minute,
+			expected: "001h 20m 00s",
+		},
+		{
+			name:     "hours duration",
+			input:    20 * time.Hour,
+			expected: "020h 00m 00s",
+		},
+		{
+			name:     "hours + minutes + seconds duration",
+			input:    80*time.Minute + 10*time.Second,
+			expected: "001h 20m 10s",
+		},
+		{
+			name:     "weeks duration",
+			input:    24 * 7 * 2 * time.Hour,
+			expected: "336h 00m 00s",
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			val := FormatDurationToTimeString(tc.input)
+			require.Equal(t, tc.expected, val)
+		})
+	}
+}
--- a/cmd/verify/verify.go
+++ b/cmd/verify/verify.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/molt/cmd/internal/cmdutil"
+	"github.com/cockroachdb/molt/moltlogger"
 	"github.com/cockroachdb/molt/retry"
 	"github.com/cockroachdb/molt/verify"
 	"github.com/cockroachdb/molt/verify/inconsistency"
@@ -38,7 +39,7 @@ func Command() *cobra.Command {
 		}
 		verifyLimitRowsPerSecond int
 		verifyRows               bool
-		verifyLogFile            string
+		logFile                  string
 	)
 
 	cmd := &cobra.Command{
@@ -46,7 +47,7 @@ func Command() *cobra.Command {
 		Short: "Verify table schemas and row data align.",
 		Long:  `Verify ensure table schemas and row data between the two databases are aligned.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			logger, err := cmdutil.Logger(verifyLogFile)
+			logger, err := moltlogger.Logger(logFile)
 			if err != nil {
 				return err
 			}
@@ -94,7 +95,7 @@ func Command() *cobra.Command {
 		},
 	}
 	cmd.PersistentFlags().StringVar(
-		&verifyLogFile,
+		&logFile,
 		"log-file",
 		"",
 		"If set, writes to the log file specified. Otherwise, only writes to stdout.",
@@ -216,8 +217,8 @@ func Command() *cobra.Command {
 			panic(err)
 		}
 	}
+	moltlogger.RegisterLoggerFlags(cmd)
 	cmdutil.RegisterDBConnFlags(cmd)
-	cmdutil.RegisterLoggerFlags(cmd)
 	cmdutil.RegisterNameFilterFlags(cmd)
 	cmdutil.RegisterMetricsFlags(cmd)
 	return cmd
--- a/fetch/csv_pipe.go
+++ b/fetch/csv_pipe.go
@@ -5,6 +5,7 @@ import (
 	"io"
 
 	"github.com/cockroachdb/molt/dbtable"
+	"github.com/cockroachdb/molt/moltlogger"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/rs/zerolog"
@@ -54,6 +55,7 @@ func (p *csvPipe) Pipe(tn dbtable.Name) error {
 	r := csv.NewReader(p.in)
 	r.ReuseRecord = true
 	m := importedRows.WithLabelValues(tn.SafeString())
+	dataLogger := moltlogger.GetDataLogger(p.logger)
 	for {
 		record, err := r.Read()
 		if err != nil {
@@ -67,7 +69,7 @@ func (p *csvPipe) Pipe(tn dbtable.Name) error {
 		p.numRows++
 		m.Inc()
 		if p.numRows%100000 == 0 {
-			p.logger.Info().Int("num_rows", p.numRows).Msgf("row import status")
+			dataLogger.Info().Int("num_rows", p.numRows).Msgf("row import status")
 		}
 		for _, s := range record {
 			p.currSize += len(s) + 1
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -11,7 +11,9 @@ import (
 	"github.com/cockroachdb/molt/dbconn"
 	"github.com/cockroachdb/molt/fetch/datablobstorage"
 	"github.com/cockroachdb/molt/fetch/dataexport"
+	"github.com/cockroachdb/molt/moltlogger"
 	"github.com/cockroachdb/molt/molttelemetry"
+	"github.com/cockroachdb/molt/utils"
 	"github.com/cockroachdb/molt/verify/dbverify"
 	"github.com/cockroachdb/molt/verify/tableverify"
 	"github.com/rs/zerolog"
@@ -58,7 +60,8 @@ func Fetch(
 	}
 	reportTelemetry(logger, cfg, conns, blobStore)
 
-	logger.Debug().
+	dataLogger := moltlogger.GetDataLogger(logger)
+	dataLogger.Debug().
 		Int("flush_size", cfg.FlushSize).
 		Int("flush_num_rows", cfg.FlushRows).
 		Str("store", fmt.Sprintf("%T", blobStore)).
@@ -104,7 +107,9 @@ func Fetch(
 			logger.Err(err).Msgf("error closing export source")
 		}
 	}()
-	logger.Info().
+
+	summaryLogger := moltlogger.GetSummaryLogger(logger)
+	summaryLogger.Info().
 		Int("num_tables", len(tables)).
 		Str("cdc_cursor", sqlSrc.CDCCursor()).
 		Msgf("starting fetch")
@@ -148,7 +153,7 @@ func Fetch(
 		return err
 	}
 
-	logger.Info().
+	summaryLogger.Info().
 		Int("num_tables", stats.numImportedTables).
 		Strs("tables", stats.importedTables).
 		Str("cdc_cursor", sqlSrc.CDCCursor()).
@@ -199,9 +204,12 @@ func fetchTable(
 		}()
 	}
 
-	logger.Info().
+	exportDuration := e.EndTime.Sub(e.StartTime)
+	summaryLogger := moltlogger.GetSummaryLogger(logger)
+	summaryLogger.Info().
 		Int("num_rows", e.NumRows).
-		Dur("export_duration", e.EndTime.Sub(e.StartTime)).
+		Dur("export_duration_ms", exportDuration).
+		Str("export_duration", utils.FormatDurationToTimeString(exportDuration)).
 		Msgf("data extraction from source complete")
 
 	if blobStore.CanBeTarget() {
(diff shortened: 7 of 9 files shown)


ryanluu12345:
@ZhouXing19 @Jeremyyang920 open question, do you think I should just combine `summary` and `data` logging so that it's the same entity? I'm wondering if users really care that certain data is summary data vs. just data about the run as it's going.

My thought is that it might actually be more context switching to have both since people probably want to see all the data associated with a run in structured format. But also, it does help to separate end results from intermediary ones.

I guess we can keep it because people can just do a grep for both summary and data logs if they want:
```
(⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % cat task.log | grep -e '"summary"' -e '"data"' 
{"level":"info","type":"summary","num_tables":1,"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:24-08:00","message":"starting fetch"}
{"level":"info","table":"public.employees","type":"data","num_rows":100000,"time":"2023-12-08T16:55:25-08:00","message":"row import status"}
{"level":"info","table":"public.employees","type":"data","num_rows":200000,"time":"2023-12-08T16:55:25-08:00","message":"row import status"}
{"level":"info","table":"public.employees","type":"summary","num_rows":200000,"export_duration":2068.466084,"time":"2023-12-08T16:55:26-08:00","message":"data extraction from source complete"}
{"level":"info","table":"public.employees","type":"summary","net_duration":5464.5905,"import_duration":3328.998625,"export_duration":2068.466084,"num_rows":200000,"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:30-08:00","message":"data import on target for table complete"}
{"level":"info","type":"summary","num_tables":1,"tables":["public.employees"],"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:30-08:00","message":"fetch complete"}
```

moltlogger/logger.go
```diff
@@ -1,4 +1,4 @@
-package cmdutil
```
ryanluu12345, line 1: Moved loggers here because I want this publicly accessible by all of the tools within `molt`. Also, didn't want to make all the `cmd/internal/cmdutils` public because that will expose things that shouldn't be touched outside the interfaces.

Jeremyyang920:
> @ZhouXing19 @Jeremyyang920 open question, do you think I should just combine `summary` and `data` logging so that it's the same entity? I'm wondering if users really care that certain data is summary data vs. just data about the run as it's going.
> 
> My thought is that it might actually be more context switching to have both since people probably want to see all the data associated with a run in structured format. But also, it does help to separate end results from intermediary ones.
> 
> I guess we can keep it because people can just do a grep for both summary and data logs if they want:
> 
> ```
> (⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % cat task.log | grep -e '"summary"' -e '"data"' 
> {"level":"info","type":"summary","num_tables":1,"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:24-08:00","message":"starting fetch"}
> {"level":"info","table":"public.employees","type":"data","num_rows":100000,"time":"2023-12-08T16:55:25-08:00","message":"row import status"}
> {"level":"info","table":"public.employees","type":"data","num_rows":200000,"time":"2023-12-08T16:55:25-08:00","message":"row import status"}
> {"level":"info","table":"public.employees","type":"summary","num_rows":200000,"export_duration":2068.466084,"time":"2023-12-08T16:55:26-08:00","message":"data extraction from source complete"}
> {"level":"info","table":"public.employees","type":"summary","net_duration":5464.5905,"import_duration":3328.998625,"export_duration":2068.466084,"num_rows":200000,"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:30-08:00","message":"data import on target for table complete"}
> {"level":"info","type":"summary","num_tables":1,"tables":["public.employees"],"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:30-08:00","message":"fetch complete"}
> ```

I think its fine to have the different types. Since depending on what im searching for, I can just change the indexed field when using my log viewer.

ryanluu12345:
> > @ZhouXing19 @Jeremyyang920 open question, do you think I should just combine `summary` and `data` logging so that it's the same entity? I'm wondering if users really care that certain data is summary data vs. just data about the run as it's going.
> > My thought is that it might actually be more context switching to have both since people probably want to see all the data associated with a run in structured format. But also, it does help to separate end results from intermediary ones.
> > I guess we can keep it because people can just do a grep for both summary and data logs if they want:
> > ```
> > (⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % cat task.log | grep -e '"summary"' -e '"data"' 
> > {"level":"info","type":"summary","num_tables":1,"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:24-08:00","message":"starting fetch"}
> > {"level":"info","table":"public.employees","type":"data","num_rows":100000,"time":"2023-12-08T16:55:25-08:00","message":"row import status"}
> > {"level":"info","table":"public.employees","type":"data","num_rows":200000,"time":"2023-12-08T16:55:25-08:00","message":"row import status"}
> > {"level":"info","table":"public.employees","type":"summary","num_rows":200000,"export_duration":2068.466084,"time":"2023-12-08T16:55:26-08:00","message":"data extraction from source complete"}
> > {"level":"info","table":"public.employees","type":"summary","net_duration":5464.5905,"import_duration":3328.998625,"export_duration":2068.466084,"num_rows":200000,"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:30-08:00","message":"data import on target for table complete"}
> > {"level":"info","type":"summary","num_tables":1,"tables":["public.employees"],"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:30-08:00","message":"fetch complete"}
> > ```
> 
> I think its fine to have the different types. Since depending on what im searching for, I can just change the indexed field when using my log viewer.

Cool, sounds like a plan. Any qualms around the `type` field? Maybe that might conflict with reserved words? Do you think a rename is necessary?

Jeremyyang920:
> > > @ZhouXing19 @Jeremyyang920 open question, do you think I should just combine `summary` and `data` logging so that it's the same entity? I'm wondering if users really care that certain data is summary data vs. just data about the run as it's going.
> > > My thought is that it might actually be more context switching to have both since people probably want to see all the data associated with a run in structured format. But also, it does help to separate end results from intermediary ones.
> > > I guess we can keep it because people can just do a grep for both summary and data logs if they want:
> > > ```
> > > (⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % cat task.log | grep -e '"summary"' -e '"data"' 
> > > {"level":"info","type":"summary","num_tables":1,"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:24-08:00","message":"starting fetch"}
> > > {"level":"info","table":"public.employees","type":"data","num_rows":100000,"time":"2023-12-08T16:55:25-08:00","message":"row import status"}
> > > {"level":"info","table":"public.employees","type":"data","num_rows":200000,"time":"2023-12-08T16:55:25-08:00","message":"row import status"}
> > > {"level":"info","table":"public.employees","type":"summary","num_rows":200000,"export_duration":2068.466084,"time":"2023-12-08T16:55:26-08:00","message":"data extraction from source complete"}
> > > {"level":"info","table":"public.employees","type":"summary","net_duration":5464.5905,"import_duration":3328.998625,"export_duration":2068.466084,"num_rows":200000,"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:30-08:00","message":"data import on target for table complete"}
> > > {"level":"info","type":"summary","num_tables":1,"tables":["public.employees"],"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:30-08:00","message":"fetch complete"}
> > > ```
> > 
> > 
> > I think its fine to have the different types. Since depending on what im searching for, I can just change the indexed field when using my log viewer.
> 
> Cool, sounds like a plan. Any qualms around the `type` field? Maybe that might conflict with reserved words? Do you think a rename is necessary?

I think its fine. If we want to rename it I think the other sensible one is `kind`

utils/time_test.go
```diff
@@ -0,0 +1,58 @@
...
+			expected: "020h 00m 00s",
+		},
+		{
+			name:     "hours + minutes + seconds duration",
+			input:    80*time.Minute + 10*time.Second,
+			expected: "001h 20m 10s",
+		},
+		{
+			name:     "weeks duration",
+			input:    24 * 7 * 2 * time.Hour,
```
Jeremyyang920, line 48: I believe there is a time.Week if im not mistaken. So you can just do 2*time.Week

cmd/fetch/fetch.go
```diff
@@ -34,8 +35,7 @@ func Command() *cobra.Command {
 		Long:  `Imports data from source directly into target tables.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := context.Background()
-
-			logger, err := cmdutil.Logger(logFile)
+			logger, err := moltlogger.Logger(logFile)
```
Jeremyyang920, line 38: for consistency with verify, lets either name this fetchLogFile or just name them both logFile

fetch/copy.go
```diff
@@ -23,22 +24,23 @@ func Copy(
 	table dbtable.VerifiedTable,
 	resources []datablobstorage.Resource,
 ) (CopyResult, error) {
+	dataLogger := moltlogger.GetDataLogger(logger)
```
Jeremyyang920, line 27: This is a general comment, but should we be always calling GetDataLogger in each spot or should we be passing int he right logger into the function already.

ryanluu12345: I think we should always pass in the root logger without any of the tagging because a function will almost always emit a task log (i.e. error logs) but won't always do data logs. If you pass in the data logger here, you can only emit data logs.

So my recommendation is to just call `GetDataLogger`, which will annotate the log (pretty quick); I'd recommend doing it at the start of the method that needs it so that it doesn't get re-called in a loop for example.

cmd/fetch/fetch.go
ryanluu12345: I'll make them both `logFile`. I think the prefix is unnecessary especially when the variables are defined within a function within another package.

utils/time_test.go
ryanluu12345: Sadly not :(

image

Gonna keep this code as it is.

fetch/copy.go
Jeremyyang920: Oh I mean like at the base of the fetch process we should already have dataLogger and taskLogger. And those have already called the respective gets. That eleimates the need for the get call in the function. 

I don't have a strong opinion on either, but it just seems like we don't need to call the Get here if we are the ones controlling what is passed in.

ryanluu12345:
> > > > @ZhouXing19 @Jeremyyang920 open question, do you think I should just combine `summary` and `data` logging so that it's the same entity? I'm wondering if users really care that certain data is summary data vs. just data about the run as it's going.
> > > > My thought is that it might actually be more context switching to have both since people probably want to see all the data associated with a run in structured format. But also, it does help to separate end results from intermediary ones.
> > > > I guess we can keep it because people can just do a grep for both summary and data logs if they want:
> > > > ```
> > > > (⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % cat task.log | grep -e '"summary"' -e '"data"' 
> > > > {"level":"info","type":"summary","num_tables":1,"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:24-08:00","message":"starting fetch"}
> > > > {"level":"info","table":"public.employees","type":"data","num_rows":100000,"time":"2023-12-08T16:55:25-08:00","message":"row import status"}
> > > > {"level":"info","table":"public.employees","type":"data","num_rows":200000,"time":"2023-12-08T16:55:25-08:00","message":"row import status"}
> > > > {"level":"info","table":"public.employees","type":"summary","num_rows":200000,"export_duration":2068.466084,"time":"2023-12-08T16:55:26-08:00","message":"data extraction from source complete"}
> > > > {"level":"info","table":"public.employees","type":"summary","net_duration":5464.5905,"import_duration":3328.998625,"export_duration":2068.466084,"num_rows":200000,"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:30-08:00","message":"data import on target for table complete"}
> > > > {"level":"info","type":"summary","num_tables":1,"tables":["public.employees"],"cdc_cursor":"0/3DB3098","time":"2023-12-08T16:55:30-08:00","message":"fetch complete"}
> > > > ```
> > > 
> > > 
> > > I think its fine to have the different types. Since depending on what im searching for, I can just change the indexed field when using my log viewer.
> > 
> > 
> > Cool, sounds like a plan. Any qualms around the `type` field? Maybe that might conflict with reserved words? Do you think a rename is necessary?
> 
> I think its fine. If we want to rename it I think the other sensible one is `kind`

I'll keep it as type.

fetch/copy.go
ryanluu12345: That was option 2 that I called out in the PR description. Currently requires massive refactoring because we need to drill this change in layers deep, which I'm not that thrilled about right now.

I do think that is a better option longer term to have a `MOLTLogger` struct that contains all the types of loggers internally. This would make it so that each method only needs to pass one logger that has subloggers.

Curious to hear @ZhouXing19 's thoughts here as well.

Jeremyyang920: Im fine with it now since I don't envision the logging getting too much more complicated unelss you want to undertake this refactor now.

moltlogger/logger.go
```diff
@@ -10,6 +10,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	LoggerTypeKey     = "type"
+	LoggerTypeSummary = "summary"
```
ZhouXing19, line 15: What are the definition / difference of summary and data log? I think we will need to document this for customers too...?

ZhouXing19: Might be helpful to have inline comments too.

fetch/copy.go
ZhouXing19: My only concern with calling the function each time is that in `Logger.With()`, each time it's called, it append the newly created context to the context list. Seems that the context slice is fixed with length 500. 

```go
l.context = make([]byte, 0, 500)
```

I'm not sure if this will bring any hazard tho.

ZhouXing19: Maybe one way to test possible hazard is to try call `Logger.With()` > 500 times, and see if any error would happen?

moltlogger/logger.go
ryanluu12345: Went with this: 
```
// Summary logs are ones that report on either task or step summaries
	// (i.e. 500 tables processed; task: "10 minutes")
	LoggerTypeSummary = "summary"
	// Data logs are ones that report on row or individual level operations
	// and quantitiative measures (i,e. 10000 rows completed; failed to verify item with PK 1000)
	LoggerTypeData = "data"
```

fetch/copy.go
ryanluu12345: I think the problem here would happen if we pass in results from `dataLogger = GetDataLogger`. That way, we keep appending to the same context. 

However, the intention here is that we _**should never pass in dataloggers or summaryloggers**_ as parameters. We only pass the root logger, which will have a pretty fixed context, as defined in `moltlogger/logger.go`.

Hope that clears up the concern?

ryanluu12345: I also tried the test you requested:
```
logger, err := moltlogger.Logger(logFile)
			if err != nil {
				return err
			}
			cmdutil.RunMetricsServer(logger)

			for i := 1; i <= 500; i++ {
				logger = moltlogger.GetDataLogger(logger)
			}
			logger.Info().Msg("done")
			os.Exit(1)
```

Output:
```
{"level":"info","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","type":"data","time":"2023-12-11T12:42:47-08:00","message":"done"}
```

ZhouXing19: > I think the problem here would happen if we pass in results from dataLogger = GetDataLogger. That way, we keep appending to the same context.

Hmm not sure if I got this. IIUC, we append to the root logger's context each time we call GetDataLogger or GetSummaryLogger. My concern is calling these 2 functions too many times will make the root loggers context list out of boundary.

moltlogger/logger.go
ZhouXing19: Nice, thanks!!

fetch/copy.go
ZhouXing19: Yeah, if we have tested it that calling these 2 funcs >500 times and no performance issue, then it looks good to me. Wondering if we wanna add such an automatic test as well.

ZhouXing19 (approved):
LGTM but I'll let Jeremy finalize the review

moltlogger/logger.go
ryanluu12345: Of course!

fetch/copy.go
ryanluu12345: Well there would be a performance issue logging this much context each time haha. It effectively increases the amount of bytes by a large degree. But I think in practice, this will be easy to catch. 

How about this as a safeguard (if possible)? I'll check if there is a `type`. And if there is already one, we can skip.

ZhouXing19: Yeah if we can reuse the logger (or context) for the same type that sounds much better!

Jeremyyang920: Actually the real question here is, is it even possible to call this 500 times within 1 process?

ZhouXing19: Hmm, I guess it's possible if we enable multiple fetch attempts in parallel?

ryanluu12345: @Jeremyyang920 but for calling it 500 times with 1 process, not on the same logger. If we have, say, 500 tables, then we would call 500 times because some of these `With`s happen within goroutines. But it would be calling it on the root logger as mentioned before.

Jeremyyang920: Ah right, then it should be fine since its created from the root one.

Jeremyyang920: We won't have that duplicate type `"type":"data","type":"data","type":"data"` thing like this from your example though right?

ZhouXing19: Agreed, the refactoring is not necessary at this moment, and backlogging it sounds good to me. Thanks for being patient with me!

ryanluu12345: Tracking here then: https://cockroachlabs.atlassian.net/browse/CC-26618

@Jeremyyang920 , we won't have the duplicated `type` so long as we only pass in the root logger, which will typically be the `logger` parameter in most methods.

When you create a `dataLogger`, just assign a new variable to the `GetDataLogger` call.

PR #103, opened 2023-12-12
verify: support TINYINT to BOOL  mapping

ryanluu12345:
Previously, there were many warnings about mismatching column definitions because there is not a direct translation from a MySQL TINYINT(1), which is used for booleans, and the CRDB BOOL value. This change aims to add support for this in comparable types.

Release Note: TINYINT(1) in MySQL now maps to BOOL types in Cockroach, which resolves a mismatching table definition error.

TODO:
- [ ] Investigate panic when comparing types that are different
- [ ] Add unit/integration tests

**Fetch Testing**
```
(⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % go run . fetch \    
  --source 'mysql://root:password@localhost/molt' \
  --target 'postgres://root@localhost:26257/defaultdb?sslmode=disable' --log-file fetch.log --direct-copy
{"level":"info","time":"2023-12-12T15:15:13-08:00","message":"default compression to none"}
{"level":"info","time":"2023-12-12T15:15:13-08:00","message":"checking database details"}
{"level":"info","source_table":"public.employees","target_table":"public.employees","time":"2023-12-12T15:15:13-08:00","message":"found matching table"}
{"level":"info","time":"2023-12-12T15:15:13-08:00","message":"verifying common tables"}
{"level":"info","time":"2023-12-12T15:15:13-08:00","message":"establishing snapshot"}
{"level":"info","type":"summary","num_tables":1,"cdc_cursor":"5e2fed98-d333-11ed-bc69-b9ae403342b1:1-172","time":"2023-12-12T15:15:13-08:00","message":"starting fetch"}
{"level":"warn","table":"public.employees","reason":"column type mismatch on unique_id: bytea vs uuid","time":"2023-12-12T15:15:13-08:00","message":"not migrating column {public employees} as it mismatches"}
{"level":"info","table":"public.employees","time":"2023-12-12T15:15:13-08:00","message":"data extraction phase starting"}
{"level":"info","table":"public.employees","type":"data","num_rows":100000,"time":"2023-12-12T15:15:14-08:00","message":"row import status"}
{"level":"info","table":"public.employees","type":"data","num_rows":200000,"time":"2023-12-12T15:15:15-08:00","message":"row import status"}
{"level":"info","table":"public.employees","type":"summary","num_rows":200000,"export_duration_ms":2720.051875,"export_duration":"000h 00m 02s","time":"2023-12-12T15:15:15-08:00","message":"data extraction from source complete"}
{"level":"info","type":"summary","num_tables":1,"tables":["public.employees"],"cdc_cursor":"5e2fed98-d333-11ed-bc69-b9ae403342b1:1-172","time":"2023-12-12T15:15:15-08:00","message":"fetch complete"}
```

We can see that it shows the relevant rows:
```
  id | unique_id |    name    |       created_at       | updated_at | is_hired | age | salary  | bonus
-----+-----------+------------+------------------------+------------+----------+-----+---------+---------
   1 | NULL      | Employee_1 | 2023-11-03 09:00:00+00 | 2023-11-03 |   true   |  24 | 5000.00 | 100.25
   2 | NULL      | Employee_2 | 2023-11-03 09:00:00+00 | 2023-11-03 |   true   |  24 | 5000.00 | 100.25
   3 | NULL      | Employee_3 | 2023-11-03 09:00:00+00 | 2023-11-03 |   true   |  24 | 5000.00 | 100.25
   4 | NULL      | Employee_4 | 2023-11-03 09:00:00+00 | 2023-11-03 |   true   |  24 | 5000.00 | 100.25
   5 | NULL      | Employee_5 | 2023-11-03 09:00:00+00 | 2023-11-03 |   true   |  24 | 5000.00 | 100.25
```

**Verify Testing**
Before
```
```

After
```
(⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % go run . verify \
  --source 'mysql://root:password@localhost/molt' \     
  --target 'postgres://root@localhost:26257/defaultdb?sslmode=disable' --log-file verify.log
{"level":"info","time":"2023-12-12T15:16:50-08:00","message":"verification in progress"}
{"level":"warn","type":"data","table_schema":"public","table_name":"employees","mismatch_info":"column type mismatch on unique_id: bytea vs uuid","time":"2023-12-12T15:16:50-08:00","message":"mismatching table definition"}
{"level":"info","time":"2023-12-12T15:16:50-08:00","message":"starting verify on public.employees, shard 1/1"}
{"level":"info","time":"2023-12-12T15:16:50-08:00","message":"verification complete"}
panic: unsupported comparison: bool to int
```

### DIFF
--- a/pgconv/datum_converter.go
+++ b/pgconv/datum_converter.go
@@ -39,7 +39,14 @@ func ConvertRowValue(typMap *pgtype.Map, val any, typOID oid.Oid) (tree.Datum, e
 
 	switch typOID {
 	case pgtype.BoolOID:
-		return tree.MakeDBool(tree.DBool(val.(bool))), nil
+		// Convert the boolean to an integer so that it's reasonable to compare with
+		// another TINYINT for MySQL and other dialects that don't have a BOOL.
+		intVal := 0
+		if val.(bool) {
+			intVal = 1
+		}
+
+		return tree.ParseDInt(fmt.Sprintf("%d", intVal))
 	case pgtype.QCharOID:
 		return tree.NewDString(fmt.Sprintf("%c", val.(int32))), nil
 	case pgtype.VarcharOID, pgtype.TextOID, pgtype.BPCharOID:
@@ -70,7 +77,7 @@ func ConvertRowValue(typMap *pgtype.Map, val any, typOID oid.Oid) (tree.Datum, e
 		if err != nil {
 			return nil, errors.Wrapf(err, "error decoding UUID %v", val)
 		}
-		return tree.NewDUuid(tree.DUuid{UUID: u}), nil
+		return tree.NewDString(u.String()), nil
 	case pgtype.TimestampOID:
 		return tree.MakeDTimestamp(val.(time.Time), time.Microsecond)
 	case pgtype.TimestamptzOID:
--- a/verify/tableverify/tableverify.go
+++ b/verify/tableverify/tableverify.go
@@ -302,5 +302,9 @@ func allowableTypes(a oid.Oid, b oid.Oid) bool {
 	if a == oid.T_varchar && b == oid.T_uuid {
 		return true
 	}
+	if a == oid.T_int2 && b == oid.T_bool {
+		return true
+	}
+
 	return false
 }
--- a/docker/cockroach-init/cockroach-init.sql
+++ b/docker/cockroach-init/cockroach-init.sql
@@ -5,6 +5,7 @@ CREATE TABLE employees (
 	created_at TIMESTAMPTZ,
 	updated_at DATE,
 	is_hired BOOL,
+	age INT2,
 	salary DECIMAL(8,2),
 	bonus FLOAT4
 );
--- a/docker/mysql-init/mysql-init.sql
+++ b/docker/mysql-init/mysql-init.sql
@@ -2,7 +2,7 @@ CREATE DATABASE molt;
 use molt;
 CREATE TABLE employees (
     id INT AUTO_INCREMENT PRIMARY KEY,
-    unique_id BINARY(16),
+    unique_id VARCHAR(100),
     name VARCHAR(50),
     created_at DATETIME,
     updated_at DATE,
@@ -23,7 +23,7 @@ BEGIN
     WHILE i <= 200000 DO
         INSERT INTO employees (unique_id, name, created_at, updated_at, is_hired, age, salary, bonus)
         VALUES (
-            UNHEX(REPLACE('550e8400-e29b-41d4-a716-446655440000', '-', '')),
+            '550e8400-e29b-41d4-a716-446655440000',
             CONCAT('Employee_', i),
             '2023-11-03 09:00:00',
             '2023-11-03',
--- a/docker/mysql-init/mysql-init.sql
+++ b/docker/mysql-init/mysql-init.sql
@@ -7,6 +7,7 @@ CREATE TABLE employees (
     created_at DATETIME,
     updated_at DATE,
     is_hired TINYINT(1),
+    age TINYINT(2),
     salary DECIMAL(8, 2),
     bonus FLOAT
 );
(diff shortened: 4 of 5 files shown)


ryanluu12345:
Hold on, I'm running into a really odd panic, probably because we are trying to convert a boolean to an integer.
```
(⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % go run . verify \
  --source 'mysql://root:password@localhost/molt' \     
  --target 'postgres://root@localhost:26257/defaultdb?sslmode=disable' --log-file verify.log
{"level":"info","time":"2023-12-12T15:16:50-08:00","message":"verification in progress"}
{"level":"warn","type":"data","table_schema":"public","table_name":"employees","mismatch_info":"column type mismatch on unique_id: bytea vs uuid","time":"2023-12-12T15:16:50-08:00","message":"mismatching table definition"}
{"level":"info","time":"2023-12-12T15:16:50-08:00","message":"starting verify on public.employees, shard 1/1"}
{"level":"info","time":"2023-12-12T15:16:50-08:00","message":"verification complete"}
panic: unsupported comparison: bool to int
```
Let me take some time to figure this out.

ryanluu12345:
I narrowed down the code issue to here: `mysqlconv/datum_converter.go` in `ConvertRowValue`. Essentially, we need to make a mapping rule so that if we see a boolean, we parse the value as a boolean.
```
func ConvertRowValue(typMap *pgtype.Map, val []byte, typOID oid.Oid) (tree.Datum, error) {
	fmt.Println(typOID)
	if val == nil {
		return tree.DNull, nil
	}
	switch typOID {
	case pgtype.BoolOID:
		fmt.Println(typOID, string(val))
		return tree.ParseDBool(string(val))
	case pgtype.VarcharOID, pgtype.TextOID:
```
Doing this doesn't work properly, because if we look deeper at the code in `pgtype`, we see that the column with `BOOL` is actually seen as: 21
```
Int2OID                = 2
```

ryanluu12345:
I put a draft idea up where we convert the BOOL to integer always. This makes it so that the comparison doesn't error out and panic here. But I don't like how the types don't end up matching how they're actually defined. The gist of this approach is that we force an apples to apples comparison (int to int) instead of apples (int) to oranges (bool).

An associated mismatch log looks like this:
```
{"level":"warn","type":"data","table_schema":"public","table_name":"employees","source_values":{"is_hired":"1"},"target_values":{"is_hired":"0"},"primary_key":["1"],"time":"2023-12-12T16:38:46-08:00","message":"mismatching row value"}
```
The value should have actually been `target: is_hired: false` and `source: is_hired: 1`

pgconv/datum_converter.go
```diff
@@ -70,7 +77,7 @@ func ConvertRowValue(typMap *pgtype.Map, val any, typOID oid.Oid) (tree.Datum, e
 		if err != nil {
 			return nil, errors.Wrapf(err, "error decoding UUID %v", val)
 		}
-		return tree.NewDUuid(tree.DUuid{UUID: u}), nil
+		return tree.NewDString(u.String()), nil
```
ryanluu12345, line 80: This change leads to `pgconv` datadrive test failures, because it's now treated as a `Dstring` type when it should actually be a `DUuid`. The solution here is to figure out a way to do heterogeneous comparisons on the data, because right now, the unsupported type panic is getting hit.

verify/tableverify/tableverify.go
```diff
@@ -302,5 +302,9 @@ func allowableTypes(a oid.Oid, b oid.Oid) bool {
 	if a == oid.T_varchar && b == oid.T_uuid {
 		return true
 	}
+	if a == oid.T_int2 && b == oid.T_bool {
```
ZhouXing19, line 305: nit: just did a quick scan, but should we allow the other direction as well? i.e. a == bool && b == int2.  Maybe same for the varchar and uuid comparison as well.

ZhouXing19:
I'm poking around this and would like to go down the path of casting the row value. i.e. if we saw `int <> bool`, cast the `tree.Datum` from a `tree.DBool` to a `tree.DInt`. Listing my finding here.

1. In the CRDB codebase, we have `eval.PerformCast()` that do exactly what we need. Unfortunately, it requires a notoriously tedious `EvalContext`, introducing which requires a lot of painful mocking.

2. A workaround is to construct a `tree.CastExpr` and run the stringified statement to obtained the casted value (i.e. run `SELECT <true/false>::int`).  Trying on this path now :)

ZhouXing19:
Fixed in https://github.com/cockroachdb/molt/pull/164, shall we close this one?

PR #106, opened 2023-12-13
verify: switched usage from Compare to CompareError to resolve panic

ryanluu12345:
TLDR; we prevent panics and mark the row now as a mismatch if columns mismatch.

Previously, if a comparison between heterogeneous types occurred, verify would panic and stop execution. With this change to use CompareError instead of Compare, verify no longer panics, but rather records the error and reports it to the logging and also marks the row as mismatched, which semantically is correct because even if values are the same, they are different types.

@rafiss recommended converting usage from `Compare` to `CompareError` so that we can handle the error on our side and figure out what to do when it happens.

Release Note: None

**Tests**
Before
```
{"level":"info","time":"2023-12-13T13:12:25-08:00","message":"starting verify on public.employees, shard 1/1"}
panic: unsupported comparison: uuid to string

goroutine 48 [running]:
github.com/cockroachdb/cockroachdb-parser/pkg/sql/sem/tree.(*DUuid).Compare(0x140006a2380?, {0x102c33ab0?, 0x103d99ee0?}, {0x102c43200?, 0x140008d1c20?})
        /Users/ryanluu/go/pkg/mod/github.com/cockroachdb/cockroachdb-parser@v0.0.0-20230705064001-302c9ad52e1a/pkg/sql/sem/tree/datum.go:1722 +0x50
github.com/cockroachdb/molt/verify/rowverify.verifyRows({0x102c33670, 0x103d99ee0}, {{0x102c36c20, 0x140006a21c0}, {0x102c36c20, 0x140006a2380}}, {{{{0x101e48959, 0x6}, {0x1400085f440, 0x9}}, ...}, ...}, ...)
        /Users/ryanluu/go/src/github.com/cockroachdb/molt/verify/rowverify/rowverify.go:159 +0x6b4
github.com/cockroachdb/molt/verify/rowverify.VerifyRowsOnShard({_, _}, {{0x102c3c660, 0x140004d3e90}, {0x102c3c330, 0x140005b8d80}}, {{{{0x101e48959, 0x6}, {0x1400085f440, 0x9}}, ...}, ...}, ...)
        /Users/ryanluu/go/src/github.com/cockroachdb/molt/verify/rowverify/rowverify.go:83 +0x664
github.com/cockroachdb/molt/verify.verifyRowShard({_, _}, {{0x102c3c660, 0x140004e2840}, {0x102c3c330, 0x14000867b80}}, {_, _}, {{_, _}, ...}, ...)
        /Users/ryanluu/go/src/github.com/cockroachdb/molt/verify/verify.go:307 +0x234
github.com/cockroachdb/molt/verify.Verify.func1()
        /Users/ryanluu/go/src/github.com/cockroachdb/molt/verify/verify.go:252 +0x4dc
golang.org/x/sync/errgroup.(*Group).Go.func1()
        /Users/ryanluu/go/pkg/mod/golang.org/x/sync@v0.2.0/errgroup/errgroup.go:75 +0x58
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 1
        /Users/ryanluu/go/pkg/mod/golang.org/x/sync@v0.2.0/errgroup/errgroup.go:72 +0x98
exit status 2
```

Now
```
{"level":"warn","type":"data","table_schema":"public","table_name":"employees","source_values":{"unique_id":"550e8400-e29b-41d4-a716-446655440000"},"target_values":{"unique_id":"550e8400-e29b-41d4-a716-446655440000"},"primary_key":["18250"],"time":"2023-12-13T13:11:18-08:00","message":"mismatching row value - unsupported comparison: uuid to string"}
{"level":"info","type":"summary","table_schema":"public","table_name":"employees","num_truth_rows":200000,"num_success":0,"num_missing":0,"num_mismatch":200000,"num_extraneous":0,"num_live_retry":0,"time":"2023-12-13T13:12:11-08:00","message":"finished row verification on public.employees (shard 1/1)"}
```

### DIFF
--- a/verify/inconsistency/stats.go
+++ b/verify/inconsistency/stats.go
@@ -8,14 +8,16 @@ import (
 
 // RowStats includes all details about the total rows processed.
 type RowStats struct {
-	Schema        string
-	Table         string
-	NumVerified   int
-	NumSuccess    int
-	NumMissing    int
-	NumMismatch   int
-	NumExtraneous int
-	NumLiveRetry  int
+	Schema                string
+	Table                 string
+	NumVerified           int
+	NumSuccess            int
+	NumConditionalSuccess int
+	NumMissing            int
+	NumMismatch           int
+	NumColumnMismatch     int
+	NumExtraneous         int
+	NumLiveRetry          int
 }
 
 func (s *RowStats) String() string {
@@ -33,14 +35,20 @@ func (s *RowStats) String() string {
 // reportRunningSummary reports the number of total rows and errors seen
 // during the execution of verify.
 func reportRunningSummary(l zerolog.Logger, s RowStats, m string) {
+	if s.NumConditionalSuccess > 0 {
+		m = fmt.Sprintf("%s - please check all warnings and errors for column mismatches to determine success", m)
+	}
+
 	l.Info().
 		Str("table_schema", s.Schema).
 		Str("table_name", s.Table).
 		Int("num_truth_rows", s.NumVerified).
 		Int("num_success", s.NumSuccess).
+		Int("num_conditional_success", s.NumConditionalSuccess).
 		Int("num_missing", s.NumMissing).
 		Int("num_mismatch", s.NumMismatch).
 		Int("num_extraneous", s.NumExtraneous).
 		Int("num_live_retry", s.NumLiveRetry).
+		Int("num_column_mismatch", s.NumColumnMismatch).
 		Msg(m)
 }
--- a/verify/inconsistency/stats.go
+++ b/verify/inconsistency/stats.go
@@ -36,7 +36,7 @@ func (s *RowStats) String() string {
 // during the execution of verify.
 func reportRunningSummary(l zerolog.Logger, s RowStats, m string) {
 	if s.NumConditionalSuccess > 0 {
-		m = fmt.Sprintf("%s - please check all warnings and errors for column mismatches to determine success", m)
+		m = fmt.Sprintf("%s - please check all warnings and errors to determine whether column mismatches can be ignored", m)
 	}
 
 	l.Info().
--- a/verify/rowverify/row_event_listener.go
+++ b/verify/rowverify/row_event_listener.go
@@ -79,9 +79,9 @@ func (n *defaultRowEventListener) OnColumnMismatchNoOtherIssues(
 	// so we don't double count mismatching columns and reporting for mismatching columns.
 	if reportLog {
 		n.reporter.Report(row)
-		numMismatchingCols := float64(int(len(row.MismatchingColumns)))
-		rowStatusMetric.WithLabelValues("mismatching_column").Add(numMismatchingCols)
-		n.stats.NumColumnMismatch += len(row.MismatchingColumns)
+		numMismatchingCols := len(row.MismatchingColumns)
+		rowStatusMetric.WithLabelValues("mismatching_column").Add(float64(numMismatchingCols))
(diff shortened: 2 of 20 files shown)


ryanluu12345:
Spoke with @Jeremyyang920 offline:
- Log only once per shard for this column mismatching issue
- Make a new type in reporting so that it increments a counter + logs out which one mismatched
- Report in the final count as num_column_mismatch

The benefits of this solution:
- Verify still continues and runs
- The exact issue and mapping difference is reported

ryanluu12345:
@taroface this is one of those cases where we'll need some copy updates and documentation about what these results mean. Which ticket should I add this to?

verify/rowverify/row_event_listener.go
```diff
@@ -71,6 +72,21 @@ func (n *defaultRowEventListener) OnMatch() {
...
 }
 
+func (n *defaultRowEventListener) OnColumnMismatchNoOtherIssues(
+	row inconsistency.MismatchingColumn, reportLog bool,
+) {
+	// This logic happens at most once per shard per table
+	// so we don't double count mismatching columns and reporting for mismatching columns.
+	if reportLog {
+		n.reporter.Report(row)
+		numMismatchingCols := float64(int(len(row.MismatchingColumns)))
```
Jeremyyang920, line 82: Is the int cast extra? Len should already be an int so you just need the float64.

Jeremyyang920 (approved):
Changes LGTM, barring any copy changes to error messages we may want to make.

ryanluu12345:
@taroface I went ahead and put the copy needs here as item 3: https://cockroachlabs.atlassian.net/browse/DOC-9223?filter=11743

verify/rowverify/row_event_listener.go
ryanluu12345: Addressed!

verify/inconsistency/stats.go
```diff
@@ -33,14 +35,20 @@ func (s *RowStats) String() string {
 // reportRunningSummary reports the number of total rows and errors seen
 // during the execution of verify.
 func reportRunningSummary(l zerolog.Logger, s RowStats, m string) {
+	if s.NumConditionalSuccess > 0 {
```
ryanluu12345, line 38: @taroface please let me know your thoughts on this copy. I basically want to communicate that conditional success means that the user must check the warnings and errors to see if what we flagged is actually an issue or not. A manual review step since we couldn't directly compare the column types.

taroface: does `for column mismatches` mean the user has to manually check whether there is a column mismatch, or does it mean that the warnings and errors intrinsically pertain to column mismatches that the user has to check to see whether they are actually an issue?

ryanluu12345: It means that we already marked some previous items as column mismatch and they need to manually verify if they are actually an issue. Your latter case.

verify/inconsistency/stats.go
```diff
@@ -33,14 +35,20 @@ func (s *RowStats) String() string {
...
+		m = fmt.Sprintf("%s - please check all warnings and errors for column mismatches to determine success", m)
```
taroface, line 39: Something like this?

```suggestion
		m = fmt.Sprintf("%s - check all warnings and errors to determine whether column mismatches are breaking", m)
```

taroface (commented):
Feel free to tweak my suggestion; I don't feel 100% on it

verify/inconsistency/stats.go
ryanluu12345: Done!

PR #107, opened 2023-12-15
verify: added in high level metrics for verify and refactored into the verifymetrics package

ryanluu12345:
Added in a new package responsible for centralizing high level metrics for verify. We now track more detailed types of metrics about the tables processed and their category: verified, missing, extraneous. Additionally, moved some high level metrics to this new package so that all metrics are colocated for other devs to review.

Resolves: CC-26355
Release Note: Added in new metrics for consumption in dashboards - molt_verify_num_tables (split up by category), molt_verify_duration_ms (to see how long a MOLT run took).

**Testing Results**
```
# HELP molt_verify_duration_ms Duration (in ms) for the verify run.
# TYPE molt_verify_duration_ms gauge
molt_verify_duration_ms 4760

molt_verify_num_tables{category="extraneous"} 0
molt_verify_num_tables{category="missing"} 1
molt_verify_num_tables{category="verified"} 1

molt_verify_num_row_fixups{category="extraneous"} 0
molt_verify_num_row_fixups{category="mismatching"} 0
molt_verify_num_row_fixups{category="missing"} 10992
```

**Existing Metrics (number of rows of each type**
```
molt_verify_row_verification_status{status="conditional_success"} 0
molt_verify_row_verification_status{status="extraneous"} 0
molt_verify_row_verification_status{status="mismatching"} 0
molt_verify_row_verification_status{status="mismatching_column"} 0
molt_verify_row_verification_status{status="missing"} 200000
molt_verify_row_verification_status{status="success"} 0
# HELP molt_verify_rows_read Rate of rows that are being read from source database.
# TYPE molt_verify_rows_read counter
molt_verify_rows_read 200000
```

### DIFF
--- a/cmd/verify/verify.go
+++ b/cmd/verify/verify.go
@@ -8,9 +8,12 @@ import (
 	"github.com/cockroachdb/molt/cmd/internal/cmdutil"
 	"github.com/cockroachdb/molt/moltlogger"
 	"github.com/cockroachdb/molt/retry"
+	"github.com/cockroachdb/molt/utils"
 	"github.com/cockroachdb/molt/verify"
 	"github.com/cockroachdb/molt/verify/inconsistency"
 	"github.com/cockroachdb/molt/verify/rowverify"
+	"github.com/cockroachdb/molt/verify/verifymetrics"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/cobra"
 )
 
@@ -92,6 +95,7 @@ func Command() *cobra.Command {
 			}
 
 			logger.Info().Msg("verification in progress")
+			timer := prometheus.NewTimer(prometheus.ObserverFunc(verifymetrics.OverallDuration.Set))
 			if err := verify.Verify(
 				ctx,
 				conns,
@@ -108,8 +112,11 @@ func Command() *cobra.Command {
 			); err != nil {
 				return errors.Wrapf(err, "error verifying")
 			}
-
-			logger.Info().Msg("verification complete")
+			verifyDuration := timer.ObserveDuration()
+			logger.Info().
+				Dur("net_duration_ms", verifyDuration).
+				Str("net_duration", utils.FormatDurationToTimeString(verifyDuration)).
+				Msg("verification complete")
 			return nil
 		},
 	}
--- a/verify/verifymetrics/metrics.go
+++ b/verify/verifymetrics/metrics.go
@@ -1,0 +1,104 @@
+package verifymetrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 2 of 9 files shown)


ryanluu12345:
Note for fetch:

1. fetchmetrics package where I centralize management and init of all metrics
2. Reporting helpers that will report different high level data
3. We can add any labels we want later on as we need them. Not spending much time on that right now because I just wanna set up our metrics collection infra.

verify/verifymetrics/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+const (
+	Namespace = "molt"
+	Subsystem = "verify"
+)
+
+var (
+	NumVerificationShards = promauto.NewGauge(prometheus.GaugeOpts{
```
ryanluu12345, line 14: @Jeremyyang920 open question, should I migrate all other metrics to this package? Or leave them colocated right now.

Jeremyyang920: in general, I liked my metrics colocated with the package that is using them. Its easier to find that way.

cmd/verify/verify.go
```diff
@@ -108,8 +111,12 @@ func Command() *cobra.Command {
 			); err != nil {
 				return errors.Wrapf(err, "error verifying")
 			}
-
-			logger.Info().Msg("verification complete")
+			verifyDuration := time.Since(verifyStart)
```
Jeremyyang920, line 114: Instead of this, lets follow the timer principal with promethues. 
https://github.com/prometheus/client_golang/blob/main/prometheus/example_timer_gauge_test.go

verify/verifymetrics/metrics.go
Jeremyyang920: Also i think naming this verifymetrics inside of the verify folder package is redundant.

Jeremyyang920 (approved):
Approving for now with the 2 minor comments to not block you on this as its not critical functionality.

cmd/verify/verify.go
ryanluu12345: I want to also be able to log out the details, which is why I want to keep this method.

verify/verifymetrics/metrics.go
ryanluu12345: Yeah good point. Let me move everything here and remove the stutter in the name.

cmd/verify/verify.go
ryanluu12345: We can revisit this for durations later on.

ryanluu12345: Nevermind, the method actually returns the duration. So I can use this timer principal and also log out the duration properly :) .

PR #112, opened 2024-01-12
docker/grafana: added initial cut for MOLT dashboard

ryanluu12345:
Added in MOLT fetch metrics which shows counts, rates, etc. by tables.

Made a variable that allows you to filter the search of the tables you want to look at.

Release Note: None

image

image

### DIFF
--- a/docker/grafana/dashboard.json
+++ b/docker/grafana/dashboard.json
@@ -1,0 +1,1051 @@
+{
+    "annotations": {
+      "list": [
+        {
+          "builtIn": 1,
+          "datasource": {
+            "type": "grafana",
+            "uid": "-- Grafana --"
+          },
+          "enable": true,
+          "hide": true,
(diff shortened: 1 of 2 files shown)


docker/grafana/dashboard.json
```diff
@@ -0,0 +1,1051 @@
...
+        },
+        "id": 14,
+        "panels": [],
+        "title": "Fetch Metrics",
+        "type": "row"
+      },
+      {
+        "datasource": {
+          "type": "prometheus",
+          "uid": "P1809F7CD0C75ACF3"
```
Jeremyyang920, line 42: You will need to change the UID to just "prometheus" like we do in the LMS dashboard, otherwise we cant import this very easily.

ryanluu12345: Addressed!

Jeremyyang920 (approved):

PR #115, opened 2024-01-12
docker/grafana: added dashboard section for verify

ryanluu12345:
Added in a new panel that has the top level metrics as single stats and then shows each of the row data as time visualization.

image

Release Note: None

### DIFF
--- a/docker/grafana/dashboard.json
+++ b/docker/grafana/dashboard.json
@@ -24,7 +24,7 @@
     "liveNow": false,
     "panels": [
       {
-        "collapsed": true,
+        "collapsed": false,
         "gridPos": {
           "h": 1,
           "w": 24,
@@ -1611,7 +1611,7 @@
             "type": "prometheus",
             "uid": "prometheus"
           },
-          "definition": "label_values(molt_fetch_rows_exported,table)",
+          "definition": "label_values({__name__=~ \"molt_verify_rows_read|molt_fetch_rows_exported\"}, table)",
           "description": "Table to filter by",
           "hide": 0,
           "includeAll": true,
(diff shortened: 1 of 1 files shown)


docker/grafana/dashboard.json
```diff
@@ -1009,9 +1603,9 @@
         {
           "allValue": ".*",
           "current": {
-            "selected": false,
-            "text": "public.test_table_large",
-            "value": "public.test_table_large"
+            "selected": true,
+            "text": "All",
+            "value": "$__all"
```
Jeremyyang920, line 1608: I can't comment on the actual line I want to because its not in the PR, but the way you are populating the molt_fetch_rows exported metric, but what if they only ran verify without fetch. That metric would not have been emitted and you would have no matching labels.

ryanluu12345: Fixed this with the latest commit! For broader vis, this was the recommendation that fixed it: https://stackoverflow.com/questions/60825749/combine-label-values-from-2-different-metrics-grafana

image

docker/grafana/dashboard.json
```diff
@@ -1009,15 +1603,15 @@
         {
           "allValue": ".*",
           "current": {
-            "selected": false,
-            "text": "public.test_table_large",
-            "value": "public.test_table_large"
+            "selected": true,
+            "text": "public.employees",
+            "value": "public.employees"
```
Jeremyyang920, line 1608: This should probably not be the default value.

ryanluu12345: Fixed

Jeremyyang920 (approved):
lgtm, we can adjust the board as needed as we use it more

PR #118, opened 2024-01-18
fetch: added in status and exception database helpers

ryanluu12345:
The helpers include mapping structs, helpers to create and update resources in the database, constants to represent different states.

Resolves: CC-26805
Release Note: None

**Tested with this bit of code**
```
 // TODO: logic for seeing source dialect
targetPgxConn := conns[1].(*dbconn.PGConn).Conn
if err := status.CreateStatusAndExceptionTables(ctx, targetPgxConn); err != nil {
    return err
}
fetchStatus := status.FetchStatus{
    Name:          "the-best",
    Status:        status.StatusInProgress,
    StartedAt:     time.Now(),
    SourceDialect: "PostgreSQL",
}
if err := fetchStatus.Create(ctx, targetPgxConn); err != nil {
    return err
}
fmt.Println(fetchStatus.ID)
fetchException := status.ExceptionLog{
    FetchID:  fetchStatus.ID,
    FileName: "yo.log",
    Table:    "employees",
    Schema:   "public",
    Message:  "it just all failed",
    SQLState: 1000,
    Command:  "SELECT *",
    Time:     time.Now(),
}
if err := fetchException.Create(ctx, targetPgxConn); err != nil {
    return err
}
fmt.Println(fetchException.ID)
fetchStatus.MarkFailed(ctx, targetPgxConn)
```

Results look like this in the database:
```
root@:26257/defaultdb> SELECT * FROM _molt_fetch_exception;
                   id                  |               fetch_id               | table_name | schema_name |      message       | sql_state | file_name | command  |            time
---------------------------------------+--------------------------------------+------------+-------------+--------------------+-----------+-----------+----------+-----------------------------
  70e973c1-8382-4c25-a407-9eaaa17e331e | f4ed9eae-1ba6-47a6-86ba-eee273633931 | employees  | public      | it just all failed |      1000 | yo.log    | SELECT * | 2024-01-17 16:29:50.234931

root@:26257/defaultdb> SELECT * FROM _molt_fetch_status;
                   id                  |   name   |   status    |         started_at         | finished_at | source_dialect
---------------------------------------+----------+-------------+----------------------------+-------------+-----------------
  dfaf7f39-520c-4860-9c2d-76bc88de7e2f | the-best | IN PROGRESS | 2024-01-17 16:29:26.172858 | NULL        | PostgreSQL
  f4ed9eae-1ba6-47a6-86ba-eee273633931 | the-best | IN PROGRESS | 2024-01-17 16:29:50.232368 | NULL        | PostgreSQL


root@:26257/defaultdb> SELECT * FROM _molt_fetch_status ORDER BY started_at DESC;
                   id                  |   name   |   status    |         started_at         |        finished_at         | source_dialect
---------------------------------------+----------+-------------+----------------------------+----------------------------+-----------------
  26788893-8f7b-466d-a580-5e9b85e2f898 | the-best | FAILED      | 2024-01-17 16:38:15.82192  | 2024-01-17 16:38:15.827129 | PostgreSQL
  ebb0b30a-f213-4d0c-bf93-dd571273e593 | the-best | SUCCEEDED   | 2024-01-17 16:37:21.293364 | 2024-01-17 16:37:21.301917 | PostgreSQL
  f4ed9eae-1ba6-47a6-86ba-eee273633931 | the-best | IN PROGRESS | 2024-01-17 16:29:50.232368 | NULL                       | PostgreSQL
  dfaf7f39-520c-4860-9c2d-76bc88de7e2f | the-best | IN PROGRESS | 2024-01-17 16:29:26.172858 | NULL                       | PostgreSQL
```

### DIFF
--- a/fetch/status/exceptions.go
+++ b/fetch/status/exceptions.go
@@ -1,0 +1,66 @@
+package status
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/cockroachdb/cockroachdb-parser/pkg/util/uuid"
+	"github.com/jackc/pgx/v5"
+)
+
+const createExceptionsTable = `CREATE TABLE IF NOT EXISTS _molt_fetch_exception (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+	fetch_id UUID NOT NULL REFERENCES _molt_fetch_status (id),
+    table_name STRING,
+    schema_name STRING,
+    message STRING,
+    sql_state INT,
+    file_name STRING,
+    command STRING,
+    time TIMESTAMP,
+	INDEX(fetch_id, sql_state)
+);
+`
+
+type ExceptionLog struct {
+	ID       uuid.UUID
+	FetchID  uuid.UUID
+	Table    string
+	Schema   string
+	Message  string
+	SQLState int
+	FileName string
+	Command  string
+	Time     time.Time
+}
+
+func (e *ExceptionLog) CreateEntry(ctx context.Context, conn *pgx.Conn) error {
+	curTime := time.Now().UTC()
+	query := `INSERT INTO _molt_fetch_exception (fetch_id, table_name, schema_name, message, sql_state, file_name, command, time) VALUES(@fetch_id, @table_name, @schema_name, @message, @sql_state, @file_name, @command, @time) RETURNING id`
+	args := pgx.NamedArgs{
+		"fetch_id":    e.FetchID,
+		"table_name":  e.Table,
+		"schema_name": e.Schema,
+		"message":     e.Message,
+		"command":     e.Command,
+		"time":        curTime,
+	}
+
+	if e.SQLState > 0 {
+		args["sql_state"] = e.SQLState
+	}
+
+	if strings.TrimSpace(e.FileName) != "" {
+		args["file_name"] = e.FileName
+	}
+
+	row := conn.QueryRow(ctx, query, args)
+
+	if err := row.Scan(&e.ID); err != nil {
+		return err
+	}
+	e.Time = curTime
+
+	return nil
+}
--- a/fetch/status/status.go
+++ b/fetch/status/status.go
@@ -1,0 +1,79 @@
+package status
+
+import (
+	"context"
+	"time"
+
+	"github.com/cockroachdb/cockroachdb-parser/pkg/util/uuid"
+	"github.com/jackc/pgx/v5"
+)
+
+const (
+	StatusInProgress = "IN PROGRESS"
+	StatusFailed     = "FAILED"
+	StatusSucceeded  = "SUCCEEDED"
+)
+
+const createStatusTable = `CREATE TABLE IF NOT EXISTS _molt_fetch_status (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    name STRING,
+    status STRING,
+    started_at TIMESTAMP,
+    finished_at TIMESTAMP,
+    source_dialect STRING
+);
+`
+
+type FetchStatus struct {
+	ID            uuid.UUID
+	Name          string
+	Status        string
+	StartedAt     time.Time
+	FinishedAt    time.Time
+	SourceDialect string
+}
+
+func (s *FetchStatus) CreateEntry(ctx context.Context, conn *pgx.Conn) error {
+	startTime := time.Now().UTC()
+	query := `INSERT INTO _molt_fetch_status (name, status, started_at, source_dialect) VALUES(@name, @status, @started_at, @source_dialect) RETURNING id, status`
+	args := pgx.NamedArgs{
+		"name":           s.Name,
+		"source_dialect": s.SourceDialect,
+		"status":         StatusInProgress,
+		"started_at":     startTime,
+	}
+	row := conn.QueryRow(ctx, query, args)
+
+	if err := row.Scan(&s.ID, &s.Status); err != nil {
+		return err
+	}
+
+	s.StartedAt = startTime
+	return nil
+}
+
+func (s *FetchStatus) markComplete(ctx context.Context, conn *pgx.Conn, status string) error {
+	endTime := time.Now().UTC()
+	query := `UPDATE _molt_fetch_status SET status=@status, finished_at=@finished_at WHERE id=@id`
+	args := pgx.NamedArgs{
+		"id":          s.ID,
+		"status":      status,
+		"finished_at": endTime,
+	}
+
+	if _, err := conn.Exec(ctx, query, args); err != nil {
+		return err
+	}
+
+	s.Status = status
+	s.FinishedAt = endTime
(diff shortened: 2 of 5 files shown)


fetch/status/exceptions.go
```diff
@@ -0,0 +1,71 @@
...
+	Table    string
+	Schema   string
+	Message  string
+	SQLState int
+	FileName string
+	Command  string
+	Time     time.Time
+}
+
+func (e *ExceptionLog) Create(ctx context.Context, conn *pgx.Conn) error {
```
ZhouXing19, line 38: nit: Rather than `Create`, i wonder if `InsertEntry` would be more explicit.

fetch/status/status.go
```diff
@@ -0,0 +1,82 @@
...
+type FetchStatus struct {
+	ID            uuid.UUID
+	Name          string
+	Status        string
+	StartedAt     time.Time
+	FinishedAt    time.Time
+	SourceDialect string
+}
+
+func (s *FetchStatus) Create(ctx context.Context, conn *pgx.Conn) error {
```
ZhouXing19, line 36: ditto for the func name

go.mod
```diff
@@ -54,6 +54,7 @@ require (
 	github.com/dustin/go-humanize v1.0.0 // indirect
 	github.com/getsentry/sentry-go v0.12.0 // indirect
 	github.com/go-ole/go-ole v1.3.0 // indirect
+	github.com/gofrs/uuid v4.4.0+incompatible // indirect
```
ZhouXing19, line 57: Rather than importing the generic uuid lib, I think we should use `github.com/cockroachdb/cockroachdb-parser/pkg/util/uuid` to be aligned with the dependency of crdb.

ZhouXing19 (changes requested):
just nits

fetch/status/exceptions.go
Jeremyyang920: I think CreateEntry would work as well.

fetch/status/exceptions.go
```diff
@@ -0,0 +1,71 @@
...
+
+	if e.SQLState > 0 {
+		args["sql_state"] = e.SQLState
+	}
+
+	if strings.TrimSpace(e.FileName) != "" {
+		args["file_name"] = e.FileName
+	}
+
+	var returnedId []byte
```
Jeremyyang920, line 57: nit: returnedID

go.mod
Jeremyyang920: Are there any benefits from the cockroach uuid library over this? I don't have a preference since we use the generic one in managed-service. So different repos are already diverged in terms of dependencies.

fetch/status/status.go
```diff
@@ -0,0 +1,82 @@
...
+	s.ID = id
+	return nil
+}
+
+func (s *FetchStatus) markComplete(ctx context.Context, conn *pgx.Conn, status string) error {
+	query := `UPDATE _molt_fetch_status SET status=@status, finished_at=@finished_at WHERE id=@id`
+	args := pgx.NamedArgs{
+		"id":          s.ID,
+		"status":      status,
+		"finished_at": time.Now(),
```
Jeremyyang920, line 65: Can you add an explicit .UTC() to the end of this. And lets make sure in the rest of our repo we are consistent on UTC time so we dont have any weird timezone issues.

go.mod
ZhouXing19: The crdb uuid lib has some customized settings for format: https://github.com/cockroachdb/cockroach/blob/6e499de9245d6bc3ac5a5a858f3946bc896134b7/pkg/util/uuid/uuid.go#L34-L35 

which is added in 2022. 

> we use the generic one in managed-service

Oh I didn't know this. Seems that the usage of the generic uuid in managed-service dated back to as early as 2020 ([example](https://github.com/cockroachlabs/managed-service/blame/b541dea35f8023b86b1bf1e57c1908f48b2b3bae/console/server/admin/clusters.go#L59-L64)) while the `cockroachdb-parser` lib is extracted in 2022. I suppose they didn't pay attention to this when writing the code for managed-service. It might not be an issue but to be more prudent I think all packages are supposed to be aligned in terms of dependency.

fetch/status/exceptions.go
ryanluu12345: Yeah I also agree that either of those would be more appropriate. Maybe I'll name it `CreateEntry`. It's a good middle ground.

go.mod
ryanluu12345: So for newer code written, is it preferrable to use the CRDB uuid library? And are there other codebases that use this? I can certainly do that. It allows us to not explicitly install the gofrs/uuid.

ZhouXing19: > So for newer code written, is it preferrable to use the CRDB uuid library?

Personally yes, but would like to hear from Jeremy as well.

> And are there other codebases that use this? I can certainly do that. 

Not sure if i can exhaust them but we use it in LMS: https://github.com/cockroachlabs/crdb-proxy/blob/6882581fd6e6d6f648184617195b5f425494234b/mysqlproxy/internal/mysqlbackend/mysql_result.go#L25

Jeremyyang920: Lets go with CRDB uuid one as were dealing with UUID rows generated by CRDB lets not have any chance of anything weird happening as well as standardizing across our MOLT tooling (SCT aside)

fetch/status/exceptions.go
ryanluu12345: Done!

fetch/status/status.go
ryanluu12345: Done!

ryanluu12345:
Note that I'm going to be working on adding some test cases to ensure that these helpers work as expected. This will also serve as documentation of how it should be used.

go.mod
ryanluu12345: Yep went with that. I also go mod tidy'd so we don't have the gofrs/uuid dep anymore.

ZhouXing19: Thanks!

fetch/status/exceptions.go
```diff
@@ -0,0 +1,71 @@
...
+	Schema   string
+	Message  string
+	SQLState int
+	FileName string
+	Command  string
+	Time     time.Time
+}
+
+func (e *ExceptionLog) CreateEntry(ctx context.Context, conn *pgx.Conn) error {
+	query := `INSERT INTO _molt_fetch_exception (fetch_id, table_name, schema_name, message, sql_state, file_name, command, time) VALUES(@fetch_id, @table_name, @schema_name, @message, @sql_state, @file_name, @command, @time) RETURNING id`
```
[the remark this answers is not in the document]
ryanluu12345: Good catch. Lemme fix it to call time.Now for both.

ZhouXing19 (approved):
Thank you for being patient with me!

Jeremyyang920 (approved):

PR #119, opened 2024-01-19
[CC-26909] fetch: Add CreateFromContinuationPoint to datablob interface

Jeremyyang920:
This commit adds the CreateFromContinuationPoint functionality to the S3 and GCP data stores. This allows us to return the list of files that should be proceseed for a subsequent run of fetch after an exception occured.

Release note: None


go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/cockroachdb/datadriven v1.0.3-0.20230801171734-e384cf455877
 	github.com/cockroachdb/errors v1.9.1
 	github.com/go-sql-driver/mysql v1.7.2-0.20230527164328-99976f4f587d
+	github.com/googleapis/google-cloud-go-testing v0.0.0-20210719221736-1c9a4c676720
```
Jeremyyang920, line 16: This dependency is technically archived, but because its mostly just interfaces for testing, I think its fine to bring in. Google was pushing to not use mocks at all for their testing which is evident with the way the go SDK is structured.

fetch/datablobstorage/gcp_test.go
```diff
@@ -34,3 +39,68 @@ func TestGCPResource_ImportURL(t *testing.T) {
...
+func (it *gcpObjectITMock) Next() (a *storage.ObjectAttrs, err error) {
+	if it.i == len(it.next) {
+		err = iterator.Done
+		return
+	}
+
+	a = &it.next[it.i]
+	it.i += 1
+	return
+}
```
Jeremyyang920, line 76: The reason for needing to mock this way is due to the way GCP calls need to be made and how dependency injection doesnt really work cleanly.

`it := client.Bucket(bucket).Objects(ctx, &storage.Query{
		Prefix: prefix,
		// The StartOffeset parameter is similar to the StartAfter flag
		// for S3 except that it is inclusive of the key so
		// we don't need to do any extra filtering of the
		// results.
		StartOffset: key,
	})`

Noticed the chained call of client.Bucket.Object() but we can really only pass in the root level client interface easily from the way the rest of the code is set up which makes sense. So in order for the Objects call to return the right thing we needed to mock Bucket and Objects to conform to their interfaces and then can mock the return value.

fetch/datablobstorage/gcp.go
```diff
@@ -59,6 +61,45 @@ func (s *gcpStore) CreateFromReader(
 	}, nil
 }
 
+func (s *gcpStore) CreateFromContinuationPoint(
+	ctx context.Context, table dbtable.VerifiedTable, fileName string,
+) ([]Resource, error) {
+	key := fmt.Sprintf("%s/%s", table.SafeString(), fileName)
+	prefix := table.SafeString()
+	if s.bucketPath != "" {
```
ryanluu12345, line 69: Could you add a short comment why this formatting is different in the case where bucket path is empty? Even as the person who implemented this flag it took me a second to remember why this logic needs to happen.

fetch/datablobstorage/gcp.go
```diff
@@ -59,6 +61,45 @@ func (s *gcpStore) CreateFromReader(
...
+	}
+	return createFromContinuationPointGCP(ctx, stiface.AdaptClient(s.client), key, prefix, s.bucket)
+}
+
+func createFromContinuationPointGCP(
+	ctx context.Context, client stiface.Client, key, prefix, bucket string,
+) ([]Resource, error) {
+	it := client.Bucket(bucket).Objects(ctx, &storage.Query{
+		Prefix: prefix,
+		// The StartOffeset parameter is similar to the StartAfter flag
```
ryanluu12345, line 81: Clean!

fetch/datablobstorage/gcp.go
```diff
@@ -59,6 +61,45 @@ func (s *gcpStore) CreateFromReader(
...
+		key = fmt.Sprintf("%s/%s/%s", s.bucketPath, table.SafeString(), fileName)
+		prefix = fmt.Sprintf("%s/%s", s.bucketPath, table.SafeString())
+	}
+	return createFromContinuationPointGCP(ctx, stiface.AdaptClient(s.client), key, prefix, s.bucket)
+}
+
+func createFromContinuationPointGCP(
+	ctx context.Context, client stiface.Client, key, prefix, bucket string,
+) ([]Resource, error) {
+	it := client.Bucket(bucket).Objects(ctx, &storage.Query{
```
ryanluu12345, line 79: Is there a chance that this could accidentally pick up extra or conflicting files? The scenario I'm thinking about is:
1. User runs full fetch and it succeeds, but the intermediate store is not cleaned up
2. User runs full fetch with same exact store path and it fails, leaving you with a continuation token; store is not cleaned up
3. Now the bucket has the files from the first and second runs. It's unclean and not representative of the data that needs to be moved.

fetch/datablobstorage/local.go
```diff
@@ -115,6 +115,12 @@ func (l *localStore) CreateFromReader(
 	}
 }
 
+func (l *localStore) CreateFromContinuationPoint(
```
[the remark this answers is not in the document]
ryanluu12345: If this is a TODO, please specify it as such and add a ticket.

fetch/datablobstorage/s3.go
```diff
@@ -124,6 +125,61 @@ func (s *s3Store) CreateFromReader(
...
 }
 
+// CreateFromContinuationPoint will create the list of s3 resources
+// that will be processed for this iteration of fetch. It uses the
+// passed in table name to construct the key and prefix to look at
+// in the s3 bucket.
+func (s *s3Store) CreateFromContinuationPoint(
+	ctx context.Context, table dbtable.VerifiedTable, fileName string,
+) ([]Resource, error) {
+	key := fmt.Sprintf("%s/%s", table.SafeString(), fileName)
```
ryanluu12345, line 135: Actually think that this logic to get the key/prefix/etc. could be wrapped up in a helper method so that you don't have to write this twice.

```
getKeyAndPrefix(table dbTable.name) (string, string) {
....
}
```

ryanluu12345 (changes requested):
Mostly small comments, but helpful for future clarity.

fetch/datablobstorage/gcp.go
Jeremyyang920: Yes. We name the files lexicographically so we are assuming that contract holds true.

fetch/datablobstorage/local.go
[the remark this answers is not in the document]
Jeremyyang920: Oh I forgot we had file path as a flag for local. I was thking it was tmp directory in which case I would have said we don't need to do it. But Ill add the logic for it.

fetch/datablobstorage/gcp.go
Jeremyyang920: If you ran it with the same parameters and the same file chunks, then all parts will be overwritten the second run.

The only issue that would happen is if the user chooses to do less chunks and something fails, as the contiuation token process would pick up all files (even ones that might not be part of the new process), but this seems more of a user error than something we can code against since we would never know.

fetch/datablobstorage/gcp.go
```diff
@@ -59,6 +61,45 @@ func (s *gcpStore) CreateFromReader(
 	}, nil
 }
 
+func (s *gcpStore) CreateFromContinuationPoint(
+	ctx context.Context, table dbtable.VerifiedTable, fileName string,
+) ([]Resource, error) {
+	key := fmt.Sprintf("%s/%s", table.SafeString(), fileName)
```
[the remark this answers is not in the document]
Jeremyyang920: Im not going to make the formatting a constant since its different from the one used to create a file and I moved the key prefix to a helper so its only used there once.

fetch/datablobstorage/local.go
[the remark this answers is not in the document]
Jeremyyang920: Implementation done. Going to add the mock os tests for it tomorrow.

fetch/datablobstorage/datablobstorage.go
```diff
@@ -2,13 +2,15 @@ package datablobstorage
...
 	"context"
+	"fmt"
 	"io"
 
 	"github.com/cockroachdb/molt/dbtable"
 )
 
 type Store interface {
 	CreateFromReader(ctx context.Context, r io.Reader, table dbtable.VerifiedTable, iteration int, fileExt string) (Resource, error)
+	CreateFromContinuationPoint(ctx context.Context, table dbtable.VerifiedTable, fileName string) ([]Resource, error)
```
ZhouXing19, line 13: nit: I was a bit confused by the name of this function at first and then realized this func doesn't actually create a new file in the bucket, but list out the files that are already on the bucket. 
I.e. `CreateFromReader` calls `s.client.Bucket(s.bucket).Object(key).NewWriter(ctx)` to create new files, but `CreateFromContinuationPoint` calls `client.Bucket(bucket).Objects()` to list out the existing ones. 

So just to confirm if I understood the story line here correctly: `CreateFromContinuationPoint` is called only when `exportTable()` has finished successfully, but `importTable()` failed from previous fetch attempt. We just want to have `importTable()` restarts from the continuation point for the next attempt.

If this is correct, I wonder if we can rename this function to `ListFromContinuationPoint`, to be in contrast with `CreateFromReader`. Also, comments inline to distinguish these 2 funcs would be helpful!

Jeremyyang920: Your understanding is correct. I wanted to keep the naming somewhat consistent, but I can change it to your suggestion.

Jeremyyang920: I added comments at the interface definition level. Let me know if that helps!

fetch/datablobstorage/gcp.go
ZhouXing19: > if the user chooses to do less chunks

Could you say more about this? In what way (i.e. is there an option) for a user to get into this state?

fetch/datablobstorage/datablobstorage.go
ZhouXing19: Yes it's crystal clear now, thanks!!

fetch/datablobstorage/gcp.go
Jeremyyang920: We allow the user to specify the flush rows or the file size. If the first run they chose to say flush ever 100 files, but then reran and now did every 200. We would now have half as many files, but if they didnt delete the files from the old run, we would pick those up as its in the directory still.

ZhouXing19: Makes sense, and seems to me that this could be a common user error. When hitting an error, user might tend to be more conservative and adjust the granularity. Tho not necessarily in this PR, it might be worth documenting or adding a hint for such possible hazard when it notices the bucket is not empty.

ZhouXing19 (approved):
LGTM, thanks for being patient with my questions!

fetch/datablobstorage/gcp.go
Jeremyyang920: I think this would be something we document with the Fetch docs. Or in the future it could be if we notice the bucket is not empty and its not a continuation run, then we first delete items in the bucket.

fetch/datablobstorage/gcp.go
ryanluu12345: Makes sense. We are the ones who enforce that too, so this works.

fetch/datablobstorage/datablobstorage.go
```diff
@@ -2,13 +2,25 @@ package datablobstorage
...
 )
 
 type Store interface {
+	// CreateFromReader is responsible for the creation of the individual
+	// CSVs from the data export process. It will create the file and upload
+	// it to the respetive data store and return the resource object which
+	// will be used in the data import phase.
 	CreateFromReader(ctx context.Context, r io.Reader, table dbtable.VerifiedTable, iteration int, fileExt string) (Resource, error)
+	// ListFromContinuationPoint is used when restarting Fetch from
+	// a contiuation point. It will query the respective data store
```
ryanluu12345, line 18: Super nit: mispelling of `continuation`

Jeremyyang920: fixed

fetch/datablobstorage/local.go
```diff
@@ -115,6 +115,30 @@ func (l *localStore) CreateFromReader(
...
+	baseDir := path.Join(l.basePath, table.SafeString())
+	fmt.Println(baseDir)
+	files, err := os.ReadDir(baseDir)
+	if err != nil {
+		return nil, err
+	}
+
+	resources := []Resource{}
+	for _, f := range files {
+		if f.Name() >= fileName {
```
ryanluu12345, line 130: TIL you can do a string comp this way

ryanluu12345 (approved):
LGTM thanks for this!

fetch/datablobstorage/local.go
```diff
@@ -115,6 +115,30 @@ func (l *localStore) CreateFromReader(
 	}
 }
 
+func (l *localStore) ListFromContinuationPoint(
+	ctx context.Context, table dbtable.VerifiedTable, fileName string,
+) ([]Resource, error) {
+	baseDir := path.Join(l.basePath, table.SafeString())
+	fmt.Println(baseDir)
```
ryanluu12345, line 122: Nit: Stray fmt.Println statement

fetch/datablobstorage/local.go
Jeremyyang920: You can do this in any language really since it just starts to compare the ascii values lexicographically.

PR #120, opened 2024-01-20
fetch: added an exception reporter helper and integrated in copy/direct copy/import

ryanluu12345:
Now there is a helper that parses the PG error, extracts key information necessary to store in the exception log, and writes it / links it to the fetch run. Also added in portions of the code to create the dependency tables and status runs so that the relations are correct.

Resolves: CC-26806
Release Note: None

Problems still needed to fix:
- Parsing out the log file name(s)
- Deduplicating when retries for a certain action (import or copy) happens
- Datadriven testing

Example entries:
```
root@:26257/defaultdb> SELECT * FROM _molt_fetch_exception;
                   id                  |               fetch_id               | table_name | schema_name |                                                                        message                                                                        | sql_state | file_name | command |            time
---------------------------------------+--------------------------------------+------------+-------------+-------------------------------------------------------------------------------------------------------------------------------------------------------+-----------+-----------+---------+-----------------------------
  1c0085ab-6d1b-437f-9545-5d0a0e831f85 | a318d73f-fad2-4d59-aa65-41ae497151f6 | employees  | public      | addsstable [/Table/104/2/1/0,/Table/104/2/200000/0/NULL): checking for key collisions: ingested key collides with an existing one: /Table/104/2/1/0;  |      NULL | table.log |         | 2024-01-20 00:16:12.31587
  9c198a32-a214-4781-87e0-acb34ee435f5 | a318d73f-fad2-4d59-aa65-41ae497151f6 | employees  | public      | addsstable [/Table/104/2/1/0,/Table/104/2/200000/0/NULL): checking for key collisions: ingested key collides with an existing one: /Table/104/2/1/0;  |      NULL | table.log |         | 2024-01-20 00:16:16.762109
  de042ddb-36ec-4ff8-aa00-da7fd483ee6f | a318d73f-fad2-4d59-aa65-41ae497151f6 | employees  | public      | addsstable [/Table/104/2/1/0,/Table/104/2/200000/0/NULL): checking for key collisions: ingested key collides with an existing one: /Table/104/2/1/0;  |      NULL | table.log |         | 2024-01-20 00:16:13.756314
  f49688e3-2887-4290-845e-12e609486e5a | a318d73f-fad2-4d59-aa65-41ae497151f6 | employees  | public      | addsstable [/Table/104/2/1/0,/Table/104/2/200000/0/NULL): checking for key collisions: ingested key collides with an existing one: /Table/104/2/1/0;  |      NULL | table.log |         | 2024-01-20 00:16:13.025836
(4 rows)
```


fetch/copy.go
```diff
@@ -56,7 +57,8 @@ func Copy(
 				r,
 				dataquery.CopyFrom(table),
 			); err != nil {
-				pgErr := status.MaybeReportException(ctx, logger, conn, table.Name, err)
+				fileName := key[strings.LastIndex(key, "/")+1:]
```
ZhouXing19, line 60: Can we use `path.Base()` to get the filename?

fetch/import_table.go
```diff
@@ -166,7 +166,7 @@ func importTable(
 			ctx,
 			importQuery,
 		); err != nil {
-			pgErr := status.MaybeReportException(ctx, logger, baseConn.(*dbconn.PGConn).Conn, table.Name, err)
+			pgErr := status.MaybeReportException(ctx, logger, baseConn.(*dbconn.PGConn).Conn, table.Name, err, "")
```
ZhouXing19, line 169: nit: add inline comment for this "" value with `""/* filename */`

fetch/status/exceptions.go
```diff
@@ -116,3 +127,9 @@ func MaybeReportException(
 
 	return inputErr
 }
+
+var fileNameRegEx = regexp.MustCompile(`(part.+csv?)`)
```
ZhouXing19, line 131: Should this be aligned with the file pattern used in the storage engine? For example, for gcp, it's `key := fmt.Sprintf("%s/part_%08d.%s", table.SafeString(), iteration, fileExt)`. So I wonder if we should be more strict with `%08d` in the file name regex too.

ZhouXing19 (changes requested):

fetch/copy.go
ryanluu12345: Yep using it now!

fetch/import_table.go
ryanluu12345: Done!

fetch/status/exceptions.go
ryanluu12345: I'm not sure this is necessary because we get everything between part and csv (inclusive). I don't see a case where the intermediary store will have something that doesn't match this exact pattern unless people manually tamper with it. I'm not too concerned here.

fetch/fetch.go
```diff
@@ -64,6 +93,7 @@ func Fetch(
 	}
 
 	if err := dbconn.RegisterTelemetry(conns); err != nil {
+		retErr = err
```
ZhouXing19, line 96: This is not needed -- since we have named `retErr error` as the returned value of this function, the returned error is automatically propagated as `retErr`.
Example in playground: https://go.dev/play/p/vQCz8EEtv-E

fetch/fetch.go
```diff
@@ -78,9 +108,11 @@ func Fetch(
 	logger.Info().Msgf("checking database details")
 	dbTables, err := dbverify.Verify(ctx, conns)
 	if err != nil {
+		retErr = err
```
ZhouXing19, line 111: ditto

fetch/fetch.go
```diff
@@ -78,9 +108,11 @@ func Fetch(
...
 		return err
 	}
 	if dbTables, err = dbverify.FilterResult(tableFilter, dbTables); err != nil {
+		retErr = err
```
ZhouXing19, line 115: ditto

fetch/fetch.go
```diff
@@ -103,11 +135,13 @@ func Fetch(
 	logger.Info().Msgf("verifying common tables")
 	tables, err := tableverify.VerifyCommonTables(ctx, conns, dbTables.Verified)
 	if err != nil {
+		retErr = err
```
ZhouXing19, line 138: ditto

fetch/fetch.go
```diff
@@ -103,11 +135,13 @@ func Fetch(
...
 		return err
 	}
 	logger.Info().Msgf("establishing snapshot")
 	sqlSrc, err := dataexport.InferExportSource(ctx, cfg.ExportSettings, conns[0])
 	if err != nil {
+		retErr = err
```
ZhouXing19, line 144: ditto

fetch/fetch.go
```diff
@@ -160,6 +194,7 @@ func Fetch(
 	}()
 
 	if err := g.Wait(); err != nil {
+		retErr = err
```
ZhouXing19, line 197: ditto

fetch/fetchcontext/fetchcontext.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/cockroachdb/cockroachdb-parser/pkg/util/uuid"
+)
+
+type fetchCtxKey int
+
+const key fetchCtxKey = iota
```
ZhouXing19, line 12: Is this the recommended practice to define a key as an int? Seems to me it can cause collision if other keys are defined in a similar way. According to the usages of `context.WithValue()` and `ctx.Value()` in the crdb code base (example [[1](https://github.com/cockroachdb/cockroach/blob/ba52733b8f92c17e8902de81591debf04d0fb076/pkg/sql/conn_executor.go#L4486)][[2](https://github.com/cockroachdb/cockroach/blob/ba52733b8f92c17e8902de81591debf04d0fb076/pkg/roachpb/tenant.go#L140)][[3](https://github.com/cockroachdb/cockroach/blob/ba52733b8f92c17e8902de81591debf04d0fb076/pkg/sql/conn_executor.go#L4491)]), we tend to use the actual struct obj as the key. 

So I wonder if we can do something similar here:

```go
type fetchKey struct{}

type FetchReq struct {
	RunID     uuid.UUID
	StartedAt time.Time
}

func WithFetchReq(ctx context.Context, req FetchReq) context.Context {
    return context.WithValue(ctx, fetchKey{}, req)
}

func FetchReqFromCtx(ctx context.Context) FetchReq {
	req := ctx.Value(fetchKey{})
	if req == nil {
		return FetchReq{}
	}
	return req.(FetchReq)
}
```

fetch/status/exceptions.go
```diff
@@ -64,3 +72,64 @@ func (e *ExceptionLog) CreateEntry(ctx context.Context, conn *pgx.Conn) error {
 
 	return nil
 }
+
+const excludedSqlState = "42704"
```
ZhouXing19, line 76: Can we add some comments inline for this exception? IIUC, this error code stands for `undefined_object`, but in which cases do we see this error?

fetch/status/exceptions.go
ZhouXing19: OK sounds good to me!

ZhouXing19 (approved):
Mostly nit comments module the context key one. Will let Jeremy to finalize the review.

fetch/copy.go
```diff
@@ -55,7 +57,9 @@ func Copy(
 				r,
 				dataquery.CopyFrom(table),
 			); err != nil {
-				return err
+				fileName := path.Base(key)
+				pgErr := status.MaybeReportException(ctx, logger, conn, table.Name, err, fileName)
+				return pgErr
```
Jeremyyang920, line 62: nit: since we dont use pgErr again, lets just do `return status.MaybeReportException`

fetch/fetch.go
```diff
@@ -321,3 +356,25 @@ func reportTelemetry(
...
+) (*status.FetchStatus, error) {
+	// Setup the status and exception tables.
+	if err := status.CreateStatusAndExceptionTables(ctx, conn); err != nil {
+		return nil, err
+	}
+
+	createdAt := time.Now().UTC()
+	fetchStatus := &status.FetchStatus{
+		Name:          fmt.Sprintf("run at %d", createdAt.Unix()),
+		Status:        status.StatusInProgress,
```
Jeremyyang920, line 371: Are we still going with Status based on comments from Adam on the design doc?

fetch/status/exceptions.go
```diff
@@ -64,3 +72,64 @@ func (e *ExceptionLog) CreateEntry(ctx context.Context, conn *pgx.Conn) error {
...
+		FetchID:  fetchContext.RunID,
+		Table:    table.Table.String(),
+		Schema:   table.Schema.String(),
+		Message:  errMsg,
+		SQLState: sqlState,
+		FileName: fileName,
+		Time:     createdAt,
+	}
+
+	// TODO: figure out to deduplicate or upsert and do nothing if conflict.
```
Jeremyyang920, line 122: Not needed to be addressed here, but how can we duplicate rows if createdAt is a time.Now() column.

fetch/fetch_test.go
```diff
@@ -187,3 +189,32 @@ func TestDataDriven(t *testing.T) {
 		})
 	}
 }
+
+func TestInitStatusEntry(t *testing.T) {
+	ctx := context.Background()
+	dbName := "fetch_test_status"
+
+	t.Run("successfully initiliazed when tables not created", func(t *testing.T) {
```
Jeremyyang920, line 197: nit: initialized spelling

fetch/fetch_test.go
```diff
@@ -187,3 +189,32 @@ func TestDataDriven(t *testing.T) {
...
+		require.NoError(t, err)
+		pgConn := conn.(*dbconn.PGConn).Conn
+
+		actual, err := initStatusEntry(ctx, pgConn, "PostgreSQL")
+		require.NoError(t, err)
+		require.NotEqual(t, uuid.Nil, actual.ID)
+		require.Equal(t, status.StatusInProgress, actual.Status)
+	})
+
+	t.Run("successfully initiliazed when tables created beforehand", func(t *testing.T) {
```
Jeremyyang920, line 208: same spelling note as above

fetch/status/exceptions.go
```diff
@@ -64,3 +72,64 @@ func (e *ExceptionLog) CreateEntry(ctx context.Context, conn *pgx.Conn) error {
...
+	if !errors.As(inputErr, &pgErr) || pgErr.Code == excludedSqlState {
+		return inputErr
+	}
+
+	errMsg := fmt.Sprintf("%s; %s", pgErr.Message, pgErr.Detail)
+	sqlState := -1
+	if pgErr.Code != "" {
+		if val, err := strconv.Atoi(pgErr.Code); err == nil {
+			sqlState = val
+		}
```
Jeremyyang920, line 101: Did we store status code as an int in the table? I dont't think we can do that. Some codes have characters in them
https://github.com/cockroachdb/cockroach/blob/786ed22e6c319b8d28399f8ee1f82b88efe7c1ef/pkg/sql/pgwire/pgcode/codes.go#L326

Jeremyyang920 (commented):
Mostly looks good to me, barring the comment from Jane about the ctx that needs addressing.

fetch/fetch.go
```diff
@@ -45,7 +48,33 @@ func Fetch(
 	conns dbconn.OrderedConns,
 	blobStore datablobstorage.Store,
 	tableFilter dbverify.FilterConfig,
-) error {
+) (retErr error) {
+	// Setup fetch status tracking.
+	targetPgxConn := conns[1].(*dbconn.PGConn).Conn
```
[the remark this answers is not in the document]
ryanluu12345: Good suggestion! Let me do this!

fetch/fetch.go
ryanluu12345: Thanks for the note here. Makes perfect sense to me!

fetch/fetchcontext/fetchcontext.go
ryanluu12345: Yep, I like this method better. I went ahead and made the change.

fetch/status/exceptions.go
ryanluu12345: @Jeremyyang920 I took this bit from your testing. Wondering if you have insights on when you hit this.

fetch/copy.go
ryanluu12345: Good point!

fetch/fetch.go
ryanluu12345: I plan to discuss that in our weekly today. For now I'm keeping it but we can certainly remove this later.

fetch/status/exceptions.go
ryanluu12345: By duplicate I mean on the retries that `import` does by default. It creates a new row for each of these. So I mean deduplicate the error message + details. Not the time it occurred, because technically they are 3 different retries.

fetch/fetch_test.go
ryanluu12345: Thanks!

fetch/status/exceptions.go
Jeremyyang920: I took that from code that exists inside Fetch already.

fetch/status/exceptions.go
ryanluu12345: Yep, good point. I ended up switching it from an int to a string and change the mapping columns everywhere too. Works!

fetch/status/exceptions.go
ryanluu12345: @ZhouXing19 I left a comment for what it represents. Let's just leave it at that.

Jeremyyang920 (approved):

PR #121, opened 2024-01-24
fetch: added in testing knobs to simulate corrupted CSV failures

ryanluu12345:
Added in testing knobs structure and formatting so that we can properly simulate errors. Currently, the corrupt CSV failure mode is simulated for better testing.

Resolves: CC-26918
Release Note: None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,3 +14,6 @@ build_molt_cli:
 
 sync_hooks:
 	cp -a .githooks/ .git/hooks/
+
+rewrite_fetch_tests:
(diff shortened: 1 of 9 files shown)


ZhouXing19 (approved):
LGTM, but seems only pg datadriven tests are updated. Do we plan to update the mysql and crdb ones in another PR?

ryanluu12345:
> LGTM, but seems only pg datadriven tests are updated. Do we plan to update the mysql and crdb ones in another PR?

The code paths tested should be agnostic of dialect. I'm fine keeping it just to pg since it tests the code path for exceptions logging.

Jeremyyang920 (approved):

PR #122, opened 2024-01-25
[CC-26908] fetch: Add new flags for continuations

Jeremyyang920:
This commit adds the 3 new flags that will be needed for the continuation process work. Fetch-id is the parent flag and is a dependency for the child flags continuation-token and continuation-file-name.

Release note: None

### DIFF
--- a/cmd/fetch/fetch.go
+++ b/cmd/fetch/fetch.go
@@ -19,6 +19,11 @@ import (
 )
 
 func Command() *cobra.Command {
+	const (
+		fetchID              = "fetch-id"
+		continuationToken    = "continuation-token"
+		continuationFileName = "continuation-file-name"
+	)
 	var (
 		s3Bucket                string
 		gcpBucket               string
@@ -34,6 +39,17 @@ func Command() *cobra.Command {
 		Use:   "fetch",
 		Short: "Moves data from source to target.",
 		Long:  `Imports data from source directly into target tables.`,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// Ensure that if continuation-token is set that fetch-id is set
+			if err := cmdutil.CheckFlagDependency(cmd, fetchID, []string{continuationToken}); err != nil {
+				return err
+			}
+			// Ensure if continuation-file-name is set that continuation-token is set.
+			if err := cmdutil.CheckFlagDependency(cmd, continuationToken, []string{continuationFileName}); err != nil {
+				return err
+			}
+			return nil
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := context.Background()
 			logger, err := moltlogger.Logger(logFile)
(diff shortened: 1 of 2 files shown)


cmd/fetch/fetch.go
```diff
@@ -19,6 +19,11 @@ import (
 )
 
 func Command() *cobra.Command {
+	const (
+		fetchID              = "fetch-id"
```
ryanluu12345, line 23: What is the behavior if a continuation token is not passed in? And also, wouldn't there be a continuation token per table? Should we give a way to define a mapping for table:token? Or should continuation-token be a string slice flag?

Jeremyyang920: Its documented in the doc. If no conttinuation token is passed in then we look up all rows in the exceptions table with fetchID equal to the one passed in and restart all tables with it.

Jeremyyang920: > When users restart fetch, they can supply the task UUID with the flag `fetch-id` which will retry all tables from the continuation point and they can optionally supply the continuation token with the flag `continuation-token` which will only try to load a specific table again.

Jeremyyang920: Since the fetchID is unique per run of fetch this is fine since we would look up the records from the last run only.

ryanluu12345: I think it's worth leaving a comment here about that behavior. Or putting that in the README, so people know the expectation.

Overall, I think it'd be worthwhile to have a `continuation` README section under `fetch`.

ryanluu12345: Ok so to ensure I'm understanding: if a specific continuation token is passed, then it will just do one table? The table that is mapped to it in the exceptions table? Why not support a list of tokens so that people can do a subset of tables?

Jeremyyang920: We can add the comments once the rest of the feature is done. This is just scaffolding up some of the parts needed. Behavior can change later on and we will need Fetch docs written once everything is finalized.

Jeremyyang920: If you pass fetch-id and continuation token, its going to resume for only that table as the continuation token is unique per entry. We can choose to make it a list of tokens. Its not a problem. I went with what was written down first.

ryanluu12345: Ok sounds good. Let's do it that way. Shouldn't be too hard to extend later. Then this lgtm.

cmd/fetch/fetch.go
```diff
@@ -244,6 +260,24 @@ func Command() *cobra.Command {
...
+		"",
+		"If set, restarts the fetch process for all failed tables of the given ID",
+	)
+	cmd.PersistentFlags().StringVar(
+		&cfg.ContinuationToken,
+		continuationToken,
+		"",
+		"If set, restarts the fetch process for the given continuation token for a specific table",
+	)
+	cmd.PersistentFlags().StringVar(
```
ryanluu12345, line 275: One thing I was thinking of is, how do you differentiate between `part_000001.csv` for one table vs. another. There could be the case where `part_000001.csv` fails for multiple tables

Jeremyyang920: Well if we support single table only its not a problem. If we support a list of continuation tokens then this has to change to a map type

ryanluu12345: How do you anticipate this will be used? Will this always be set with the continuation token? What's the point of passing in the file name and the continuation token if you can just pass the file name?

Jeremyyang920: Its to look up the rest of the information with that continuation token. EG table and all the other stuff of where the file is stored.

ryanluu12345: Wait don't you just need the continuation token itself? That's id for the exceptions log table, where you can query rest of it

Jeremyyang920: The file name flag here is for if someone wants to override the one we stored in the exceptions table, but we still need to lookup the entry for the given continuation token and fetchid.

ryanluu12345: Ok got it. As long as fetch ID and continuation token are passed in, it should be ok.

ryanluu12345 (approved):

PR #123, opened 2024-01-25
fetch: added in test cases for various constraints

ryanluu12345:
Increased test coverage to include other cases that can result from improper target schema creation, especially around constraints. Makes sure that these get recorded in the exceptions logging table.

Referenced this: https://github.com/cockroachdb/cockroach/blob/786ed22e6c319b8d28399f8ee1f82b88efe7c1ef/pkg/sql/pgwire/pgcode/codes.go#L326

Release Note: None


fetch/testdata/pg/exceptions.ddt
Jeremyyang920, line 1: Can we split the different tests up a bit more so this exceptions.ddt file doesnt become so long and unreadable after a certain point?

ryanluu12345: Sure, let me split the new tests up into constraint-exceptions.ddt

ryanluu12345: @Jeremyyang920 how are the latest changes?

Jeremyyang920 (approved):

fetch/testdata/pg/exceptions.ddt
ryanluu12345: Yeah I agree. We can always split them out to their own test files in the future. I don't have a problem with that. Let's see how this feature evolves with continuation work first.

PR #124, opened 2024-01-26
fetch: added in helpers to exception logs to fetch exception logs by ID or the most recent ones by table

ryanluu12345:
The helpers query the database in order to get the relevant exception logs for a given continuation token or the fetch ID that one wants to resume from. These end up helping the user to figure out how they can continue from where they left off.

Resolves: CC-26910
Release Note: None

### DIFF
--- a/fetch/status/exceptions.go
+++ b/fetch/status/exceptions.go
@@ -49,8 +49,18 @@ type ExceptionLog struct {
 	Time     time.Time
 }
 
+// Added logic to be able to set the time from the struct itself to simplify testing.
+// For some of the tests, we must test exceptions created at multiple times in order
+// to properly validate behavior. Testutils.hook for date doesn't work since all
+// entries would end up having the same.
+// In the actual use case (non-testing) the time will be generated as the current time
+// as the logic below shows.
 func (e *ExceptionLog) CreateEntry(ctx context.Context, conn *pgx.Conn, stage string) error {
 	curTime := time.Now().UTC()
+	if !e.Time.IsZero() {
+		curTime = e.Time
+	}
+
 	query := `INSERT INTO _molt_fetch_exception (fetch_id, table_name, schema_name, message, sql_state, file_name, command, stage, time) VALUES(@fetch_id, @table_name, @schema_name, @message, @sql_state, @file_name, @command, @stage, @time) RETURNING id, stage`
 	args := pgx.NamedArgs{
 		"fetch_id":    e.FetchID,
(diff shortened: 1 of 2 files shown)


ryanluu12345:
Couple of considerations here I'd like your opinion on:
1. Right now, we do each table concurrently so each one of these tables runs a `fetchTable` function. This begs the question: how do we get the proper continuation points for each table?
- One way we can do it (not as preferred since more inefficient, but simpler): have each concurrent call make a call to the database for the exact table + token
- Another way we can do it (preferred but more complex): do the querying before hand and pass in a mapping of table to ExceptionLog (if relevant). Inside of each individual method, have it check the mapping to see if the exception log is there

2. What happens in the case that we fetch ExceptionLogs where only some have file_name? Do we just early exit? Or do we continue with the ones we can? I'd vote for not allowing continuation in this case.

Jeremyyang920:
> Right now, we do each table concurrently so each one of these tables runs a fetchTable function. This begs the question: how do we get the proper continuation points for each table?

Assuming this is when they only supplied the fetch-id and we need all tables since if you supply continuation-token thats a specific table. Can't we query for things before we start the go routines?

ryanluu12345:
> > Right now, we do each table concurrently so each one of these tables runs a fetchTable function. This begs the question: how do we get the proper continuation points for each table?
> 
> Assuming this is when they only supplied the fetch-id and we need all tables since if you supply continuation-token thats a specific table. Can't we query for things before we start the go routines?

Yep, that was what I was thinking too. We can keep these helpers as is. And where we use it, we can go ahead and map the schema + table to the exception log entry and then pass those into the actual fetch call. This works for both cases of single table and all tables.

fetch/status/exceptions.go
```diff
@@ -51,6 +51,10 @@ type ExceptionLog struct {
 
 func (e *ExceptionLog) CreateEntry(ctx context.Context, conn *pgx.Conn, stage string) error {
 	curTime := time.Now().UTC()
+	if !e.Time.IsZero() {
+		curTime = e.Time
+	}
+
```
Jeremyyang920, line 57: What is this change for? Since don't we want a new generated time for every record?

ryanluu12345: It's so if the time is set, we set the current time to that. Otherwise, just default to creating a new time on the entry.

Jeremyyang920: But in what cases is time already set when we call CreateEntry? Like why would we be creating entry using a timestamp in the past.

ryanluu12345: Testing. You can see it in the test files for this PR. It's helpful to be able to set it. In most cases, it will just be created without timestamp set (in the actual app)

Jeremyyang920: Can we use a hook on Time instead of this since that follows more of the pattern for time tests in other places.

ryanluu12345: More complex case here, because in a single test I need to set 3 separate times for entries in order to test that my query only picks up the latest ordered by time. That's why I opted for this method over the `testutils.Hook`

Jeremyyang920: Okay then lets at least comment above this block why this is here since it doesnt make any intuitive sense when reading over without context.

fetch/status/exceptions.go
```diff
@@ -80,6 +84,56 @@ func (e *ExceptionLog) CreateEntry(ctx context.Context, conn *pgx.Conn, stage st
...
+		return nil, err
+	}
+
+	return e, nil
+}
+
+func GetAllExceptionLogsByFetchID(
+	ctx context.Context, conn *pgx.Conn, fetchID string,
+) ([]*ExceptionLog, error) {
+	query := `SELECT DISTINCT ON (schema_name, table_name) id, fetch_id, table_name, schema_name, message, sql_state, file_name, command, stage, time
```
Jeremyyang920, line 110: Is this distinct clause actually used? Since each fetchID is unique to the run, so how can the same schema and table name be populated with the same fetchID

ryanluu12345: Distinct is on the table/schema name. It is used because there could be multiple entries per table logged. An example of this is when import retries and logs this for each retry. We don't want all of those entries, we just want the latest one.

Jeremyyang920 (approved):
LGTM once you add the comment on the time.iszero

ryanluu12345:
@Jeremyyang920 I'll address your comment, but I'll also put the helper that creates the string to exception log mapping in this PR too. I've already started work on it. Might as well loop this in as the `helpers` PR.

ZhouXing19 (approved):

fetch/status/exceptions.go
ryanluu12345: Addressed.

PR #125, opened 2024-01-28
fetch: fix csv pipe deadlock + add row count per file to blob resource

ZhouXing19:
## fetch: Add row count per file to blob resource 
(Authored by @Jeremyyang920)

This commit adds a row count field to ever storage resource. The reason for this is mainly for progress tracking but more specfically for the IMPORT use case where we are not able to know the number of rows imported. We now have this information since each resource has its own file count.

Release note: None

----

## fetch: fix deadlock when failed to upload to blob resource

Previous, if error happens when writing result to the storage (gcp/aws/local),
the error won't be properly propagated back to the main routine in exportTable(),
causing the main routine to hang.

This patch is to fix it so that the main process exits in this case.

Context:

- https://cockroachlabs.slack.com/archives/C04TQ67V1JL/p1705445765473889
- https://github.com/cockroachdb/molt/pull/114#discussion_r1453968635

Release note (bug fix): resolve deadlock when failed to upload csv to the data
blob storage.

### DIFF
--- a/fetch/csv_pipe.go
+++ b/fetch/csv_pipe.go
@@ -24,7 +24,7 @@ type csvPipe struct {
 	currRows  int
 	numRows   int
 	numRowsCh chan int
-	newWriter func(numRowsCh chan int) io.WriteCloser
+	newWriter func(numRowsCh chan int) (io.WriteCloser, error)
 
 	testingKnobs testutils.FetchTestingKnobs
 }
@@ -34,7 +34,7 @@ func newCSVPipe(
 	logger zerolog.Logger,
 	flushSize int,
 	flushRows int,
-	newWriter func(numRowsCh chan int) io.WriteCloser,
+	newWriter func(numRowsCh chan int) (io.WriteCloser, error),
 ) *csvPipe {
 	return &csvPipe{
 		in:        in,
@@ -63,7 +63,9 @@ func (p *csvPipe) Pipe(tn dbtable.Name) error {
 			}
 			return err
 		}
-		p.maybeInitWriter()
+		if err := p.maybeInitWriter(); err != nil {
+			return err
+		}
 		p.currRows++
 		p.numRows++
 		m.Inc()
@@ -114,9 +116,14 @@ func (p *csvPipe) flush() error {
 	return nil
 }
 
-func (p *csvPipe) maybeInitWriter() {
+func (p *csvPipe) maybeInitWriter() error {
 	if p.csvWriter == nil {
-		p.out = p.newWriter(p.numRowsCh)
+		out, err := p.newWriter(p.numRowsCh)
+		if err != nil {
+			return err
+		}
+		p.out = out
 		p.csvWriter = csv.NewWriter(p.out)
 	}
+	return nil
 }
--- a/fetch/csv_pipe.go
+++ b/fetch/csv_pipe.go
@@ -18,12 +18,14 @@ type csvPipe struct {
 	out       io.WriteCloser
 	logger    zerolog.Logger
 
-	flushSize    int
-	flushRows    int
-	currSize     int
-	currRows     int
-	numRows      int
-	newWriter    func() io.WriteCloser
+	flushSize int
+	flushRows int
+	currSize  int
+	currRows  int
+	numRows   int
+	numRowsCh chan int
+	newWriter func(numRowsCh chan int) io.WriteCloser
+
 	testingKnobs testutils.FetchTestingKnobs
 }
 
@@ -32,17 +34,22 @@ func newCSVPipe(
 	logger zerolog.Logger,
 	flushSize int,
 	flushRows int,
-	newWriter func() io.WriteCloser,
+	newWriter func(numRowsCh chan int) io.WriteCloser,
 ) *csvPipe {
 	return &csvPipe{
 		in:        in,
 		logger:    logger,
 		flushSize: flushSize,
 		flushRows: flushRows,
+		numRowsCh: make(chan int, 1),
 		newWriter: newWriter,
 	}
 }
 
+// Pipe is responsible for reading data from the SQL Pipe
+// and creating a CSV file. The CSV data is either flushed
+// when max rows or max file size is reached. The output
+// is written to another pipe which is used as an io.Reader.
 func (p *csvPipe) Pipe(tn dbtable.Name) error {
 	r := csv.NewReader(p.in)
 	r.ReuseRecord = true
@@ -87,8 +94,14 @@ func (p *csvPipe) Pipe(tn dbtable.Name) error {
 	}
 }
 
+// Flush flushes the current csv files when either
+// we reached the end of the file or our file limits.
+// It also sends the current number of rows processed
+// to a channel to be processed by the data storage
+// backend.
 func (p *csvPipe) flush() error {
 	if p.csvWriter != nil {
+		p.numRowsCh <- p.currRows
 		p.csvWriter.Flush()
 		if err := p.out.Close(); err != nil {
 			return err
@@ -103,7 +116,7 @@ func (p *csvPipe) flush() error {
 
 func (p *csvPipe) maybeInitWriter() {
 	if p.csvWriter == nil {
-		p.out = p.newWriter()
+		p.out = p.newWriter(p.numRowsCh)
 		p.csvWriter = csv.NewWriter(p.out)
 	}
 }
--- a/fetch/datablobstorage/datablobstorage.go
+++ b/fetch/datablobstorage/datablobstorage.go
@@ -19,7 +19,7 @@ type Store interface {
 	// CSVs from the data export process. It will create the file and upload
 	// it to the respetive data store and return the resource object which
 	// will be used in the data import phase.
-	CreateFromReader(ctx context.Context, r io.Reader, table dbtable.VerifiedTable, iteration int, fileExt string) (Resource, error)
+	CreateFromReader(ctx context.Context, r io.Reader, table dbtable.VerifiedTable, iteration int, fileExt string, numRows chan int) (Resource, error)
 	// ListFromContinuationPoint is used when restarting Fetch from
 	// a continuation point. It will query the respective data store
 	// and create the slice of resources that will be used by the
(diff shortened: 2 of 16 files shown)


ZhouXing19:
Note that this PR is relatively large, including several refactoring of existing code (mostly to isolate them into helper funcs), so you might want to review it by commits!

fetch/datablobstorage/datablobstorage.go
```diff
@@ -48,3 +57,88 @@ func getKeyAndPrefix(fileName, bucketPath string, table dbtable.VerifiedTable) (
...
+
+type S3Payload struct {
+	S3Bucket   string
+	BucketPath string
+	Region     string
+}
+
+type DatastoreCreationPayload struct {
+	DirectCopyPl *DirectCopyPayload
+	GcpPl        *GcpPayload
```
Jeremyyang920, line 84: same GCP

fetch/datablobstorage/gcp.go
```diff
@@ -48,16 +56,49 @@ func (s *gcpStore) CreateFromReader(
...
-		return nil, err
+
+	rows := <-numRows
+
+	// If any error happens before io.Copy returns, the
+	// error will be propagated to the goroutine in exportTable(),
+	// triggering forwardRead.CloseWithError(), which let p.out.Close() in
+	// csvPipe.flush() returns with the same error. This is because `forwardRead`
+	// and `p.out` are the 2 ends of a pipe. Once the read side is closed with
+	// error, the same error would be propagated to the write side.
```
Jeremyyang920, line 67:
```suggestion
	// If any error happens before io.Copy returns, the
	// error will be propagated to the goroutine in exportTable(),
	// triggering forwardRead.CloseWithError(), which will allow p.out.Close() in
	// csvPipe.flush() to return with the same error. This is because `forwardRead`
	// and `p.out` are the 2 ends of a pipe. Once the read side is closed with
	// error, the same error will be propagated to the write side.
```

fetch/datablobstorage/gcp_testutil.go
```diff
@@ -0,0 +1,59 @@
...
+}
+type GcpBucketMock struct {
+	stiface.BucketHandle
+	mock.Mock
+}
+type GcpObjectITMock struct {
+	stiface.ObjectIterator
+	i    int
+	next []storage.ObjectAttrs
+}
```
Jeremyyang920, line 25: Do we need to export these since its in the same package? If not then we can revert it back to `gcpClientMock`

fetch/datablobstorage/s3.go
```diff
@@ -101,27 +109,54 @@ func NewS3Store(
...
+		opts = append(opts, func(uploader *s3manager.Uploader) {
+			uploader.RequestOptions = []request.Option{
+				func(req *request.Request) {
+					// Will be called in req.ApplyOptions() in uploader.singlePart().
+					// Err will be propagated via aws.Request.Send().
+					req.Error = errors.Newf(AWSUploadFileMockErrMsg)
+				},
+			}
+		})
+	}
```
Jeremyyang920, line 143: This can probably be addressed in a later PR since its not urgent, but I don't think we should do a testingKnob for AWS like this just because with a small refactor we can just borrow https://pkg.go.dev/github.com/aws/aws-sdk-go/service/s3/s3manager/s3manageriface the s3manageriface interface and inject that for the s3 call to UploadWithConext and mock the return similar to in ListFromContinuationPoint where it mocks s3iface for ListObjects. It provides some more flexability in terms of errors we want to return in different tests if needed.

Jeremyyang920 (commented):
submitting these for now, but will do another pass later.

fetch/datablobstorage/s3.go
ZhouXing19: Sounds good, tracking here: https://github.com/cockroachdb/molt/issues/127

fetch/datablobstorage/datablobstorage.go
```diff
@@ -48,3 +57,88 @@ func getKeyAndPrefix(fileName, bucketPath string, table dbtable.VerifiedTable) (
...
+	GCPPl        *GCPPayload
+	S3Pl         *S3Payload
+	LocalPathPl  *LocalPathPayload
+
+	logger zerolog.Logger
+
+	TestFailedWriteToBucket bool
+}
+
+func GenerateDatastore(ctx context.Context, cfg DatastoreCreationPayload) (Store, error) {
```
ryanluu12345, line 93: Alternatively, you can pass in `any` and do a type cast to see if it matches the case, instead of having nested structs. But what you have works fine here. It may be confusing if people accidentally set multiple configs for different modes in the future.

fetch/datablobstorage/gcp.go
```diff
@@ -48,16 +56,50 @@ func (s *gcpStore) CreateFromReader(
...
+
+	// If any error happens before io.Copy returns, the
+	// error will be propagated to the goroutine in exportTable(),
+	// triggering forwardRead.CloseWithError(), which will allow p.out.Close() in
+	// csvPipe.flush() to return with the same error. This is because `forwardRead`
+	// and `p.out` are the 2 ends of a pipe. Once the read side is closed with
+	// error, the same error will be propagated to the write side.
+	// See also: https://go.dev/play/p/H-pHiEffcZE.
+
+	if testingKnobs.FailedWriteToBucket.FailedAfterReadFromPipe {
```
ryanluu12345, line 70: Good use of testing knobs throughout.

fetch/datablobstorage/gcp_test.go
```diff
@@ -40,41 +38,6 @@ func TestGCPResource_ImportURL(t *testing.T) {
 	}
 }
 
-type gcpClientMock struct {
```
ryanluu12345, line 43: Why is this no longer needed?

ryanluu12345: Disregard, i see that you moved to another file.

fetch/csv_pipe.go
```diff
@@ -32,17 +34,22 @@ func newCSVPipe(
...
 	flushRows int,
-	newWriter func() io.WriteCloser,
+	newWriter func(numRowsCh chan int) (io.WriteCloser, error),
 ) *csvPipe {
 	return &csvPipe{
 		in:        in,
 		logger:    logger,
 		flushSize: flushSize,
 		flushRows: flushRows,
+		numRowsCh: make(chan int, 1),
```
ryanluu12345, line 44: What's the reasoning for making a channel with size 1?

fetch/datablobstorage/gcp.go
```diff
@@ -48,16 +56,50 @@ func (s *gcpStore) CreateFromReader(
...
+		// We need a mock writer which simulates the failed upload.
+		mockWc := &GCPStorageWriterMock{wc}
+		if _, err := io.Copy(mockWc, r); err != nil {
+			// This should trigger an error anyways, but we keep this else
+			// condition just in case.
+			return nil, err
+		} else {
+			return nil, errors.AssertionFailedf("didn't see the forced error")
+		}
+	} else {
```
ryanluu12345, line 80: Is this else nesting necessary if the result of the testing block will return anyways? You can probably reduce nesting + complexity by taking the `else block` out.

fetch/datablobstorage/local.go
```diff
@@ -192,7 +207,13 @@ func (l *localResource) Key() (string, error) {
 	return rel, nil
 }
 
+func (l *localResource) Rows() (int, error) {
```
ryanluu12345, line 210: Why do rows return `int, error`? Each of the stores seems to return `rows, nil`

fetch/csv_pipe.go
ZhouXing19: Because for table, the writing to a csv is one batch after another, i.e. there is no batch parallelsation here, and there can only be one numRows being processed at a time. @Jeremyyang920 may correct me if i didn't understand it correctly.

fetch/datablobstorage/local.go
Jeremyyang920: Its just following the convention of everything else above. Key() does the same thing. We can modift the signature though. It can't really error at all since its just an int.

fetch/datablobstorage/datablobstorage.go
ZhouXing19: Good idea, done.

fetch/datablobstorage/gcp.go
ZhouXing19: Good call out, done.

fetch/datablobstorage/local.go
Jeremyyang920: Actually we can change this. I see key has to do the relative path stuff. @ZhouXing19 Could you make tthis quick update to just change the interface signature to just return int? Thanks!

Jeremyyang920 (approved):
LGTM! Thanks for taking this on and fixing up the bug.

fetch/datablobstorage/local.go
ZhouXing19: done removing the error return field.

ZhouXing19:
FYI, this PR doesn't include tests for the original row count add-on for each blob storage, it only includes test for the deadlock issue. The test is tracked in this jira ticket: https://cockroachlabs.atlassian.net/browse/CC-27091

ryanluu12345:
> FYI, this PR doesn't include tests for the original row count add-on for each blob storage, it only includes test for the deadlock issue. The test is tracked in this jira ticket: https://cockroachlabs.atlassian.net/browse/CC-27091

@ZhouXing19 could you please track these test details in this sheet? It's our master sheet of fetch testing needs. https://docs.google.com/spreadsheets/d/1YsLWaIYgQReBWzxid6MKrsRYoTDeEjRWDGFNgAQl0y8/edit?usp=sharing

ZhouXing19:
> @ZhouXing19 could you please track these test details in this sheet? It's our master sheet of fetch testing needs. https://docs.google.com/spreadsheets/d/1YsLWaIYgQReBWzxid6MKrsRYoTDeEjRWDGFNgAQl0y8/edit?usp=sharing

Done.

PR #126, opened 2024-01-30
fetch: integrate continuation with fetch uploaded resources and exception logs

ryanluu12345:
Added in logic for the import/copy only mode that skips the export phase if fetch id or continuation token is passed in. Also integrated with the logic to continue from a certain file point.

Resolves: CC-26910
Release Note: None

- [x] Manual test
- [x] Log out fetch ID and continuation token
- [x] Automated tests

### DIFF
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -292,6 +292,7 @@ func fetchTable(
 				return
 			}
 
+			logger.Info().Msg("cleaning up resources created during fetch run")
 			for _, r := range e.Resources {
 				if r == nil {
 					continue
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -3,6 +3,7 @@ package fetch
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -138,6 +139,13 @@ func Fetch(
 		}
 	}()
 
+	// Wait until all the verification portions are completed first before deferring this.
+	// If verify fails, we don't need to report fetch_id.
+	defer func() {
+		logger.Info().
+			Str("fetch_id", fetchStatus.ID.String()).Msg("continue from this fetch ID")
+	}()
+
 	numTables := len(tables)
 	summaryLogger := moltlogger.GetSummaryLogger(logger)
 	summaryLogger.Info().
@@ -153,6 +161,11 @@ func Fetch(
 	}
 	var stats statsMu
 
+	exceptionLogMapping, err := getExceptionLogMapping(ctx, cfg, targetPgxConn)
+	if err != nil {
+		return err
+	}
+
 	workCh := make(chan tableverify.Result)
 	g, _ := errgroup.WithContext(ctx)
 	for i := 0; i < cfg.Concurrency; i++ {
@@ -162,8 +175,20 @@ func Fetch(
 				if !ok {
 					return nil
 				}
-				if err := fetchTable(ctx, cfg, logger, conns, blobStore, sqlSrc, table, testingKnobs); err != nil {
-					return err
+
+				var relevantExceptionLog *status.ExceptionLog
+				if v, ok := exceptionLogMapping[table.SafeString()]; ok {
+					relevantExceptionLog = v
+				}
+
+				// We want to run the fetch in only two cases:
+				// 1. Export + import mode combined (when fetch ID is not passed in; means new fetch)
+				// 2. When the fetch ID is passed in and exception log is not nil, which means it is a table we want to continue from.
+				// This means we want to skip if we are trying to continue but there is no entry that specifies where to continue from.
+				if (cfg.FetchID != "" && relevantExceptionLog != nil) || (cfg.FetchID == "") {
+					if err := fetchTable(ctx, cfg, logger, conns, blobStore, sqlSrc, table, relevantExceptionLog, testingKnobs); err != nil {
+						return err
+					}
 				}
 
 				stats.Lock()
@@ -187,6 +212,7 @@ func Fetch(
 
 	ovrDuration := utils.MaybeFormatDurationForTest(cfg.TestOnly, timer.ObserveDuration())
 	summaryLogger.Info().
+		Str("fetch_id", fetchStatus.ID.String()).
 		Int("num_tables", stats.numImportedTables).
 		Strs("tables", stats.importedTables).
 		Str("cdc_cursor", utils.MaybeFormatCDCCursor(cfg.TestOnly, sqlSrc.CDCCursor())).
@@ -196,6 +222,9 @@ func Fetch(
 	return nil
 }
 
+// Note that if `ExceptionLog` is not nil, then that means
+// there is an exception log and import/copy only mode
+// was specified.
 func fetchTable(
 	ctx context.Context,
 	cfg Config,
@@ -204,8 +233,9 @@ func fetchTable(
 	blobStore datablobstorage.Store,
 	sqlSrc dataexport.Source,
 	table tableverify.Result,
+	exceptionLog *status.ExceptionLog,
 	testingKnobs testutils.FetchTestingKnobs,
-) error {
+) (retErr error) {
 	tableStartTime := time.Now()
 	// Initialize metrics for this table so we can calculate a rate later.
 	fetchmetrics.ExportedRows.WithLabelValues(table.SafeString())
@@ -223,13 +253,45 @@ func fetchTable(
 
 	logger.Info().Msgf("data extraction phase starting")
 
-	e, err := exportTable(ctx, cfg, logger, sqlSrc, blobStore, table.VerifiedTable, testingKnobs)
-	if err != nil {
-		return err
+	var e exportResult
+	// In the case that exception log is nil or fetch id is empty,
+	// this means that we want to export the table because it means
+	// we want export + copy mode.
+	if exceptionLog == nil || cfg.FetchID == "" {
+		er, err := exportTable(ctx, cfg, logger, sqlSrc, blobStore, table.VerifiedTable, testingKnobs)
+		if err != nil {
+			return err
+		}
+		e = er
+	} else {
+		if exceptionLog.FileName == "" {
+			logger.Warn().Msgf("skipping table %s because no file name is present in the exception log", table.SafeString())
+			return errors.Newf("table %s not imported because no file name is present in the exception log", table.SafeString())
+		}
+		logger.Warn().Msgf("skipping export for table %s due to running in import-copy only mode", table.SafeString())
+
+		// TODO: future PR needs to add number of rows estimation. and populate exportResult.NumRows
+		// TODO: need to figure out start and end time too.
+		rsc, err := blobStore.ListFromContinuationPoint(ctx, table.VerifiedTable, exceptionLog.FileName)
+		if err != nil {
+			return err
+		}
+		e.Resources = rsc
+
+		if len(e.Resources) == 0 {
+			return errors.Newf("exported resources for table %s is empty, please make sure you did not accidentally delete from the intermediate store", table.SafeString())
+		}
 	}
 
+	// We actually need to skip the cleanup for something that has an error
+	// On a continuation run we can cleanup so long as it's successful.
 	if cfg.Cleanup {
 		defer func() {
+			if retErr != nil {
+				logger.Info().Msg("skipping cleanup because an error occurred and files may need to be kept for continuation")
+				return
+			}
+
 			for _, r := range e.Resources {
 				if r == nil {
 					continue
@@ -242,6 +304,7 @@ func fetchTable(
 		}()
 	}
 
+	// TODO: consider if we want to skip this portion since we don't export anything....
 	exportDuration := utils.MaybeFormatDurationForTest(cfg.TestOnly, e.EndTime.Sub(e.StartTime))
 	summaryLogger := moltlogger.GetSummaryLogger(logger)
 	summaryLogger.Info().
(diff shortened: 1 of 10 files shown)


fetch/fetch.go
```diff
@@ -162,7 +168,9 @@ func Fetch(
 				if !ok {
 					return nil
 				}
-				if err := fetchTable(ctx, cfg, logger, conns, blobStore, sqlSrc, table, testingKnobs); err != nil {
+
+				relevantExceptionLog := exceptionLogMapping[table.SafeString()]
```
Jeremyyang920, line 172: We should check if this key exists right? Otherwise this can panic.

fetch/fetch.go
```diff
@@ -223,25 +237,53 @@ func fetchTable(
...
+		extension := "csv"
+		if cfg.Compression == compression.GZIP {
+			extension = ".tar.gz"
+		}
+
+		fileName := exceptionLog.FileName
+		if exceptionLog.FileName == "" {
+
+			fileName = fmt.Sprintf("%s.%s", defaultFileName, extension)
+		}
```
Jeremyyang920, line 257: So we are starting with part 1 if the file is unknown?

fetch/fetch.go
```diff
@@ -365,3 +407,34 @@ func initStatusEntry(
 
 	return fetchStatus, nil
 }
+
+// TODO: handle the case where the file override happens.
```
Jeremyyang920, line 411: File override should probably be handled inside the fetch table stuff. The fileName parameter is is the thing we use to know what to start at.

fetch/fetch.go
ryanluu12345: I'm adding logic locally right now to make it skip in this case. But we do want to consider: is the proper thing to do here to error out since we don't want people to continue in this case and think everything is ok.

fetch/fetch.go
ryanluu12345: Yep good point, I'll do this diff:
```
var relevantExceptionLog *status.ExceptionLog
				if v, ok := exceptionLogMapping[table.SafeString()]; ok {
					relevantExceptionLog = v
				}
```

fetch/fetch.go
```diff
@@ -223,25 +240,47 @@ func fetchTable(
...
-		defer func() {
-			for _, r := range e.Resources {
-				if r == nil {
-					continue
-				}
+		if !IsImportCopyOnlyMode(cfg) {
+			defer func() {
+				for _, r := range e.Resources {
+					if r == nil {
+						continue
```
ZhouXing19, line 270: Just out of curiosity, how could this case happen? Or it's simply a sanity check?

ZhouXing19 (approved):
LGTM but would like to have Jeremy to give a final stamp

fetch/fetch.go
Jeremyyang920: Yup. Im not sure the right thing here is too. I would imagine its err on the side of caution and error out and let them know why. Its sort of the same principle with why we wanted to stop at the error so users dont think everything is ok.

fetch/datablobstorage/datablobstorage_test.go
```diff
@@ -0,0 +1,43 @@
+package datablobstorage
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_matchesFileConvention(t *testing.T) {
```
Jeremyyang920, line 9: Nit: TestMatchesFileConvention

fetch/fetch.go
```diff
@@ -223,25 +240,47 @@ func fetchTable(
...
-	e, err := exportTable(ctx, cfg, logger, sqlSrc, blobStore, table.VerifiedTable, testingKnobs)
-	if err != nil {
-		return err
+	var e exportResult
+	if exceptionLog == nil {
+		er, err := exportTable(ctx, cfg, logger, sqlSrc, blobStore, table.VerifiedTable, testingKnobs)
+		if err != nil {
+			return err
+		}
+		e = er
```
Jeremyyang920, line 249: I might be missing something, but if you do it this way coupled with the changes above on line 171-177, wouldn't this mean we exportTable for any table thats nil on exception restart even if it might have been successful already? 

Since we are still iterating over workCh which has all tables, the successful table wont have an entry in the exception log mapping, and you would pass nil into here.

fetch/fetch.go
ryanluu12345: Yeah, I agree. I'll make this return an error. This will make the error group exit early, which I think is the intended behavior here. Which means right now we can't continue `import into` right now.

fetch/fetch.go
ryanluu12345: I think a simple sanity check just in case we yield nil resources. I think it's worth having this before we try to run `MarkForCleanup`. This is a safer operation.

fetch/datablobstorage/datablobstorage_test.go
ryanluu12345: Done!

fetch/fetch.go
ryanluu12345: Discussed this offline with @Jeremyyang920 . Plan here is two fold:
1. Do a run pre-check validation so that if an invalid continuation file name is passed in, we fail early.
2. In the helper here, update the exceptionLog file with the passed in cfg.ContinuationFileName if it is set.

ryanluu12345: This is what the error message looks like:
```
continuation file name "asdk" doesn't match the file convention "part_\d+(\.csv|\.tar\.gz)"
exit status 1
zsh: command not found: lf
(⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % 
```

Jeremyyang920: I think this will be something we have to document for this feature for sure in the readme or official docs of file naming

ryanluu12345: Yep makes sense to me! The error message indicates this case already

fetch/fetch.go
ryanluu12345: Ok that makes sense to me. Then perhaps what we want to do here is skip the fetch table if the exception log retrieved is nil?

ryanluu12345: The resulting diff would be like this:
```
				var relevantExceptionLog *status.ExceptionLog
				if v, ok := exceptionLogMapping[table.SafeString()]; ok {
					relevantExceptionLog = v
				}

				if relevantExceptionLog != nil {
					if err := fetchTable(ctx, cfg, logger, conns, blobStore, sqlSrc, table, relevantExceptionLog, testingKnobs); err != nil {
						return err
					}
				}
```
Reasoning is that if we have no exception log for that table, we shouldn't process that table at all. We should assume that it succeeded in the previous run.

ryanluu12345: Correction, we need to do this actually. Because we need to run fetch table for both cases.
```
if (cfg.FetchID != "" && relevantExceptionLog != nil) || (cfg.FetchID == "") {
					if err := fetchTable(ctx, cfg, logger, conns, blobStore, sqlSrc, table, relevantExceptionLog, testingKnobs); err != nil {
						return err
					}
				}
```

ryanluu12345: Look at this commit: https://github.com/cockroachdb/molt/pull/126/commits/d3b68d650d2f2edeb32c53ea5fb97f8cb396b8af

ryanluu12345: Discussed offline with @Jeremyyang920 looks good.

ryanluu12345:
Manually tested the following series of steps:
1. Ran a fetch with default settings using `--live-copy` mode -> successfully fetched
2. Ran a fetch with the same settings again -> which predictably failed -- duplicate key violation
3. Ran a fetch with the fetch id continue passed in -> skipped import/copy but ended up failing because duplicate key violation
4. Truncated data in the target employees table
5. Ran a fetch with the fetch id continue passed in -> skipped import/copy and succeeded
```
{"level":"info","time":"2024-01-31T14:25:13-08:00","message":"default compression to none"}
{"level":"info","time":"2024-01-31T14:25:13-08:00","message":"checking database details"}
{"level":"info","source_table":"public.employees","target_table":"public.employees","time":"2024-01-31T14:25:13-08:00","message":"found matching table"}
{"level":"info","time":"2024-01-31T14:25:13-08:00","message":"verifying common tables"}
{"level":"info","time":"2024-01-31T14:25:13-08:00","message":"establishing snapshot"}
{"level":"info","type":"summary","num_tables":1,"cdc_cursor":"0/3F3E0B0","time":"2024-01-31T14:25:13-08:00","message":"starting fetch"}
{"level":"info","time":"2024-01-31T14:25:13-08:00","message":"data extraction phase starting"}
{"level":"warn","time":"2024-01-31T14:25:13-08:00","message":"skipping export for table public.employees because import-copy only mode"}
{"level":"info","type":"summary","num_rows":0,"export_duration_ms":0,"export_duration":"000h 00m 00s","time":"2024-01-31T14:25:13-08:00","message":"data extraction from source complete"}
{"level":"info","time":"2024-01-31T14:25:13-08:00","message":"starting data import on target"}
{"level":"info","type":"data","num_rows":200000,"table":"public.employees","file":"public.employees/part_00000001.csv","time":"2024-01-31T14:25:21-08:00","message":"row copy status"}
{"level":"info","type":"data","duration":8285.410208,"time":"2024-01-31T14:25:21-08:00","message":"table COPY complete"}
{"level":"info","type":"summary","net_duration_ms":8704.371125,"net_duration":"000h 00m 08s","import_duration_ms":8285.410208,"import_duration":"000h 00m 08s","export_duration_ms":0,"export_duration":"000h 00m 00s","num_rows":0,"cdc_cursor":"0/3F3E0B0","time":"2024-01-31T14:25:21-08:00","message":"data import on target for table complete"}
{"level":"info","type":"summary","fetch_id":"cb5f204e-3bb7-4b6e-a62f-ae03dbf0cf62","num_tables":1,"tables":["public.employees"],"cdc_cursor":"0/3F3E0B0","net_duration_ms":8747.065166,"net_duration":"000h 00m 08s","time":"2024-01-31T14:25:21-08:00","message":"fetch complete"}
{"level":"info","fetch_id":"cb5f204e-3bb7-4b6e-a62f-ae03dbf0cf62","time":"2024-01-31T14:25:21-08:00","message":"continue from this fetch ID"}
```

fetch/fetch.go
```diff
@@ -67,6 +68,10 @@ func Fetch(
 		RunID:     fetchStatus.ID,
 		StartedAt: fetchStatus.StartedAt,
 	})
+	defer func() {
+		logger.Info().
+			Str("fetch_id", fetchStatus.ID.String()).Msg("continue from this fetch ID")
+	}()
```
Jeremyyang920, line 74: Is this too early in the call stack to register the defer? we have calls to like db.Verify and such?

fetch/fetch.go
```diff
@@ -223,13 +250,45 @@ func fetchTable(
...
+			return errors.Newf("exported resources for table %s is empty, please make sure you did not accidentally delete from the intermediate store", table.SafeString())
+		}
 	}
 
+	// We actually need to skip the cleanup for something that has an error
+	// On a continuation run we can cleanup so long as it's successful.
 	if cfg.Cleanup {
 		defer func() {
+			if retErr != nil {
+				logger.Info().Msg("skipping cleanup because error is hit and files may need to be kept for continuation")
```
Jeremyyang920, line 288:
```suggestion
				logger.Info().Msg("skipping cleanup because an error occured and files may need to be kept for continuation")
```

fetch/fetch.go
```diff
@@ -223,13 +250,45 @@ func fetchTable(
...
+		if err != nil {
+			return err
+		}
+		e = er
+	} else {
+		if exceptionLog.FileName == "" {
+			logger.Warn().Msgf("skipping table %s because no file name is present in the exception log", table.SafeString())
+			return errors.Newf("table %s not imported because no file name is present in the exception log", table.SafeString())
+		}
+		logger.Warn().Msgf("skipping export for table %s because import-copy only mode", table.SafeString())
```
Jeremyyang920, line 268:
```suggestion
		logger.Warn().Msgf("skipping export for table %s due to running in import-copy only mode", table.SafeString())
```

utils/formatting.go
```diff
@@ -2,10 +2,17 @@ package utils
...
+	"regexp"
 
 	"github.com/cockroachdb/cockroachdb-parser/pkg/sql/sem/tree"
 )
 
 func SchemaTableString(schema, table tree.Name) string {
 	return fmt.Sprintf("%s.%s", string(schema), string(table))
 }
+
+var FileConventionRegex = regexp.MustCompile(`part_[\d+]{8}(\.csv|\.tar\.gz)`)
```
Jeremyyang920, line 14: This is the same regex as var fileNameRegEx = regexp.MustCompile(`part_[\d+]{8}(\.csv|\.tar\.gz)`)
We should maybe consider not duplicating this.

Jeremyyang920 (approved):
Logic lgtm, just the. minor nit fixes.

fetch/fetch.go
ryanluu12345: I put this after `verify`

fetch/fetch.go
ryanluu12345: Taking suggestion!

fetch/fetch.go
ryanluu12345: Done!

utils/formatting.go
ryanluu12345: Good catch. I ended up keeping the one inside of `utils`. And I reference that inside `exceptions.go`.

fetch/fetch.go
```diff
@@ -138,6 +139,13 @@ func Fetch(
 		}
 	}()
 
+	// Wait until all the verification portions are completed first before deferring this.
+	// If verify fails, we don't need to report fetch_id.
+	defer func() {
+		logger.Info().
```
ryanluu12345, line 145: @Jeremyyang920 thoughts on if we should only log this out if it fails?

Jeremyyang920: It makes sense to only log the fetch_id when something happened past verify. Since if verify failed thers nothing really useful in that case.

PR #128, opened 2024-02-01
cmd/fetch: Implement schema creation entrypoint in cli

ZhouXing19:
This serves as an entrypoint for schema creation, and replaces the existing --truncate flag.
Also include some code movement from the verify sub pkg to a more generalized utils pkg.

Release note (cli): introduce the --table-handling flag and deprecate the --truncate
flag for `molt fetch`.


cmd/fetch/fetch.go
```diff
@@ -282,9 +283,39 @@ func Command() *cobra.Command {
...
+	noneTableHandlingKey                    = "none"
+	dropOnTargetAndRecreateTableHandlingKey = "drop-on-target-and-recreate"
+	truncateIfExistsTableHandlingKey        = "truncate-if-exist"
+)
+
+var TableHandlingOptionStringRepresentations = map[TableHandlingOption][]string{
+	None:                    {noneTableHandlingKey},
+	DropOnTargetAndRecreate: {dropOnTargetAndRecreateTableHandlingKey},
+	TruncateIfExists:        {truncateIfExistsTableHandlingKey},
+}
```
Jeremyyang920, line 321: nit: Can we move these to the top of the file so the declarations are seen first when reading so it makes more sense?

Jeremyyang920: Also do you think we will need to generate the string function with the stringer tool like we do in the LMS?

cmd/fetch/fetch.go
```diff
@@ -282,9 +283,39 @@ func Command() *cobra.Command {
...
+const (
+	None TableHandlingOption = iota
+	DropOnTargetAndRecreate
+	TruncateIfExists
+)
+
+const (
+	noneTableHandlingKey                    = "none"
+	dropOnTargetAndRecreateTableHandlingKey = "drop-on-target-and-recreate"
+	truncateIfExistsTableHandlingKey        = "truncate-if-exist"
```
Jeremyyang920, line 314:
```suggestion
	truncateIfExistsTableHandlingKey        = "truncate-if-exists"
```
Just to match the const.

README.md
```diff
@@ -181,7 +181,7 @@ molt fetch \
   --target 'postgres://root@localhost:26257/defaultdb?sslmode=disable' \
   --table-filter 'good_table' \
   --s3-bucket 'otan-test-bucket' \
-  --truncate \ # automatically truncate destination tables before importing
+  --table-handling 'truncate-if-exist' \ # automatically truncate destination tables before importing
```
Jeremyyang920, line 184:
```suggestion
  --table-handling 'truncate-if-exists' \ # automatically truncate destination tables before importing
```

README.md
```diff
@@ -156,7 +156,7 @@ It currently supports the following:
 By default, data is imported using `IMPORT INTO`. You can use `--live` if you
 need target data to be queriable during loading, which uses `COPY FROM` instead.
 
-Data can be truncated automatically if run with `--truncate`.
+Data can be truncated automatically if run with `--table-handling 'truncate-if-exist'`.
```
Jeremyyang920, line 159:
```suggestion
Data can be truncated automatically if run with `--table-handling 'truncate-if-exists'`.
```

fetch/fetch.go
```diff
@@ -231,6 +231,23 @@ func fetchTable(
 		return nil
 	}
 
+	// Truncate table on the target side, if applicable.
```
ryanluu12345, line 234: This logic could be refactored into a helper method called `truncateTable`. That way if we need to move it again later it's just a method call.

README.md
```diff
@@ -181,7 +181,7 @@ molt fetch \
   --target 'postgres://root@localhost:26257/defaultdb?sslmode=disable' \
   --table-filter 'good_table' \
   --s3-bucket 'otan-test-bucket' \
-  --truncate \ # automatically truncate destination tables before importing
```
ryanluu12345, line 184: Is there a reason we are deprecating the `--truncate` flag and not just adding a `table-handling` flag in addition to `truncate`?

cmd/fetch/fetch.go
```diff
@@ -16,12 +17,33 @@ import (
...
+
 func Command() *cobra.Command {
 	const (
 		fetchID              = "fetch-id"
 		continuationToken    = "continuation-token"
 		continuationFileName = "continuation-file-name"
 	)
+
+	const (
+		None TableHandlingOption = iota
```
ryanluu12345, line 30: What's the behavior if this is not passed in? No table-handling action?

ryanluu12345, line 30: Worthwhile to add comments for what each of these modes will do. Also add to readme and flag desc.

fetch/testdata/pg/direct.ddt
```diff
@@ -46,3 +46,31 @@ id	t
 6	Лукас
 7	ルカス
 tag: SELECT 7
+
+# To test the default "truncate" option is effective.
+fetch direct
```
ryanluu12345, line 51: Wait how does this test that the `truncate` option is effective. Is `truncate` the default?

utils/obj_filter.go
```diff
@@ -1,4 +1,4 @@
-package dbverify
+package utils
```
ryanluu12345, line 1: Generally what's the reasoning for migrating to a `utils` package here? And I'm wondering if there is a more specific name for this package that still captures what you want to do. I believe that `utils` sounds a bit too general. These all sound like `verifcationutils`

README.md
ZhouXing19: It's because I think truncate is a subset of table handling logic. It also doesn't really make sense if the user specify both `truncate` and `create schema`. So I just catogorize them under the same flag so that they are in parallel.

cmd/fetch/fetch.go
ZhouXing19: yeah, just start right on searching with table name on the target db and do the import/copy job.

fetch/testdata/pg/direct.ddt
ZhouXing19: yes, see `truncate := true` in `fetch_test.go`

utils/obj_filter.go
ZhouXing19: i think utils are basically a garage for all stuff shared by both fetch and verify, so it is meant to be general.

fetch/fetch.go
ZhouXing19: Done.

cmd/fetch/fetch.go
ZhouXing19: Done.

fetch/testdata/pg/direct.ddt
ryanluu12345: ok sounds good. I'm wondering if we should make `truncate` not the default behavior. So that people can explicitly set it or disable it. Although I think we have `notruncate`

utils/obj_filter.go
ryanluu12345: Ok sounds good, so generally shared. That should be ok then.

cmd/fetch/fetch.go
ryanluu12345: Ok sounds good. What happens if someone wants both table handling for create schema AND truncate? Is that possible?

ZhouXing19: if they chose `drop and create schema` then the new table is already empty, so no need for truncate. if they chose `truncate` then it means that they want the existing schema on the target side then no point to create a new one. Having both of them are contradicting to me.

fetch/testdata/pg/direct.ddt
ZhouXing19: I think for test in most cases we want to start with an empty table so truncate as default makes sense to me.

Jeremyyang920: I agree for tests we should truncate always to ensure it in a clean state unless you have a reason to not truncate for some other test.

cmd/fetch/fetch.go
ryanluu12345: Understood

README.md
ZhouXing19: Yes I think so.

ZhouXing19:
Pump for another round of review :)

fetch/testdata/pg/constraint-exceptions.ddt
```diff
@@ -64,7 +64,7 @@ INSERT INTO dependent VALUES (1, 0, 'aaa')
 
 fetch live expect-error
 ----
-ERROR: "dependency" is referenced by foreign key from table "dependent" (SQLSTATE XXUUU)
+failed executing the TRUNCATE TABLE statement: ERROR: "dependency" is referenced by foreign key from table "dependent" (SQLSTATE XXUUU)
```
Jeremyyang920, line 67: Why did we truncate here on this directive?

Jeremyyang920 (approved):
Overall LGTM! Just the copy comments and the 1 comment on that DDT.

fetch/testdata/pg/constraint-exceptions.ddt
ZhouXing19: `truncate` is the default for a `fetch` command in a datadriven test

ZhouXing19:
Per the weekly meeting, added these lines in `cmd/fetch.go`

```go
			//TODO(janexing): remove once finished the implementation.
			if tableHandlingMode == DropOnTargetAndRecreate {
				return errors.New("drop-on-target-and-recreate feature not implemented yet")
			}
```

to gate the usage of this flag now.

PR #129, opened 2024-02-01
[CC-27092] fetch: Add bisect logic for import

Jeremyyang920:
This commit adds the logic to allow bisecting a batch of files from import that failed to try and narrow down the problem file to record in the exceptions table.

### DIFF
--- a/fetch/import_table.go
+++ b/fetch/import_table.go
@@ -21,6 +21,7 @@ import (
 	"github.com/cockroachdb/molt/retry"
 	"github.com/cockroachdb/molt/testutils"
 	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/rs/zerolog"
 )
 
@@ -35,9 +36,14 @@ type importProgress struct {
 	FractionCompleted float64 `db:"fraction_completed"`
 }
 
+type PGIface interface {
+	Exec(ctx context.Context, sql string, arguments ...any) (pgconn.CommandTag, error)
+}
+
 const (
 	pattern     = `%`
 	replacement = `\%`
+	batchSize   = 10
 )
 
 var re = regexp.MustCompile(pattern)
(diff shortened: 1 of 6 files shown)


fetch/import_table_test.go
```diff
@@ -177,3 +180,88 @@ ORDER BY created DESC`,
...
+		{
+			name: "file 8 error",
+			dbExpectations: func(mock pgxmock.PgxConnIface) {
+				mock.ExpectExec("IMPORT INTO").Times(1).WillReturnError(errors.New("mock error"))
+				mock.ExpectExec("IMPORT INTO").WillReturnResult(pgconn.NewCommandTag("tag"))
+				mock.ExpectExec("IMPORT INTO").Times(1).WillReturnError(errors.New("mock error"))
+				mock.ExpectExec("IMPORT INTO").WillReturnResult(pgconn.NewCommandTag("tag"))
+				mock.ExpectExec("IMPORT INTO").Times(2).WillReturnError(errors.New("mock error"))
+			},
+			expected: "part_00000008.csv",
```
Jeremyyang920, line 237: If you want to track these cases yourself, take a look at the comment importWithBisect where I explain how the files are broken up so you can see why the mock expectations are the way they are.

fetch/import_table.go
```diff
@@ -35,9 +36,14 @@ type importProgress struct {
...
 }
 
+type PGIface interface {
+	Exec(ctx context.Context, sql string, arguments ...any) (pgconn.CommandTag, error)
+}
+
 const (
 	pattern     = `%`
 	replacement = `\%`
+	batchSize   = 10
```
ryanluu12345, line 46: Why batch size of 10 as the default?

fetch/import_table.go
```diff
@@ -145,52 +151,71 @@ func importTable(
...
-		kvOptions := tree.KVOptions{}
-		if cfg.Compression == compression.GZIP {
-			kvOptions = append(kvOptions, tree.KVOption{
-				Key:   "decompress",
-				Value: tree.NewStrVal("gzip"),
-			})
+	for i := 0; i < len(locs); i += batchSize {
+		end := i + batchSize
+		// necessary to prevent going over len
+		if end > len(locs) {
```
ryanluu12345, line 166: Could there be a possibility that we end up missing some of the `locs`?

fetch/import_table.go
```diff
@@ -145,52 +151,71 @@ func importTable(
...
+func importWithBisect(
+	ctx context.Context,
+	kvOptions tree.KVOptions,
+	table dbtable.VerifiedTable,
+	logger zerolog.Logger,
+	conn PGIface,
+	locs []string,
+) (string, error) {
+	stack := [][]string{locs}
+	for len(stack) > 0 {
```
ryanluu12345, line 199: Is there any case where this could incorrectly get into an infinite loop state?

ryanluu12345 (approved):
Just a couple clarifying questions. Overall, really clean implementation and clear tests.

fetch/import_table.go
Jeremyyang920: Thats just what we decided on as a decent number and its branching factor is fairly low as well. We can change it if we want to but its a good starting place and can evaluate as we do large scale tests.

fetch/import_table.go
Jeremyyang920: How could we end up missing some? 
If my len(locs) was 34 and batchSize is 10. The last iteration of this loop would mean end=30+10 which is 40. 40 is >len(locs) so we set it to the len.

fetch/import_table.go
Jeremyyang920: There shouldn't be. You only push to the stack on error and with the base cases set to how they are, there should be no extraneous pushes to the stack.

Jeremyyang920: Also this algorithm is a variation of the iterative version of a preorder traversal of a tree just we do some pruning of nodes.

fetch/import_table.go
ryanluu12345: Yep sounds good to me.

fetch/import_table.go
ryanluu12345: Ok got it. Re-read it. It basically clips it at the length of locs, which will index up to last item.

ryanluu12345 (approved):
Everything looks good to me. Btw, could you please outline all test cases you think you need for this in this sheet: https://docs.google.com/spreadsheets/d/1YsLWaIYgQReBWzxid6MKrsRYoTDeEjRWDGFNgAQl0y8/edit?usp=sharing?

fetch/import_table.go
```diff
@@ -145,52 +151,71 @@ func importTable(
...
-		logger.Err(err).Msgf("error importing data, retrying")
-	}); err != nil {
-		return ret, err
 	}
 	ret.EndTime = time.Now()
 	return ret, nil
 }
+
+// importWithBisect handles the logic of trying to find the
+// broken file in a batch of files that were sent to import.
```
ZhouXing19, line 182: nit: maybe update the comment to show that in the case where there are multiple corrupted files in the batch, it actually returns the first broken file in the order of given locs.

Jeremyyang920: Im not sure how to fully show that besides the breakdown of the files shown below. The ordering in the example is the ordering in which the chunks happen. So if there were multiple files with errors and file 1 happens to be one of them, it would guarantee be the first to be returned due to the ordering nature of the traversal.

ZhouXing19: Oh i meant this is just a developer-intended comment. IIUC `stack = append(stack, curr[mid:], curr[:mid])` has guaranteed that it always (and only) returns the first corrupted file in locs. 
E.g. say `locs = [`5.csv`, `4.csv`, `13.csv`, `2.csv`]`, if `4.csv` and `2.csv` are both bad apples, this func only returns `4.csv`, as it's the wrong file with the smallest index in the list.

fetch/import_table.go
```diff
@@ -145,52 +151,71 @@ func importTable(
...
+	logger zerolog.Logger,
+	conn PGIface,
+	locs []string,
+) (string, error) {
+	stack := [][]string{locs}
+	for len(stack) > 0 {
+		curr := stack[len(stack)-1]
+		stack = stack[:len(stack)-1]
+		importQuery := dataquery.ImportInto(table, curr, kvOptions)
+		logger.Debug().Msgf("running import query: %q", importQuery)
```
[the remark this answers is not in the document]
Jeremyyang920: Ah right I saw this too. I think in CRDB we redact when running Copy and import to obfuscate the access key. May need to see how they do that.

fetch/import_table.go
Jeremyyang920: Oh sure. Ill add that.

fetch/import_table.go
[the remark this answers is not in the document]
Jeremyyang920: I'll create an issue for this as to not block this but good call out on this. Wondering if we can utilize the redaction code inside the CRDB main repo somehow.

fetch/import_table.go
```diff
@@ -145,52 +151,77 @@ func importTable(
...
+// [1,2,3,4,5][6,7,8,9,10]
+// [1,2][3,4,5][6,7][8,9,10]
+// [1][2][3][4,5][6][7][8][9,10]
+// [4][5][9][10]
+// Since the pushes to the stack are in FIFO order, if there are multiple
+// files with errors, there is a guarantee that the lowest file part is
+// returned first to preserve import order. So, in the example above, if file
+// 3 and 5 had errors, 3 would be returned first since the row showing the
+// breakdown [1][2][3][4,5] shows that 3 is being processed before 5 in the
+// stack.
```
Jeremyyang920, line 195: @ZhouXing19 Let me know if this comment covers your suggestion.

ZhouXing19: LGTM!

fetch/import_table.go
```diff
@@ -145,52 +151,77 @@ func importTable(
 		locs = append(locs, u)
 	}
 	conn := baseConn.(*dbconn.PGConn)
-	r, err := retry.NewRetry(retry.Settings{
```
ryanluu12345, line 148: Last q here: are we planning on having retries here? Or retries at all? I believe this logic was probably here in order to handle cases where maybe there's a transient import error.

PR #131, opened 2024-02-06
fetch: Update csv_pipe test to include row count

Jeremyyang920:
This commit updates the unit test in csv_pipe to check the rows per batch that it gets back matches the result sent over the numRows channel

Release Note: none

### DIFF
--- a/fetch/csv_pipe_test.go
+++ b/fetch/csv_pipe_test.go
@@ -8,16 +8,18 @@ import (
 
 	"github.com/cockroachdb/molt/dbtable"
 	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestCSVPipe(t *testing.T) {
 	for _, tc := range []struct {
-		desc      string
-		toWrite   string
-		files     []string
-		flushSize int
-		flushRows int
+		desc         string
+		toWrite      string
+		files        []string
+		flushSize    int
+		flushRows    int
+		rowsPerBatch []int
 	}{
 		{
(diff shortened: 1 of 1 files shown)


fetch/csv_pipe_test.go
```diff
@@ -137,18 +148,30 @@ multiline part"
 				tc.flushRows,
 				func(numRows chan int) (io.WriteCloser, error) {
 					go func() {
-						<-numRows
+						rowsCh <- <-numRows
```
ZhouXing19, line 151: fancy!

fetch/csv_pipe_test.go
```diff
@@ -137,18 +148,30 @@ multiline part"
...
 					}()
 					bufs = append(bufs, testStringBuf{})
 					return &bufs[len(bufs)-1], nil
 				},
 			)
 			require.NoError(t, pipe.Pipe(dbtable.Name{Schema: "test", Table: "test"}))
+			it := 0
+			go func() {
+				for rows := range rowsCh {
+					assert.Equal(t, rows, tc.rowsPerBatch[it])
```
ZhouXing19, line 161: nit: might want to reuse the `require` pkg with `require.Equal(t, ...)`.

ZhouXing19 (approved):

fetch/csv_pipe_test.go
Jeremyyang920: So you actually shouldn't use require inside a go routine as I found out. It doesn't behave nicely and there was a thing I read that said you should use assert inside a go routine and require outside in the main test. 

I tested with require here, and it actually just made the test hang.

PR #137, opened 2024-02-07
fetch: added in upload metadata for S3

ryanluu12345:
Added in concurrency for listing from continuation point so we don't block on our request to get specific metadata details per each file. Also added in the calls to external APIs and the upload metadata line.

Resolves: CC-27046
Release Note: None

Test Results:
```
{"level":"warn","time":"2024-02-07T17:18:07-05:00","message":"skipping export for table public.employees due to running in import-copy only mode"}
map[Numrows:0x14000c352f0]
map[Numrows:0x140008f0110]
map[Numrows:0x140008f0270]
map[Numrows:0x140008f0320]
&{0x140006fd080 0x1400092eff0 public.employees/part_00000001.tar.gz 10} &{0x140006fd080 0x1400092eff0 public.employees/part_00000002.tar.gz 10} &{0x140006fd080 0x1400092eff0 public.employees/part_00000003.tar.gz 10} &{0x140006fd080 0x1400092eff0 public.employees/part_00000004.tar.gz 10}
...
{"level":"info","type":"summary","net_duration_ms":12759.923,"net_duration":"000h 00m 12s","import_duration_ms":12220.382875,"import_duration":"000h 00m 12s","export_duration_ms":0,"export_duration":"000h 00m 00s","num_rows":0,"cdc_cursor":"0/3F3E068","time":"2024-02-07T17:18:20-05:00","message":"data import on target for table complete"}
{"level":"info","type":"summary","fetch_id":"41ca41c4-43e3-4b05-992a-f791e65d5575","num_tables":1,"tables":["public.employees"],"cdc_cursor":"0/3F3E068","net_duration_ms":12807.8045,"net_duration":"000h 00m 12s","time":"2024-02-07T17:18:20-05:00","message":"fetch complete"}
...

root@:26257/defaultdb> SELECT COUNT(*) FROM employees;
  count
----------
  200000
```

TODO:
- [x] Fix panic from this test:

```
func TestListFromContinuationPointAWS(t *testing.T) {
```

### DIFF
--- a/fetch/datablobstorage/datablobstorage.go
+++ b/fetch/datablobstorage/datablobstorage.go
@@ -16,8 +16,6 @@ import (
 	"golang.org/x/oauth2/google"
 )
 
-const numRowsKey = "numRows"
-
 type Store interface {
 	// CreateFromReader is responsible for the creation of the individual
 	// CSVs from the data export process. It will create the file and upload
--- a/fetch/datablobstorage/s3.go
+++ b/fetch/datablobstorage/s3.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/url"
+	"strconv"
 	"sync"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -19,7 +20,12 @@ import (
 	"github.com/cockroachdb/molt/testutils"
 	"github.com/cockroachdb/molt/utils"
 	"github.com/rs/zerolog"
+	"golang.org/x/sync/errgroup"
 )
+
+// AWS capitalizes the first letter of the string and lowercases the rest, hence
+// the naming discrepancy between this and GCP.
+const numRowKeysAWS = "Numrows"
 
 type s3Store struct {
 	logger      zerolog.Logger
@@ -137,11 +143,13 @@ func (s *s3Store) CreateFromReader(
 	}
 
 	rows := <-numRows
+	numRowStr := fmt.Sprintf("%d", rows)
 
 	if _, err := uploader.UploadWithContext(ctx, &s3manager.UploadInput{
-		Bucket: aws.String(bucketName),
-		Key:    aws.String(key),
-		Body:   r,
+		Bucket:   aws.String(bucketName),
+		Key:      aws.String(key),
+		Body:     r,
+		Metadata: map[string]*string{numRowKeysAWS: &numRowStr},
 	}); err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


fetch/datablobstorage/datablobstorage.go
```diff
@@ -16,6 +16,8 @@ import (
 	"golang.org/x/oauth2/google"
 )
 
+const numRowsKey = "numRows"
```
Jeremyyang920, line 19: Since you need a const for Aws, might as well move this to GCP since for local, we are probably just writing a single row at the top with no key, just value.

fetch/datablobstorage/s3.go
```diff
@@ -184,23 +196,65 @@ func listFromContinuationPointAWS(
 		return nil, err
 	}
 
-	resources := []Resource{}
-	for _, obj := range s3Objects.Contents {
+	g, _ := errgroup.WithContext(ctx)
+	resources := make([]Resource, len(s3Objects.Contents))
+	for i, obj := range s3Objects.Contents {
+		curI := i
+		curObj := obj
```
Jeremyyang920, line 203: Note for us, but we should upgrade to go 1.22 to not have to do this since they fixed the whole loopVar issue. But this is fine for now.

fetch/datablobstorage/s3.go
```diff
@@ -184,23 +196,65 @@ func listFromContinuationPointAWS(
...
+				Bucket: aws.String(s3Store.bucket),
+				Key:    aws.String(*curObj.Key),
 			})
+			if err != nil {
+				return err
+			}
+
+			mdNumRows, ok := objResp.Metadata[numRowKeysS3Resp]
+			if !ok {
+				return errors.New("could not find number of rows in metadata")
```
Jeremyyang920, line 221: We shouldnt error here. If the file is at least there, but metadata is not set for some reason, i think import shoudl still happen.

fetch/datablobstorage/s3.go
```diff
@@ -184,23 +196,65 @@ func listFromContinuationPointAWS(
 		return nil, err
 	}
 
-	resources := []Resource{}
-	for _, obj := range s3Objects.Contents {
+	g, _ := errgroup.WithContext(ctx)
```
Jeremyyang920, line 199: Whats the other argument to this were ignoring?

fetch/datablobstorage/s3.go
ryanluu12345: Yep totally makes sense. We should do a bump when we get some down time? Next week during breather week?

fetch/datablobstorage/s3.go
ryanluu12345: Yep gave the same treatment I gave in GCP

fetch/datablobstorage/s3.go
ryanluu12345: We're ignoring `context.Context`, which is not needed. We don't need the updated context.

fetch/datablobstorage/datablobstorage.go
ryanluu12345: Yep moved the consts, we should be good.

fetch/datablobstorage/s3.go
```diff
@@ -19,8 +20,12 @@ import (
 	"github.com/cockroachdb/molt/testutils"
 	"github.com/cockroachdb/molt/utils"
 	"github.com/rs/zerolog"
+	"golang.org/x/sync/errgroup"
 )
 
+// Weirdly enough, it captalizes first character and lowercases everything else....
```
Jeremyyang920, line 26: Lets make this comment more standard and not casual.

Jeremyyang920 (approved):

fetch/datablobstorage/s3.go
ryanluu12345: Done.

PR #138, opened 2024-02-08
fetch: fix csv_pipe test

ZhouXing19:
To ensure that goroutines where the numrows are push to the channel happens one
by one, i.e. no concurrent goroutines.

Release note: None


Jeremyyang920 (approved):
Nice find! Many rabbit holes this week.

ryanluu12345 (approved):

PR #146, opened 2024-02-14
Fetch: Pipe Fix/Add Pprof

Jeremyyang920:
See individual commits.

### DIFF
--- a/fetch/export_table.go
+++ b/fetch/export_table.go
@@ -52,8 +52,12 @@ func exportTable(
 
 	cancellableCtx, cancelFunc := context.WithCancel(ctx)
 	defer cancelFunc()
-	buf := new(bytes.Buffer)
-	sqlRW := NewPipe(buf)
+	// We use the standard io.Pipe here because during data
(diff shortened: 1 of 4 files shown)


fetch/export_table.go
```diff
@@ -52,8 +52,12 @@ func exportTable(
 
 	cancellableCtx, cancelFunc := context.WithCancel(ctx)
 	defer cancelFunc()
-	buf := new(bytes.Buffer)
-	sqlRW := NewPipe(buf)
+	// We use the standard io.Pipe here because during data
```
ryanluu12345, line 55: What are implications of using `io.Pipe`? Are there any downsides?

ryanluu12345 (approved):

fetch/export_table.go
Jeremyyang920: Theres no implication for this portion of the pipe as we always have a reader on the other end. During file creation we cant use this because we don't have a reader until the rows channel is populated.

PR #151, opened 2024-02-20
fetch/datablobstorage: support local files numRows metadata

ryanluu12345:
Resolves: CC-27188
Release Note: None

Existing CI passing should be enough to validate this works.

Manually tested too:
```
molt_fetch_overall_duration 1.3976359999999999
# HELP molt_fetch_rows_exported Number of rows that have been exported by table.
# TYPE molt_fetch_rows_exported counter
molt_fetch_rows_exported{table="public.employees"} 200000
# HELP molt_fetch_rows_imported Number of rows that have been imported by table.
# TYPE molt_fetch_rows_imported counter
molt_fetch_rows_imported{table="public.employees"} 200000
```

### DIFF
--- a/fetch/datablobstorage/local.go
+++ b/fetch/datablobstorage/local.go
@@ -1,6 +1,8 @@
 package datablobstorage
 
 import (
+	"bufio"
+	"compress/gzip"
 	"context"
 	"fmt"
 	"io"
@@ -9,6 +11,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/cockroachdb/errors"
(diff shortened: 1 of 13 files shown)


fetch/datablobstorage/local.go
```diff
@@ -143,16 +168,56 @@ func (l *localStore) ListFromContinuationPoint(
...
+	var reader io.Reader
+	f, err := os.Open(filePath)
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+	reader = f
+
+	// We need a gzip reader if we have a gzip extension.
+	if strings.Contains(filePath, "tar.gz") {
```
Jeremyyang920, line 202: Why not have this function take a boolean from above since you already know if its a tar.gz? Skip needing to do the string.Contains.

Jeremyyang920: We should probably use filepath.ext for this instead of strings.Contains.

fetch/import_table.go
```diff
@@ -142,9 +142,14 @@ func importTable(
...
 	}
 
+	var isLocal bool
 	var locs []string
 	var numRows []int
 	for _, resource := range resources {
+		if resource.IsLocal() {
+			isLocal = true
+		}
+
```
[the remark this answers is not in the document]
ryanluu12345: Fixed this!

fetch/datablobstorage/local.go
ryanluu12345: Went for this instead:
```
// We need a gzip reader if we have a gzip extension.
	if filepath.Ext(filePath) == ".gz" {
```

ryanluu12345: Fixed!

Jeremyyang920 (approved):

PR #152, opened 2024-02-21
[CC-27909] fetch: Add implicit auth for gcp

Jeremyyang920:
This commit allows the gcp resource to create an import url for implicit auth. This will be used when users do not specify the gcp env var or have a credential file in the well known location.

Release Note: Adds implicit auth for gcp

### DIFF
--- a/fetch/datablobstorage/gcp.go
+++ b/fetch/datablobstorage/gcp.go
@@ -197,6 +197,19 @@ type gcpResource struct {
 }
 
 func (r *gcpResource) ImportURL() (string, error) {
+	// If we do not have the GOOGLE_APPLICATION_CREDENTIALS env var
+	// set, or do not have the gcloud/application_default_credentials.json file
+	// in a well known directory, then we authed using the GCE machine itself
+	// which will generate an oauth token using the service account that is
+	// active meaning we need to use AUTH=implicit as store.creds.JSON is nil.
+	if r.store.creds.JSON == nil {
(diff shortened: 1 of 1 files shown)


fetch/datablobstorage/gcp.go
```diff
@@ -197,6 +197,19 @@ type gcpResource struct {
 }
 
 func (r *gcpResource) ImportURL() (string, error) {
+	// If we do not have the GOOGLE_APPLICATION_CREDENTIALS env var
```
ryanluu12345, line 200: Can be in another PR, but worth adding an e2e test with this for the GCP case

ryanluu12345: Tracking here: https://docs.google.com/spreadsheets/d/1YsLWaIYgQReBWzxid6MKrsRYoTDeEjRWDGFNgAQl0y8/edit?usp=sharing

ryanluu12345 (approved):

fetch/datablobstorage/gcp.go
Jeremyyang920: You can't really add an e2e for this. The implicit auth either happens on a GCE worker directly, or you are exporting the json service account file in an env var. Unless you are talking about the env var case. But in which case we then have to check in credentials to the repo which we wont be able to do. This tests more the GCP SDK more than it does the actual fetch code.

ZhouXing19 (approved):

PR #153, opened 2024-02-21
fetch: fix ordering of primary key

ajstorm:
Previously, in cases where a table in fetch/verify has a compund primary key whose ordering doesn't match that of the order of the columns in the base table, fetch and verify would order the columns incorrectly. This would produce incorrect results in verify, but more importantly, it would make fetch insanely slow. The reason why has to do with how fetch runs export. To avoid issuing a single large query, fetch breaks the export down into batch-sized chunks (the default batch size is 100K rows). To ensure that it knows where it stopped, it leverages the primary key and issues a SELECT <column names> order by <PK columns> limit <batch_size>. Then, it records the PK value at which it left off and issues subsequent SELECT <column names> where PK > <left off value> order by <PK columns> limit <batch_size>.

The problem with not getting the PK ordering right is that instead of leveraging the PK for the query, fetch instead has to sort the entire table each time it issues a query. This is most pronounced in cases where the table being fetched is large. For a 104M table, tests showed that it would take 17 hours to perform the export phase. With the fix to get the PK ordering correct, the export takes 3 minutes.

### DIFF
--- a/verify/tableverify/query.go
+++ b/verify/tableverify/query.go
@@ -186,15 +186,20 @@ func getPrimaryKey(
 		rows, err := conn.Query(
 			ctx,
 			`
-select
-    a.attname as column_name
-from
-    pg_class t
-    join pg_attribute a on a.attrelid = t.oid
-    join pg_index ix    on t.oid = ix.indrelid AND a.attnum = ANY(ix.indkey)
-    join pg_class i     on i.oid = ix.indexrelid  
-where
-    t.oid = $1 AND indisprimary;
+SELECT
+    a.attname AS column_name
+FROM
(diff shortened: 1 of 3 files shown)


Jeremyyang920 (approved):
Big Find! We should probably test and see if this also affects large MySQL data sets too.

ryanluu12345 (approved):
Echoing @Jeremyyang920 's comment. HUGE!

ajstorm:
> We should probably test and see if this also affects large MySQL data sets too. 

I added a MySQL test too, and it appears as thought he MySQL query was always good.

Thanks for the review!

PR #155, opened 2024-02-22
.github/workflows: fix create release ticket workflow

ryanluu12345:
Previously, characters were not getting escaped. Add quotes around fields so that they are properly escaped.

Release Note: None

Reference: https://stackoverflow.com/questions/61795201/github-action-failed-syntax-error-near-unexpected-token


ryanluu12345:
Test run here showing that we can now echo: https://github.com/cockroachdb/molt/actions/runs/8009269379/job/21877541940

PR #156, opened 2024-02-23
mysqlcov: fix for bits conversion, support time conversion

ZhouXing19:
Previous, the conversion of the bits type value from mysql to crdb is wrong, as it didn't correctly parse the value passed via the []uint8 val variable. We will get a 

```
22P02 Wraps: (2) could not parse string as bit array: "" is not a valid binary digit
```

error when trying to convert something like `b'11110000011'` into a `*tree.DBitArray`.

This PR is to fix it and add tests for it. 

Also added support for converting the time type.

Release note (bug fix): enable conversion of bit type entry from mysql to crdb. Add support for time type conversion too.

### DIFF
--- a/mysqlconv/testdata/rowvalue/bits.ddt
+++ b/mysqlconv/testdata/rowvalue/bits.ddt
@@ -1,0 +1,34 @@
+convert
+b'11110000011'
+----
+error: no oid provided
+
+convert oid=1560
+b'11110000011'
+----
+*tree.DBitArray:B'11110000011';
+
+convert oid=1560
+b'111111111110000011'
+----
+*tree.DBitArray:B'111111111110000011';
+
+convert oid=1560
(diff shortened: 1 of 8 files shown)


mysqlconv/testdata/rowvalue/bits.ddt
```diff
@@ -0,0 +1,19 @@
+convert 1560
```
ZhouXing19, line 1: The reason i make the oid hard-coded here, is that in the code, we retrieve the oids via query the missing table via the mysql conn (see the usage of `mysqlconv.ScanRowDynamicTypes()` and how `mysqlRows.typOIDs` is written). This requires setting a dedicated "verified table" and is just cumbersome for such a simple unit test.

Jeremyyang920 (approved):

PR #157, opened 2024-02-24
fetch: added better testing for continuation and fixed behavior for no exceptions log found

ryanluu12345:
Previously, when no exception logs for a given token were found, we would silently operate and not let the user know. Now we error and let the user know that there were no exception logs found, which helps the user to use a valid token. This could lead to less confusing behavior.

This change also adds coverage for the following cases:
- Continuing from a file greater than part 1
- If a continuation token is no longer present, we error and let user know
- If a continuation token links to an exception log that doesn't have a file name, we error out because no file name is present

Resolves: CC-27047
Release Note: None

### DIFF
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -215,9 +215,30 @@ func Fetch(
 	}
 	var stats statsMu
 
+	// If continuation file is passed in, it must conform to the file format.
+	if cfg.ContinuationFileName != "" && !utils.FileConventionRegex.Match([]byte(cfg.ContinuationFileName)) {
+		return errors.Newf("continuation file name %s doesn't match the format %s", cfg.ContinuationFileName, utils.FileConventionRegex.String())
+	}
+
 	exceptionLogMapping, err := getExceptionLogMapping(ctx, cfg, targetPgxConn)
-	if err != nil {
+	contTokenNotFoundErr := fmt.Sprintf("no exception logs that correspond to continuation-token of %s", cfg.ContinuationToken)
+
+	// In the case that we have no results for the passed in continuation token or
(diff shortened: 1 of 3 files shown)


fetch/fetch.go
```diff
@@ -215,11 +215,35 @@ func Fetch(
 	}
 	var stats statsMu
 
+	// If continuation file is passed in, it must conform to the file format.
+	if cfg.ContinuationFileName != "" && !utils.FileConventionRegex.Match([]byte(cfg.ContinuationFileName)) {
+		return errors.Newf("continuation file name %s doesn't match the format %s", cfg.ContinuationFileName, utils.FileConventionRegex.String())
```
ZhouXing19, line 220: Good practice with the check for format, helping avoid SQL injection

ZhouXing19 (approved):
LGTM, thanks so much for the detailed comments for the tests! I'll let Jeremy to give the stamp.

fetch/fetch.go
```diff
@@ -215,11 +215,35 @@ func Fetch(
...
+	if IsImportCopyOnlyMode(cfg) && len(exceptionLogMapping) == 0 {
+		errMsg := fmt.Sprintf("no exception logs that correspond to fetch-id of %s", cfg.FetchID)
+		if cfg.ContinuationToken != "" {
+			errMsg = contTokenNotFoundErr
+		}
+
+		return errors.New(errMsg)
+	}
+
+	// TODO(janexing): ingest the schema creation logic here. We create the schemas
```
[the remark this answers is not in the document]
ryanluu12345: Removing!

fetch/fetch.go
ryanluu12345: Thanks!

PR #160, opened 2024-02-26
fetch: support mysql schema creation

ZhouXing19:
- [x] **e2e test passed**: https://github.com/cockroachdb/molt/actions/runs/8068656252

Release note: support the `--table-handling 'drop-on-target-and-recreate'` flag for mysql so that a new schema can be automatically created if the source db is mysql.


fetch/fetch.go
```diff
@@ -146,9 +146,9 @@ func Fetch(
 				if err != nil {
 					return err
 				}
-				logger.Info().Msgf("dropping table with %q", dropTableStmt)
+				logger.Debug().Msgf("dropping table with %q", dropTableStmt)
 				if _, err := targetConn.Exec(ctx, dropTableStmt); err != nil {
-					return errors.Wrapf(err, "failed to create new schema %q on the target connection", t)
+					return errors.Wrapf(err, "failed to drop if exists new schema %q on the target connection", t)
```
Jeremyyang920, line 151: Hmm should this be failed to drop table?

fetch/schema_creation.go
```diff
@@ -209,12 +244,38 @@ ORDER BY
...
+		rows, err := conn.Query(q)
+		if err != nil {
+			return nil, err
+		}
+		for rows.Next() {
+			newCol := columnWithType{}
+			if err := rows.Scan(&newCol.schemaName, &newCol.tableName, &newCol.columnName, &newCol.dataType, &newCol.columnType, &newCol.nullable, &newCol.isPrimaryKey); err != nil {
+				return nil, errors.Wrap(err, "failed to scan query result to a columnWithType object")
+			}
+			logger.Debug().Msgf("collect column:%s", newCol.String())
```
Jeremyyang920, line 258: I missed this in the PG pr but I see we have the same log, but could you clarify what "collect column" means here?

fetch/schema_creation.go
```diff
@@ -187,8 +194,36 @@ LEFT JOIN (
...
+JOIN 
+    information_schema.TABLES t ON c.TABLE_SCHEMA = t.TABLE_SCHEMA AND c.TABLE_NAME = t.TABLE_NAME
+WHERE 
+    c.TABLE_SCHEMA = DATABASE() 
+    AND t.TABLE_TYPE = 'BASE TABLE'
+    AND c.TABLE_NAME = '%s'
+ORDER BY 
+    c.TABLE_SCHEMA, 
+    c.TABLE_NAME,  
+    c.ORDINAL_POSITION;
```
Jeremyyang920, line 226: Just the same note as the PG of whether any given user would be able to run this query.

Jeremyyang920 (approved):
Great work! This will be a big win for us on schema creation. Just a few minor notes, but nice extensive tests as well for this.

fetch/schema_creation.go
ZhouXing19: Oh it meant `collected column`. Updated.

fetch/schema_creation.go
ZhouXing19: Updated with more details in https://cockroachlabs.atlassian.net/jira/software/c/projects/CC/boards/375?quickFilter=1434&selectedIssue=CC-27287

PR #161, opened 2024-02-27
[CC-27316]pgconv: fix conversion for float32 (float4) and fmt for negative float

ZhouXing19:
Previously, to convert a `float32` golang value to a `tree.DFloat`, which is a `float64`, additional non-zero decimals are padded (e.g. `123.12 -> 123.12000274658203`), leading to incorrect result when comparing the 2 datums. 

e.g.:

```
{"level":"warn",...,"source_values":{"floatcol":"918.9"},"target_values":{"floatcol":"918.9000244140625"},"primary_key":["2"],"message":"mismatching row value"}
```

This commit fixes it by making sure the new tree.DFloat is consistent with the original `float32` value.

We also fixed a format issue with float type so that when negative float value is exported to a csv, it is not bracket. It's because `(-123,12)` cannot be parsed as float when being imported to a table.

- [x]  **Passed E2E test**: https://github.com/cockroachdb/molt/actions/runs/8067487310

Release note (bug fix): correctly support verification of float32 (mysql's float4) value, and fmt issue with float.


Jeremyyang920 (approved):
Nice finds!

PR #164, opened 2024-02-27
[CC 27325] verify: fix {`TINYINT` <> `BOOL`} & {`UUID` <> `STRING`} comparison

ZhouXing19:
Previously, if we have a bool column in MySQL comparing a bool column in CockroachDB, we will get a `unsupported comparison: bool to int` error. This is because in cockroach or PG we don't support comparison such as `SELECT true=1`. This can happen when we do verify between a mysql and a cockroach table.

This commit adds an explicit conversion for this case -- if we notice this is a bool vs int comparison, explicitly convert the bool to an int datum.

Similar fix is done for a `UUID` and `STRING` comparison as well. 

Informs: #103 

- [x] e2e passed: https://github.com/cockroachdb/molt/actions/runs/8082383333

Release note (bug fix): enable bool to int and uuid to string comparison for verify.


verify/rowverify/rowverify.go
```diff
@@ -164,8 +165,43 @@ func verifyRows(
...
+						leftDatum = boolDatumToIntDatum(truthDatum.(*tree.DBool))
+					} else if truthDatum.ResolvedType().Family() == types.IntFamily && targetDatum.ResolvedType().Family() == types.BoolFamily {
+						rightDatum = boolDatumToIntDatum(targetDatum.(*tree.DBool))
+					}
+
+					if truthDatum.ResolvedType().Family() == types.UuidFamily && targetDatum.ResolvedType().Family() == types.StringFamily {
+						leftDatum = uuidDatumToStringDatum(truthDatum.(*tree.DUuid))
+					} else if targetDatum.ResolvedType().Family() == types.UuidFamily && truthDatum.ResolvedType().Family() == types.StringFamily {
+						rightDatum = uuidDatumToStringDatum(targetDatum.(*tree.DUuid))
+					}
```
Jeremyyang920, line 202: nit: Im wondering if this might be cleaner to make into a switch statement on truthDatum.ResolvedType.Family() since right now all columns would try all 4 if conditions since it has to check the equality before short circuiting.

Jeremyyang920 (approved):
Just that one small comment. But thanks for fixing this!

verify/rowverify/rowverify.go
ZhouXing19: Updated the switch.

PR #165, opened 2024-02-29
fetch: added in logic to enforce one active set of continuation tokens at a time     

ryanluu12345:
The behavior introduced makes it so that customers only have one set of continuation tokens that are tied to a single fetch ID. This solves the problem of not using stale tokens that may correspond to older versions of the data set. At a high level, this is the behavior:

- If not import-copy only mode: we wipe the existing tokens before we run, generate a new fetch ID, tie the exceptions log entries to new fetch ID
- If import-copy only mode but ONLY fetch ID specified: we wipe the existing tokens before we run, generate a new fetch ID, tie the exceptions log entries to new fetch ID
- If import-copy only mode and BOTH fetch ID and cont token specified: we keep the existing tokens; if it succeeds, we delete the token from the exceptions log table; if it fails, we update the entry with the fetch ID with the latest details
 
Resolves: CC-27326
Release Note: None

### DIFF
--- a/cmd/fetch/fetch.go
+++ b/cmd/fetch/fetch.go
@@ -317,6 +317,15 @@ func Command() *cobra.Command {
 		"",
 		"If set, restarts the fetch process for at the given file name instead of recorded file in the exceptions table",
 	)
+
+	const nonInteractiveStr = "non-interactive"
+	cmd.PersistentFlags().BoolVar(
+		&cfg.NonInteractive,
+		nonInteractiveStr,
+		false,
+		`If set, automatically skips all user prompting and initiates actions such as clearing exception log data (preferable if running in CI)
+or as an automated job. If not set, prompts user for confirmation before performing actions.`,
+	)
 	moltlogger.RegisterLoggerFlags(cmd)
 	cmdutil.RegisterDBConnFlags(cmd)
 	cmdutil.RegisterNameFilterFlags(cmd)
@@ -334,6 +343,10 @@ func Command() *cobra.Command {
 	)
 
 	if err := cmd.PersistentFlags().MarkHidden(testOnlyFlagStr); err != nil {
+		panic(err)
+	}
+
+	if err := cmd.PersistentFlags().MarkHidden(nonInteractiveStr); err != nil {
 		panic(err)
 	}
 
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cockroachdb/cockroachdb-parser/pkg/util/uuid"
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/molt/compression"
 	"github.com/cockroachdb/molt/dbconn"
@@ -45,6 +46,11 @@ type Config struct {
 
 	DropAndRecreateNewSchema bool
 
+	// NonInteractive relates to if user input should be prompted. If false,
+	// user prompting is initiating before certain actions like wiping data.
+	// If true, user prompting will be skipped and actions will be confirmed automatically.
+	NonInteractive bool
+
 	Compression    compression.Flag
 	ExportSettings dataexport.Settings
 }
@@ -64,7 +70,7 @@ func Fetch(
 		return errors.New("failed to assert conn as a pgconn")
 	}
 	targetPgxConn := targetPgConn.Conn
-	fetchStatus, err := initStatusEntry(ctx, targetPgxConn, conns[0].Dialect())
+	fetchStatus, err := initStatusEntry(ctx, cfg, targetPgxConn, conns[0].Dialect())
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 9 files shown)


cmd/fetch/fetch.go
```diff
@@ -317,6 +317,13 @@ func Command() *cobra.Command {
 		"",
 		"If set, restarts the fetch process for at the given file name instead of recorded file in the exceptions table",
 	)
+	cmd.PersistentFlags().BoolVar(
+		&cfg.NonInteractive,
+		"non-interactive",
+		false,
+		`If set, automatically skips all user prompting and initiates actions such as clearing exception log data (preferable if running in CI)
+or as an automated job. If not set, prompts user for confirmation before performing actions.`,
+	)
```
Jeremyyang920, line 326: Let mark this hidden

fetch/status/exceptions.go
```diff
@@ -90,12 +94,58 @@ func (e *ExceptionLog) CreateEntry(ctx context.Context, conn *pgx.Conn, stage st
...
+func (e *ExceptionLog) DeleteEntry(ctx context.Context, conn *pgx.Conn) error {
+	query := fmt.Sprintf(`DELETE FROM %s WHERE id=@id`, exceptionTable)
+	args := pgx.NamedArgs{
+		"id": e.ID,
+	}
+
+	_, err := conn.Exec(ctx, query, args)
+	if err != nil {
+		return err
+	}
```
Jeremyyang920, line 128: nit: if _,,err := conn.Exec(ctx, query, args); err!=nil{} since this isint used anywhere else.

fetch/status/exceptions.go
```diff
@@ -90,12 +94,58 @@ func (e *ExceptionLog) CreateEntry(ctx context.Context, conn *pgx.Conn, stage st
...
+
+	return nil
+}
+
+// Used to clear the continuation tokens on fresh runs.
+func DeleteAllExceptionLogs(ctx context.Context, conn *pgx.Conn) error {
+	_, err := conn.Exec(ctx, deleteQuery)
+	if err != nil {
+		return err
+	}
```
Jeremyyang920, line 138: Same comment as above.

fetch/status/exceptions.go
```diff
@@ -90,12 +94,58 @@ func (e *ExceptionLog) CreateEntry(ctx context.Context, conn *pgx.Conn, stage st
...
+		"time":      time.Now(),
+	}
+
+	row := conn.QueryRow(ctx, query, args)
+	if err := row.Scan(&e.Message, &e.SQLState, &e.FileName, &e.Time); err != nil {
+		return err
+	}
+
+	return nil
+}
```
Jeremyyang920, line 115: This can just be simplified to return row.Scan...

cmd/fetch/fetch.go
ryanluu12345: Got it.

fetch/status/exceptions.go
ryanluu12345: Done.

fetch/status/exceptions.go
ryanluu12345: DOne!

fetch/status/exceptions.go
ryanluu12345: Yep

ryanluu12345:
@Jeremyyang920 tests are all fixed and new ones introduced

fetch/fetch.go
```diff
@@ -241,6 +252,29 @@ func Fetch(
...
+	isClearContinuationTokenMode := IsClearContinuationTokenMode(cfg)
+	if isClearContinuationTokenMode {
+		if cfg.NonInteractive {
+			logger.Warn().Msg("clearing all continuation tokens because running in clear continuation mode")
+		} else {
+			fmt.Println("Clearing all continuation tokens. Confirm (y/n)?")
+			var confirmation string
+			fmt.Scanln(&confirmation)
+
+			if confirmation != "y" {
```
Jeremyyang920, line 268: lets use strings.EqualFolds here instead for case insensitivity.

fetch/fetch.go
```diff
@@ -251,6 +285,7 @@ func Fetch(
 					return nil
 				}
 
+				fmt.Println(table.SafeString())
```
Jeremyyang920, line 288: Remove extra println

fetch/status/exceptions.go
```diff
@@ -19,9 +19,12 @@ import (
 const (
 	StageSchemaCreation = "schema_creation"
 	StageDataLoad       = "data_load"
+	exceptionTable      = "_molt_fetch_exception"
```
Jeremyyang920, line 22: Wonder if we should name this exceptions instead of exception.

Jeremyyang920 (approved):
Nice work! Thanks for the detailed notes in the test so we know what each step is.

fetch/fetch.go
ryanluu12345: Good call!

fetch/fetch.go
ryanluu12345: Will do

fetch/status/exceptions.go
ryanluu12345: Let me do a follow up change for this btw.

ryanluu12345: I have another change in flight I can tag this onto.

PR #166, opened 2024-02-29
[CC-27357] fetch: enable fetch dropped constraints

ZhouXing19:
Now we can collect the non-primary constraints from the given table. It is logged as a warn. 

Example:

```
{"level":"warn","message":"newly created schema doesn't contain the following constraints:\ntable: public.posts,\"UNIQUE (author_id)\",\"UNIQUE (number_pg)\",\"FOREIGN KEY (tenant_id, author_id) REFERENCES users(tenant_id, user_id) ON DELETE SET NULL\",\"FOREIGN KEY (tenant_id) REFERENCES tenants(tenant_id) ON DELETE RESTRICT\""}
```

- [x]  **E2E passed**: https://github.com/cockroachdb/molt/actions/runs/8102733772

Jira ticket: https://cockroachlabs.atlassian.net/browse/CC-27357
Release note: none

### DIFF
--- a/e2e/testdata/mysql/bool.ddt
+++ b/e2e/testdata/mysql/bool.ddt
@@ -37,6 +37,7 @@ fetch --source 'mysql://user:password@0.0.0.0:3306/defaultdb' --target 'postgres
 {"level":"info","message":"getting column types for table: public.bool_table"}
 {"level":"info","message":"finished getting column types for table: public.bool_table"}
 {"level":"info","message":"creating new table with \"CREATE TABLE bool_table (id INT8 NOT NULL PRIMARY KEY, bool_col INT2)\""}
+{"level":"warn","message":"newly created schema doesn't contain the following constraints:\ntable: public.bool_table,\"UNIQUE KEY `id` (`id`)\""}
 {"level":"info","message":"after recreating table, dbTables: {[[public.bool_table public.bool_table]] [] []}"}
 {"level":"info","message":"verifying common tables"}
 {"level":"info","message":"establishing snapshot"}
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -161,6 +161,16 @@ func Fetch(
 					return errors.Wrapf(err, "failed to create new schema %q on the target connection with %q", t, createTableStmt)
 				}
 				logger.Debug().Msgf("finished creating new table with %q", createTableStmt)
+
+				// TODO(janexing): maybe persist it in table?
+				droppedConstraints, err := GetConstraints(ctx, logger, conns[0], t)
+				if err != nil {
+					return err
+				}
+				if len(droppedConstraints) != 0 {
+					consWithTable := constraintsWithTable{table: t, cons: droppedConstraints}
+					logger.Warn().Msgf("newly created schema doesn't contain the following constraints:\n%s", consWithTable.String())
+				}
 			}
 			// Redo the verify.
 			dbTables, err = dbverify.Verify(ctx, conns)
--- a/fetch/schema_creation.go
+++ b/fetch/schema_creation.go
@@ -1,6 +1,7 @@
 package fetch
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"regexp"
@@ -68,16 +69,17 @@ func (cs columnsWithType) CRDBCreateTableStmt() (string, error) {
 }
 
 type columnWithType struct {
-	schemaName    string
-	tableName     string
-	columnName    string
-	dataType      string
-	columnType    string
-	typeOid       oid.Oid
-	nullable      bool
-	isPrimaryKey  bool
-	udtName       string
-	udtDefinition string
+	schemaName      string
+	tableName       string
+	columnName      string
+	dataType        string
+	columnType      string
+	typeOid         oid.Oid
+	nullable        bool
+	isPrimaryKey    bool
+	udtName         string
+	udtDefinition   string
+	ordinalPosition int
 }
 
 func (t *columnWithType) CRDBColDef(includePk bool) (*tree.ColumnTableDef, error) {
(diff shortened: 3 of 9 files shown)


e2e/testdata/mysql/bool.ddt
```diff
@@ -37,6 +37,7 @@ fetch --source 'mysql://user:password@0.0.0.0:3306/defaultdb' --target 'postgres
 {"level":"info","message":"getting column types for table: public.bool_table"}
 {"level":"info","message":"finished getting column types for table: public.bool_table"}
 {"level":"info","message":"creating new table with \"CREATE TABLE bool_table (id INT8 NOT NULL PRIMARY KEY, bool_col INT2)\""}
+{"level":"warn","message":"newly created schema doesn't contain the following constraints:\ntable: public.bool_table,\"UNIQUE KEY `id` (`id`)\""}
```
Jeremyyang920, line 40: Is this correct here? Since doesn't the statement about get created with Primary key id?

fetch/schema_creation.go
```diff
@@ -341,3 +344,100 @@ func convertMySQLEnum(
...
+        pg_catalog.pg_get_constraintdef(c.oid) AS constraint_def
+        FROM pg_catalog.pg_class s
+        JOIN pg_catalog.pg_constraint c ON (s.oid = c.conrelid)
+        WHERE conparentid = 0 
+          AND s.relkind = 'r' -- 'r' indicates a table (relation)
+          AND c.contype != 'p' -- 'p' indicates a primary key constraint
+          AND s.relname= $1
+          AND s.relnamespace::regnamespace::text = $2 
+        ORDER BY conrelid, conname;`
+		mysqlQuery = `SHOW CREATE TABLE %s`
```
Jeremyyang920, line 381: I like how simple this is in MySQL and Pg has this giant query.

fetch/fetch.go
```diff
@@ -161,6 +161,16 @@ func Fetch(
 					return errors.Wrapf(err, "failed to create new schema %q on the target connection with %q", t, createTableStmt)
 				}
 				logger.Debug().Msgf("finished creating new table with %q", createTableStmt)
+
+				// TODO(janexing): maybe persist it in table?
```
Jeremyyang920, line 165: Something to consider too is adding the constraints back for them as well (optional mode). But we can track that in a ticket.

e2e/testdata/mysql/bool.ddt
ZhouXing19: Yeah i felt torn about this too -- it's because `SERIAL` in mysql implies the unique constraint:

```
mysql> CREATE TABLE serialt (x int PRIMARY KEY, y serial);
Query OK, 0 rows affected (0.03 sec)

mysql> show create table serialt;
+---------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table   | Create Table                                                                                                                                                                                               |
+---------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| serialt | CREATE TABLE `serialt` (
  `x` int NOT NULL,
  `y` bigint unsigned NOT NULL AUTO_INCREMENT,
  PRIMARY KEY (`x`),
  UNIQUE KEY `y` (`y`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci |
+---------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

In this case it overlaps with the PK constraint. It's a bit tricky to distinguish these cases. 

What about I change the warning message to a softer notice:`newly created schema might not contain all the following constraints:...`. What do you think?

fetch/fetch.go
ZhouXing19: https://cockroachlabs.atlassian.net/browse/CC-27358
Not sure if this is really needed for GA but i added it under that parent ticket

Jeremyyang920: Yes, definitely not for GA. Its a nice to have.

e2e/testdata/mysql/bool.ddt
Jeremyyang920: Ohh I see. Hmm yea thats tricky. Might be a question for Ryan K in terms of the copy and what makes the most sense. 

Does this only affect serial in this case?

ZhouXing19: Yes in this example it's only serial that implies such implicit constraint. In MySQL's doc, the explicit constraints are PK, FK, Unique, Check, which are all tested here. I'm not sure if there are other cases that might contains implicit constraint as well. And hopefully more (maybe rather randomized) test can help reveal them in the future?

Jeremyyang920: I see. Ive though't about it some more, I think the log makes sense when you understand it from the technical sense. This might just be one of those things we document down.

fetch/schema_creation.go
```diff
@@ -341,3 +344,100 @@ func convertMySQLEnum(
...
+			res = append(res, formatMySQLConstraints(createTableStmt)...)
+		}
+	}
+	return res, nil
+}
+
+func removeTrailingComma(input string) string {
+	if strings.HasSuffix(input, ",") {
+		return input[:len(input)-1]
+	}
```
Jeremyyang920, line 418: I think you can just strings.TrimSuffix here. If theres no comma then nothing happens instead of having to reslice the input.

Jeremyyang920 (approved):
This is awesome that you were able to add this!

e2e/testdata/mysql/bool.ddt
ZhouXing19: SG, tracked in https://cockroachlabs.atlassian.net/browse/CC-27371

PR #168, opened 2024-03-01
fetch and e2e: update to _molt_fetch_exceptions and fix e2e tests to use non-interactive

ryanluu12345:
Kicked off E2E tests here: https://github.com/cockroachdb/molt/actions/runs/8116472144

Release Note: None


fetch/status/exceptions.go
```diff
@@ -20,7 +20,7 @@ import (
 const (
 	StageSchemaCreation = "schema_creation"
 	StageDataLoad       = "data_load"
-	exceptionTable      = "_molt_fetch_exception"
+	exceptionTable      = "_molt_fetch_exceptions"
```
Jeremyyang920, line 23: nit:
```suggestion
	exceptionsTable      = "_molt_fetch_exceptions"
```

Jeremyyang920 (approved):

fetch/status/exceptions.go
ryanluu12345: Done

PR #169, opened 2024-03-04
fetch: unified continuation mode behavior so that specific token and multiple token behavior is the same

ryanluu12345:
Previously, the multiple token and single-specific token case were treated differently, leading to potential continuation token loss when errors occur during non continuation logic. This means that we previously truncated the exceptions table before we even ran the import. Now, we unify the logic so that the behavior is as follows: if a table associated with a token fails, we update the associated entry with new information; if it succeeds, we remove the token. This happens within goroutines.

We also moved the connection cloning to the top of `fetchTable` so it can be used throughout the method.

Release Note: None

### DIFF
--- a/fetch/datablobstorage/copy_direct.go
+++ b/fetch/datablobstorage/copy_direct.go
@@ -7,7 +7,6 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/molt/dbtable"
 	"github.com/cockroachdb/molt/fetch/internal/dataquery"
-	"github.com/cockroachdb/molt/fetch/status"
 	"github.com/cockroachdb/molt/testutils"
 	"github.com/jackc/pgx/v5"
 	"github.com/rs/zerolog"
@@ -47,11 +46,7 @@ func (c *copyCRDBDirect) CreateFromReader(
 
 	c.logger.Debug().Int("batch", iteration).Msgf("csv batch starting")
 	if _, err := conn.PgConn().CopyFrom(ctx, r, dataquery.CopyFrom(table, false /*skipHeader*/)); err != nil {
-		// In direct copy mode, we cannot continue, so might as well just make a new entry each time.
-		// Default to not clearing the continuation tokens and pass in a nil exception log so it
-		// creates the new entry.
-		pgErr := status.MaybeReportException(ctx, c.logger, conn, table.Name, err, "" /* fileName */, status.StageDataLoad, false /* isClearContinuationTokenMode */, nil /* exceptionLog */)
-		return nil, errors.CombineErrors(pgErr, conn.Close(ctx))
+		return nil, errors.CombineErrors(err, conn.Close(ctx))
 	}
 	if testingKnobs.FailedWriteToBucket.FailedAfterReadFromPipe {
 		return nil, errors.New(DirectCopyWriterMockErrMsg)
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -212,10 +212,6 @@ func Fetch(
 	defer func() {
 		if retErr != nil {
 			fetchID := fetchStatus.ID.String()
-			if !IsClearContinuationTokenMode(cfg) && cfg.FetchID != "" {
-				fetchID = cfg.FetchID
-			}
-
 			logger.Info().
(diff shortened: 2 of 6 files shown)


fetch/datablobstorage/copy_direct.go
```diff
@@ -50,7 +51,7 @@ func (c *copyCRDBDirect) CreateFromReader(
 		// In direct copy mode, we cannot continue, so might as well just make a new entry each time.
 		// Default to not clearing the continuation tokens and pass in a nil exception log so it
 		// creates the new entry.
-		pgErr := status.MaybeReportException(ctx, c.logger, conn, table.Name, err, "" /* fileName */, status.StageDataLoad, false /* isClearContinuationTokenMode */, nil /* exceptionLog */)
+		pgErr := status.MaybeReportException(ctx, c.logger, conn, table.Name, err, "" /* fileName */, status.StageDataLoad, false /* isClearContinuationTokenMode */, nil /* exceptionLog */, &sync.Mutex{})
```
ryanluu12345, line 54: The more I think about this, we don't need exception logging for `direct-copy` since resumability can't work since we don't export files. If we use this just to write exception logs to the database, that's duplicated because we already log out errors in direct copy mode anyways....
@Jeremyyang920 I'm thinking we can just remove this `MaybeReportException` for copy_direct mode.

Jeremyyang920: Yup. Makes sense to me. TBH I really dont know why anyone would want to use this mode anyways unless its super small.

Jeremyyang920 (approved):

fetch/fetch.go
```diff
@@ -648,13 +644,6 @@ func IsImportCopyOnlyMode(cfg Config) bool {
...
-	trimmedFetchID := strings.TrimSpace(cfg.FetchID)
-	trimmedContToken := strings.TrimSpace(cfg.ContinuationToken)
-
-	// First condition: fresh fetch run without continuation; second condition: fetch run with continuation
-	// but of all tokens with a fetch ID, which means we want to update existing tokens.
-	if trimmedFetchID == "" || (trimmedFetchID != "" && trimmedContToken == "") {
-		return true
-	}
-	return false
+	// Condition: fresh fetch run without continuation.
```
ZhouXing19, line 647: nit: isn't this the same as `IsImportCopyOnlyMode()`?

fetch/import_table.go
```diff
@@ -183,7 +183,8 @@ func importTable(
 		file, err := importWithBisect(ctx, kvOptions, table, logger, conn, locBatch)
 		if err != nil {
 			fileName := status.ExtractFileNameFromErr(file)
-			pgErr := status.MaybeReportException(ctx, logger, baseConn.(*dbconn.PGConn).Conn, table.Name, err, fileName, status.StageDataLoad, isClearContinuationTokenMode, exceptionLog)
+			pgErr := status.MaybeReportException(ctx, logger, baseConn.(*dbconn.PGConn).Conn, table.Name, err, fileName,
```
ZhouXing19, line 186: super tiny nit: i wonder if we should use a cloned conn rather than the same conn as the import job. If the conn for import job dies, the exception won't be able to be logged. I.e. it might be safer if we have a dedicated conn for exception logging.

ZhouXing19 (commented):
LGTM just some nits, not urgent at all.

fetch/fetch.go
Jeremyyang920: Its the inverse of it, but I think from a readability sense for this case it makes sense not to use !IsImporttCopyOnlyMode and be more explicit with the naming when reading the cases since it can become tricky sometimes understanding the negations. 

Not opposed to either, but whatever is easier for us to understand when having to reread things.

ryanluu12345: Agreed with Jeremy, it is the inverse. I prefer keeping this because `ClearContinuationMode` is not always guaranteed to be the inverse (refer to previous revision of this code). I want to make it explicitly clear that we have a clear continuation mode, and right now it happens to be the inverse of `IsImportCopyOnlyMode`.

fetch/import_table.go
ryanluu12345: Hmm if the connection dies during this import, will it return an error that the connection died? And does it map to a certain file that it failed on? I'm trying to wonder if there is even anything that can be logged in this case

PR #171, opened 2024-03-05
fetch: Add pagination for AWS ListObjects

Jeremyyang920:
This commit adds pagination support for ListFromContinuationPoint for the AWS case. Previously we would cap out at 1000 items. But, if the item you wanted was after 1000, it wouldnt return.

Release note: None

### DIFF
--- a/fetch/datablobstorage/gcp.go
+++ b/fetch/datablobstorage/gcp.go
@@ -11,7 +11,6 @@ import (
 	"github.com/cockroachdb/molt/dbtable"
 	"github.com/cockroachdb/molt/testutils"
 	"github.com/cockroachdb/molt/utils"
-	"github.com/googleapis/google-cloud-go-testing/storage/stiface"
 	"github.com/rs/zerolog"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/iterator"
@@ -128,11 +127,11 @@ func (s *gcpStore) ListFromContinuationPoint(
 	ctx context.Context, table dbtable.VerifiedTable, fileName string,
 ) ([]Resource, error) {
 	key, prefix := getKeyAndPrefix(fileName, s.bucketPath, table)
-	return listFromContinuationPointGCP(ctx, stiface.AdaptClient(s.client), key, prefix, s.bucket, s)
+	return listFromContinuationPointGCP(ctx, s.client, key, prefix, s.bucket, s)
 }
 
 func listFromContinuationPointGCP(
-	ctx context.Context, client stiface.Client, key, prefix, bucket string, gcpStore *gcpStore,
+	ctx context.Context, client *storage.Client, key, prefix, bucket string, gcpStore *gcpStore,
 ) ([]Resource, error) {
 	it := client.Bucket(bucket).Objects(ctx, &storage.Query{
 		Prefix: prefix,
@@ -144,6 +143,7 @@ func listFromContinuationPointGCP(
 	})
 
 	resources := []Resource{}
+	// GCP's iterator paginates by default so no need to handle here
 	for {
 		if attrs, err := it.Next(); err != nil {
 			if err == iterator.Done {
--- a/fetch/datablobstorage/gcp_integration_test.go
+++ b/fetch/datablobstorage/gcp_integration_test.go
@@ -1,0 +1,49 @@
+package datablobstorage
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"strings"
+	"testing"
+
+	"cloud.google.com/go/storage"
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/api/option"
+)
+
+func TestListFromContinuationPointGCP(t *testing.T) {
+	ctx := context.Background()
+	var sb strings.Builder
+	gcpClient, err := storage.NewClient(ctx,
+		option.WithEndpoint("http://localhost:4443/storage/v1/"),
+		option.WithoutAuthentication(),
+	)
+
+	require.NoError(t, err)
+
+	gcpStore := gcpStore{
+		bucket: "fetch-test",
+		logger: zerolog.New(&sb),
+	}
+
+	// Create the test bucket
+	err = gcpClient.Bucket("fetch-test").Create(ctx, "", nil)
+	require.NoError(t, err)
+
+	// Seed the initial data with 8 files
+	for i := 1; i <= 8; i++ {
+		o := gcpClient.Bucket("fetch-test").Object(fmt.Sprintf("public.inventory/part_0000000%d.tar.gz", i))
+		wc := o.NewWriter(ctx)
+		_, err = io.Copy(wc, bytes.NewReader([]byte("abcde")))
+		require.NoError(t, err)
+		require.NoError(t, wc.Close())
+	}
+
+	// List from file 4 which should result in files 4-8 inclusive
+	resources, err := listFromContinuationPointGCP(ctx, gcpClient, "public.inventory/part_00000004.tar.gz", "public.inventory", "fetch-test", &gcpStore)
+	require.NoError(t, err)
+	require.Equal(t, 5, len(resources))
+}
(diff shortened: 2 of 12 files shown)


fetch/datablobstorage/s3.go
```diff
@@ -217,7 +228,7 @@ func listFromContinuationPointAWS(
 		// fetch/public.inventory/part_00000005.tar.gz is >= to key meaning,
 		// it is a file we need to include.
 		g.Go(func() error {
-			objResp, err := s3Client.GetObject(&s3.GetObjectInput{
+			objResp, err := s3Client.HeadObject(&s3.HeadObjectInput{
```
ZhouXing19, line 231: nit: would it be safer to use `HeadObjectWithContext` to allow possible request cancellation?

fetch/datablobstorage/s3_integration_test.go
```diff
@@ -0,0 +1,96 @@
...
+			Key:      aws.String(fmt.Sprintf("public.inventory/part_%08d.tar.gz", i)),
+			Body:     bytes.NewReader([]byte("abcde")),
+			Bucket:   aws.String("fetch-test-paginate"),
+			Metadata: map[string]*string{numRowKeysAWS: aws.String("5")},
+		})
+		require.NoError(t, err)
+	}
+
+	// List from file 13 and ensure pagination worked as expected
+	resources, err := listFromContinuationPointAWS(ctx, s3Cli, "public.inventory/part_00000013.tar.gz", "public.inventory", &s3Store, 5)
```
ZhouXing19, line 93: out of curiosity, is the behavior before this commit is that in such a test, say we have 2000 files, and start from `part_00000001.tar.gz`, we would only get `len(resources) == 1000`? Like `1000` is the default (hardcoded) limit from aws side? And we chose `1000` as the "batch size" simply to match that default?

ZhouXing19 (approved):
Just some clarification questions

fetch/datablobstorage/s3.go
Jeremyyang920: Ah yes good catch. We should typically be using the withCtx versions of all the calls.

fetch/datablobstorage/s3_integration_test.go
Jeremyyang920: Yes 1000 is the hard page limit set by AWS. So before switching to the paginated ListObjects, we would just return the first 1000 items. But, if you were looking for a file say 1500, it would not be returned at all but also we would return an emtpy slice of resources because we filtered out all the objects returned for being less than 1500.

ZhouXing19: Cool, thanks for the ctx!

fetch/datablobstorage/gcp_integration_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"cloud.google.com/go/storage"
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/api/option"
+)
+
+func TestListFromContinuationPointGCP(t *testing.T) {
+	ctx := context.Background()
+	var sb strings.Builder
+	gcpClient, err := storage.NewClient(ctx,
```
ryanluu12345, line 20: Ahh interesting, since we are using the localstack backend, we can just run the GCP API commands to spin up a new bucket per test right?

fetch/datablobstorage/gcp_integration_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	require.NoError(t, err)
+
+	gcpStore := gcpStore{
+		bucket: "fetch-test",
+		logger: zerolog.New(&sb),
+	}
+
+	// Create the test bucket
+	err = gcpClient.Bucket("fetch-test").Create(ctx, "", nil)
```
ryanluu12345, line 33: Do we want to defer and clean up the buckets after the test finishes tho?

fetch/fetch.go
```diff
@@ -514,7 +514,7 @@ func fetchTable(
 					}
 				}()
 
-				r, err := importTable(ctx, cfg, targetConn, logger, table.VerifiedTable, e.Resources, testingKnobs, isLocal, isClearContinuationTokenMode, exceptionLog)
+				r, err := importTable(ctx, cfg, targetConn, logger, table.VerifiedTable, e.Resources, isLocal, isClearContinuationTokenMode, exceptionLog)
```
ryanluu12345, line 517: Why did we remove `testingKnobs` here?

fetch/datablobstorage/gcp.go
```diff
@@ -144,6 +143,7 @@ func listFromContinuationPointGCP(
 	})
 
 	resources := []Resource{}
+	// GCP's iterator paginates by default so no need to handle here
```
ryanluu12345, line 146: Oh interesting. So how do we get each successive page? The iterator already gives us each paginated item?

fetch/datablobstorage/s3.go
```diff
@@ -192,22 +192,29 @@ func (s *s3Store) ListFromContinuationPoint(
...
-		Prefix: aws.String(prefix),
-	})
-	if err != nil {
+	params := &s3.ListObjectsV2Input{
+		Bucket:  aws.String(s3Store.bucket),
+		Prefix:  aws.String(prefix),
+		MaxKeys: aws.Int64(maxKeys),
+	}
+	contents := make([]*s3.Object, 0)
+	if err := s3Client.ListObjectsV2PagesWithContext(ctx, params, func(page *s3.ListObjectsV2Output, lastPage bool) bool {
```
ryanluu12345, line 206: Note: we should probably add a test here that tests when there are > 1000 files we list from.

fetch/datablobstorage/s3_integration_test.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// Create the test bucket
+	_, err = s3Cli.CreateBucketWithContext(ctx, &s3.CreateBucketInput{
+		Bucket: aws.String("fetch-test-paginate"),
+	})
+	require.NoError(t, err)
+
+	// Seed the initial data with 20 files
+	for i := 1; i <= 20; i++ {
```
ryanluu12345, line 82: Hmm why 20? Why not try with the 1000 file limit?

fetch/fetch.go
Jeremyyang920: It was flagged as unused by the linter. Which was true if you trace into the function.

fetch/datablobstorage/gcp_integration_test.go
Jeremyyang920: For these tests yes, you can just create a new bucket for each test.

fetch/datablobstorage/gcp.go
Jeremyyang920: We don't need to do it. The itterator does it when it calls it.Next

fetch/datablobstorage/s3.go
Jeremyyang920: Don't need to. I injected a maxKeys into the listObjects input so we dont need to create 1000 items. It also broke localstack a little bit when we were making that many api calls so quickly in the docker container.

The test injects a size of 5, with 20 files total.

fetch/datablobstorage/s3_integration_test.go
Jeremyyang920: See comment above.

fetch/datablobstorage/gcp.go
ryanluu12345: Ok sounds good

fetch/fetch.go
ryanluu12345: Yep makes sense.

fetch/datablobstorage/gcp_integration_test.go
ryanluu12345: Yeah I think for the local dev / testing case. Especially when we have multiple tests coming up. Would be good to invest in teardown for this later.

fetch/datablobstorage/s3_integration_test.go
ryanluu12345: Got it.

PR #172, opened 2024-03-05
fetch/schema-creation: incorporate SCT type mapping

ZhouXing19:
This PR 
- Introduced the type translation from SCT ([link to managed-service](https://github.com/cockroachlabs/managed-service/blob/762ff4b552edbb5953f302d39640a4a76c84aba2/migration/internal/mysql/types.go)) for mysql. This accommodate types, with better precision.
- Moved create table stmt tests to datadriven. 
- Updated existing tests.

Release note: None


fetch/testdata/mysql/create_schema.ddt
```diff
@@ -0,0 +1,349 @@
...
+CREATE TABLE posts (tenant_id INT4, post_id INT4 NOT NULL PRIMARY KEY, author_id INT4, number INT4)
+------ DROPPED CONSTRAINTS ------
+UNIQUE KEY `author_id` (`author_id`)
+UNIQUE KEY `number` (`number`)
+CONSTRAINT `posts_ibfk_1` FOREIGN KEY (`tenant_id`) REFERENCES `tenants` (`tenant_id`) ON DELETE RESTRICT
+CONSTRAINT `posts_ibfk_2` FOREIGN KEY (`tenant_id`, `author_id`) REFERENCES `users` (`tenant_id`, `user_id`) ON DELETE SET NULL
+CONSTRAINT `posts_chk_1` CHECK ((`number` > 10))
+
+
+create-schema-stmt table_with_column_indexes show-dropped-constraints=true
```
ZhouXing19, line 100: Test from here and below are all from https://github.com/cockroachlabs/managed-service/blob/762ff4b552edbb5953f302d39640a4a76c84aba2/migration/internal/mysql/testdata/datadriven/create_table

e2e/testdata/mysql/schema_creation.ddt
```diff
@@ -32,7 +32,12 @@ fetch --source 'mysql://user:password@0.0.0.0:3306/defaultdb' --target 'postgres
...
+{"level":"warn","message":"mysql type char for column test_table.char_col: short_description:\"collate utf8mb4_0900_ai_ci not supported\" message:\"collate utf8mb4_0900_ai_ci was not added\""}
+{"level":"warn","message":"mysql type varchar for column test_table.varchar_col: short_description:\"collate utf8mb4_0900_ai_ci not supported\" message:\"collate utf8mb4_0900_ai_ci was not added\""}
+{"level":"warn","message":"mysql type text for column test_table.text_col: short_description:\"collate utf8mb4_0900_ai_ci not supported\" message:\"collate utf8mb4_0900_ai_ci was not added\""}
+{"level":"warn","message":"mysql type mediumtext for column test_table.mediumtext_col: short_description:\"collate utf8mb4_0900_ai_ci not supported\" message:\"collate utf8mb4_0900_ai_ci was not added\""}
+{"level":"warn","message":"mysql type longtext for column test_table.longtext_col: short_description:\"collate utf8mb4_0900_ai_ci not supported\" message:\"collate utf8mb4_0900_ai_ci was not added\""}
```
Jeremyyang920, line 39: Curious why this is showing up since your mysql schema doesnt provide the collation on the column on creation.

utils/typeconv/mysql_conv.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
```
Jeremyyang920, line 1: Hmm. I see what you are trying to do here, but I don't know how transferable this proto really is, but I'm also not sure we need to be bringing in a protobuf for this. Sine things like blocking, issue_num, short_description, or dismissed don't exactly make sense for the fetch context.

ZhouXing19: I think `blocking` and `short_description` might make sense, for some conversion error, such as `TINYINT` is not supported, is not a blocking error -- it's more like a notice to the user that we auto-promoted this type to INT2 in CockroachDB. But others yes, it might not be necessary. If preferred, i can make a more light-weight error struct without introducing protbuf here.

ZhouXing19: Done removing the protobuf and added a TypeConvError struct.

Jeremyyang920: I think you might have to push the changes haha. Unless github is having issues.

ZhouXing19: Ah dang forgot i switched to another branch 😅 just pushed!

ZhouXing19:
RFAL, thanks!

utils/typeconv/shortdesc.go
```diff
@@ -0,0 +1,52 @@
...
+	UnsupportedOrganizationClause      ShortDesc = "Unsupported ORGANIZATION"
+	UnsupportedPartition               ShortDesc = "Unsupported PARTITION clause"
+	UnsupportedPrecision               ShortDesc = "Unsupported precision"
+	UnsupportedRowGUIDCOL              ShortDesc = "ROWGUIDCOL not supported"
+	UnsupportedSequenceOption          ShortDesc = "Unsupported sequence option"
+	UnsupportedTableConnection         ShortDesc = "Unsupported CONNECTION constraint"
+	UnsupportedTablePartitioning       ShortDesc = "Unsupported table partitioning"
+	UnsupportedTablePhysicalProperties ShortDesc = "Unsupported table physical properties"
+	UnsupportedTinyInt                 ShortDesc = "TINYINT not supported"
+	Varchar2BytesLimit                 ShortDesc = "VARCHAR2 with BYTES limit"
```
Jeremyyang920, line 42: Im wondering if we need all of these copied over. Some of these were because in the SCT we have to deal with all the DDL's that we can possible see, but since the whay we create schema here is by checking the oid's on the MySQL side, we really only have the unsupported column type and collation errors from what I saw in the tests right?

fetch/schema_creation.go
```diff
@@ -95,7 +101,27 @@ func (t *columnWithType) CRDBColDef(includePk bool) (*tree.ColumnTableDef, error
...
+			for _, sc := range scs {
+				if sc.Blocking {
+					err = errors.Newf(
+						"failed to get crdb type from mysql type %s for column %s.%s: %s",
+						t.dataType,
+						t.tableName,
+						t.columnName,
+						sc.ShortDescription,
+					)
+					logger.Err(err)
```
Jeremyyang920, line 119: Was this error logged anywhere? Was trying to see if there was a test log with the short description, but all I saw were the normal messages.

ZhouXing19: Can be found in `schema_creation.ddt`:
```
create-schema-stmt test_table_set_col
CREATE TABLE test_table_set_col (
    integer_col INT PRIMARY KEY,
    set_col SET('value1', 'value2', 'value3')
);
----
failed to get crdb type from mysql type set for column test_table_set_col.set_col: Unsupported column type set

```

utils/typeconv/shortdesc.go
ZhouXing19: Removed not used errors

PR #173, opened 2024-03-05
fetch: use a cloned connection for exceptions log reporting

ryanluu12345:
Previously, Jane highlighted that if a connection dies while processing a copy or import, the exception log cannot be successfully written if the same conneciton is used. In order to avoid this case, we create a clone of this connection before we use it to record the exception.

Release Note: None

### DIFF
--- a/fetch/copy.go
+++ b/fetch/copy.go
@@ -30,6 +30,16 @@ func Copy(
 	isClearContinuationTokenMode bool,
 	exceptionLog *status.ExceptionLog,
 ) (CopyResult, error) {
+	exceptionConn, err := baseConn.Clone(ctx)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


ryanluu12345:
@ZhouXing19 are there any perf implications of creating the connection each time we hit `Copy` or `Import`?

ZhouXing19 (approved):

ZhouXing19:
> @ZhouXing19 are there any perf implications of creating the connection each time we hit Copy or Import?

Oh i missed this comment. 

I think the overhead should be minimal -- the # of new connections being made are just `cfg.Concurrency`, which is default to 4. Also, per the comment, seems that we will adopt connection pool for new conn creation, and the overhead would be further reduced.

PR #175, opened 2024-03-07
wip

Jeremyyang920:
Release note: None


fetch/fetch_test.go
```diff
@@ -237,6 +237,7 @@ func TestDataDriven(t *testing.T) {
 								ContinuationFileName: overrideFile,
 								FlushRows:            flushRows,
 								NonInteractive:       true,
+								Shards:               1,
```
Jeremyyang920, line 240: Deciding to just use 1 shard for the basic tests since its easier to maintain the ddt's and not having to remember which shard needs to go for what input. For e2e tests can use more shards for sure.

Jeremyyang920:
@ZhouXing19 This should be ready for an initial pass. Sorry its alot of files, but most of it is from test rewrites.

ZhouXing19 (approved):
This will be a big win!! I played it locally with different setups as well and it seems solid to me :D 

Just some nit additions that can be done in follow-up PRs, mostly for tests:
1. Test import continuation from shard > 1.
2. Test ingested error for export of a particular shard. Expected behavior: 
    a. Immediately cancel ctx of all goroutines for other shards;
    b. Clean up all exported csv files.
3. Super nit, maybe I'm paranoid: ensure that test for `ListFromContinuationPoint()` check the returned resources are in correct order (right now we only check the number of files is correct).
4. As noted in the TODO, investigate the possibility / performance change if use `COPY TO` for pg source.
5. Log in debug message the txn snap id for PG (and equivalent for mysql and crdb source)

@Jeremyyang920 do you want me to work on these ^?

ZhouXing19:
LGTM!!

PR #176, opened 2024-03-11
fetch/schema-creartion: distinguish nested arr type

ZhouXing19:
CRDB doesn't not support [nested array](https://www.postgresql.org/docs/current/arrays.html#ARRAYS), so we need to distinguish this case can emit error.

Release note: show error if the original column is a pg nested array.

### DIFF
--- a/fetch/fetch_test.go
+++ b/fetch/fetch_test.go
@@ -127,9 +127,15 @@ func TestDataDriven(t *testing.T) {
 							require.NoError(t, err)
 
 							for _, missingTable := range missingTables {
-								stmt, err := GetCreateTableStmt(ctx, logger, conns[0], missingTable.DBTable)
+								srcConn := conns[0]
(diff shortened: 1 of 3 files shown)


fetch/fetch_test.go
```diff
@@ -72,6 +72,12 @@ func TestDataDriven(t *testing.T) {
 				conns[1], err = dbconn.TestOnlyCleanDatabase(ctx, "target", tc.dest, dbName)
 				require.NoError(t, err)
 
+				defer func() {
+					for _, conn := range conns {
+						require.NoError(t, conn.Close(ctx))
+					}
+				}()
+
```
Jeremyyang920, line 80: Im not sure how I feel about potentially failing our test because a conn failed to close. Since at the end of a fetch run, connections would just die/close anyways so not sure if this is a big deal to check.

Jeremyyang920 (approved):

fetch/fetch_test.go
ZhouXing19: removed

PR #178, opened 2024-03-13
ingest testing knobs

ZhouXing19:


fetch/fetch_test.go
```diff
@@ -174,6 +197,15 @@ func TestDataDriven(t *testing.T) {
 						bucketPath := ""
 						sDetails := storeDetails{}
 						numShards := 1
+						elapsedComp := elapsedComparison{}
+						var fetchStartTime time.Time
+						var fetchFinishTime time.Time
+						var failedShardIdx *int
+						var failedIterNum *int
+						var failedRowCnt *int
```
Jeremyyang920, line 205: Do these 3 have to be pointers?

fetch/fetch_test.go
```diff
@@ -321,6 +434,18 @@ func TestDataDriven(t *testing.T) {
...
+							actualElapsed := fetchFinishTime.Sub(fetchStartTime)
+							if elapsedComp.comp == longer {
+								require.GreaterOrEqual(t, actualElapsed, elapsedComp.elapsed)
+							} else if elapsedComp.comp == shorter {
+								require.LessOrEqual(t, actualElapsed, elapsedComp.elapsed)
+							} else {
+								t.Fatalf("elapsed duration comparison must be > or <")
+							}
+						}
+
```
Jeremyyang920, line 448: Are we worried at all here that if CI for some reason is slow that the conditions set in the test below fail but the rest of the condition is fine? Since I know down below there is one with a <2s comparison.

PR #179, opened 2024-03-13
fetch/export: fix bug when source connection failed to establish

ZhouXing19:
Previously, when the connection for the datasource failed to be established during export, the sql writer is not closed, causing the process hangs at `record, err := r.Read()`, as it is reading from a dead but not closed pipe. This PR is to fix it, with test added. 

Release note (bug fix): during export, if the source connection fails to establish, without closing the sqlwriter will cause the whole program to hang. We fixed it by properly closing the sqlwriter. 

jira: [CC-27433](https://cockroachlabs.atlassian.net/browse/CC-27433)


Jeremyyang920 (approved):
Great Catch!

PR #181, opened 2024-03-13
[CC-27236] fetch: Copy Edits

Jeremyyang920:
This commit adjusts the copy for some of the flag names based on a meeting that we had.

Release notes: Live,concurrency, and shards flag have been renamed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,12 +153,12 @@ It currently supports the following:
 - Pulling a table, uploading CSVs to S3/GCP/local machine and running COPY FROM on Cockroach from that CSV.
 - Pulling a table and running COPY FROM directly onto the CRDB table without an intermediate store.
 
-By default, data is imported using `IMPORT INTO`. You can use `--live` if you
+By default, data is imported using `IMPORT INTO`. You can use `--use-copy` if you
 need target data to be queriable during loading, which uses `COPY FROM` instead.
 
 Data can be truncated automatically if run with `--table-handling 'truncate-if-exists'`. Molt Fetch can also automatically create the new table on the target side if run with `--table-handling 'drop-on-target-and-recreate'`. The user can also manually create the new table schema on the target side, and run with `--table-handling 'none'` (which is the default setting of table handling options).
 
-A PG replication slot can be created for you if you use `pg-logical-replication-slot-name`,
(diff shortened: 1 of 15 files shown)


cmd/fetch/fetch.go
```diff
@@ -245,16 +245,16 @@ func Command() *cobra.Command {
...
 	cmd.PersistentFlags().IntVar(
-		&cfg.Concurrency,
-		"concurrency",
+		&cfg.TableConcurrency,
+		"table-concurrency",
 		4,
 		"Number of tables to move at a time.",
 	)
 	cmd.PersistentFlags().IntVar(
 		&cfg.Shards,
```
Jeremyyang920, line 254: Internally leaving the config name as shards since it makes sense from and engineering pov, and is what we still use on the verify side. Can rename later if we want.

cmd/fetch/fetch.go
```diff
@@ -226,10 +226,10 @@ func Command() *cobra.Command {
 		"Enables direct copy mode, which copies data directly from source to target without using an intermediate store.",
 	)
 	cmd.PersistentFlags().BoolVar(
-		&cfg.Live,
-		"live",
+		&cfg.UseCopy,
+		"use-copy",
 		false,
-		"Whether the table must be queryable during load import.",
+		"Use `COPY FROM` instead of `Import INTO` during the data movement phase. This keeps your table online during the process.",
```
taroface, line 232:
```suggestion
		"Use `COPY FROM` instead of `IMPORT INTO` during the data movement phase. This keeps your table online during the process.",
```

taroface (approved):
LGTM, just a copy tweak!

ZhouXing19 (approved):

README.md
```diff
@@ -153,12 +153,12 @@ It currently supports the following:
 - Pulling a table, uploading CSVs to S3/GCP/local machine and running COPY FROM on Cockroach from that CSV.
 - Pulling a table and running COPY FROM directly onto the CRDB table without an intermediate store.
 
-By default, data is imported using `IMPORT INTO`. You can use `--live` if you
+By default, data is imported using `IMPORT INTO`. You can use `--use-copy` if you
```
ryanluu12345, line 156: Should this also specify that while we use an intermediate store? Or is that implied?

PR #183, opened 2024-03-14
schema-creation: support precision for datetime and float types for pg

ZhouXing19:
Previously, we don't consider the precision for datetime and float types if the source db is PostgreSQL. This PR is to include it.

We also cleaned up the output for logging to removed the quote signs.

Release note: None


fetch/schema_creation_test.go
```diff
@@ -52,45 +53,50 @@ func TestGetColumnTypes(t *testing.T) {
...
 					"name": {
 						dataType: "character varying(50)",
-						typeOid:  oid.T_varchar,
+						pgMeta:   &typeconv.PGColumnMeta{TypeOid: oid.T_varchar},
 					},
 					"created_at": {
 						dataType: "timestamp with time zone",
-						typeOid:  oid.T_timestamptz,
+						pgMeta:   &typeconv.PGColumnMeta{TypeOid: oid.T_timestamptz},
+
```
Jeremyyang920, line 66: nit: lets remove the extra blank lines from the jsons.

fetch/schema_creation_test.go
```diff
@@ -118,46 +124,50 @@ func TestGetColumnTypes(t *testing.T) {
...
 						dataType: "timestamp with time zone",
-						typeOid:  oid.T_timestamptz,
+						pgMeta:   &typeconv.PGColumnMeta{TypeOid: oid.T_timestamptz},
 						pkPos:    3,
 					},
 					"is_hired": {
 						dataType: "boolean",
-						typeOid:  oid.T_bool,
+						pgMeta:   &typeconv.PGColumnMeta{TypeOid: oid.T_bool},
+
```
Jeremyyang920, line 144: same here

utils/typeconv/pg.go
```diff
@@ -0,0 +1,175 @@
...
+			tfs = append(tfs, &TypeConvError{
+				ShortDescription: InvalidDecimalArgs,
+				Message:          fmt.Sprintf("cockroachDB does not accept negative scale value %d for column %s, auto correcting scale to 0", scale, colName),
+			})
+			return types.MakeDecimal(int32(precision), 0), tfs
+		}
+		if scale > precision {
+			tfs = append(tfs, &TypeConvError{
+				ShortDescription: InvalidDecimalArgs,
+				Message:          fmt.Sprintf("precision %d cannot be bigger than cm.NumScale %d for column %s", cm.NumPrecision, cm.NumScale, colName),
```
Jeremyyang920, line 135:
```suggestion
				Message:          fmt.Sprintf("precision %d cannot be bigger than scale %d for column %s", cm.NumPrecision, cm.NumScale, colName),
```

fetch/schema_creation.go
```diff
@@ -126,46 +127,48 @@ func (t *columnWithType) CRDBColDef(
...
+		for _, sc := range scs {
+			if sc.Blocking {
+				err = errors.Newf(
+					"failed to get crdb type from type [%s] for column %s: %s",
+					t.dataType,
+					t.Name(),
+					sc.ShortDescription,
+				)
+				logger.Err(err)
+				return nil, err
```
Jeremyyang920, line 169: If we return nil,error here, we don't see all possible errors. We need to log everything out then return error.

Jeremyyang920 (approved):
Changes LGTM, just the minor small things.

fetch/schema_creation.go
ZhouXing19: Good call, fixed.

fetch/schema_creation.go
```diff
@@ -126,48 +128,52 @@ func (t *columnWithType) CRDBColDef(
...
+					t.dataType,
+					t.Name(),
+					sc.ShortDescription,
+				))
+			}
+			logger.Warn().Msgf("type [%s] for column %s: %s", t.dataType, t.Name(), sc.ShortDescription)
+		}
+		if err != nil {
+			log.Err(err)
+			return nil, err
```
Jeremyyang920, line 175: I think at this point, we would just log a generic error like exiting due to XYZ, since we already logged all the ones above. This err would be the same as the last logged err in the loop.

PR #186, opened 2024-03-19
[CC-27525] fetch: Set copy session vars

Jeremyyang920:
This commit sets the 2 copy session variables required for larger copy operations. These are the same ones we currently have to set for DMS to succeed.

Release Note: None

### DIFF
--- a/fetch/copy.go
+++ b/fetch/copy.go
@@ -41,13 +41,15 @@ func Copy(
 	}()
 
 	dataLogger := moltlogger.GetDataLogger(logger)
-	ret := CopyResult{
-		StartTime: time.Now(),
-	}
+	var ret CopyResult
 
 	rowsSoFar := 0
 	conn := baseConn.(*dbconn.PGConn).Conn
-
+	// Set the session variables required for COPY
+	if err := datablobstorage.SetCopyEnvVars(ctx, conn); err != nil {
+		return ret, err
+	}
+	ret.StartTime = time.Now()
 	for i, resource := range resources {
 		key, err := resource.Key()
 		if err != nil {
--- a/fetch/datablobstorage/datablobstorage.go
+++ b/fetch/datablobstorage/datablobstorage.go
@@ -159,3 +159,13 @@ func GenerateDatastore(
 
 	return src, err
 }
+
+func SetCopyEnvVars(ctx context.Context, conn *pgx.Conn) error {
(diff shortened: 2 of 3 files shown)


fetch/datablobstorage/datablobstorage.go
```diff
@@ -159,3 +159,13 @@ func GenerateDatastore(
 
 	return src, err
 }
+
+func SetCopyEnvVars(ctx context.Context, conn *pgx.Conn) error {
```
ryanluu12345, line 163: Can these statements run in a transaction? I'm also wondering if we can run into weird behavior if one is set but not the other.

fetch/copy.go
```diff
@@ -41,13 +41,15 @@ func Copy(
...
-	ret := CopyResult{
-		StartTime: time.Now(),
-	}
+	var ret CopyResult
 
 	rowsSoFar := 0
 	conn := baseConn.(*dbconn.PGConn).Conn
-
+	// Set the session variables required for COPY
+	if err := datablobstorage.SetCopyEnvVars(ctx, conn); err != nil {
```
ryanluu12345, line 49: Do we have any cleanup we must do of these env vars after the COPY completes?

ryanluu12345 (approved):
LGTM! Just a couple clarifying questions.

fetch/copy.go
Jeremyyang920: Hmm its a good point. We should probably reset it, but it doesn't really affect anything if we don't.

fetch/datablobstorage/datablobstorage.go
Jeremyyang920: Sure we can be explicit, but theres no real weird behavior if one is set and the other isint.

fetch/copy.go
ryanluu12345: Ok, this might be a good change to do. Can defer the cleanup

fetch/datablobstorage/datablobstorage.go
ryanluu12345: No need to do this then. Sounds good

fetch/copy.go
Jeremyyang920: Actually sorry this doesnt matter. Its a session variable not cluster level. So we dont need to reset it.

Issue #187, opened 2024-03-21
verify: dropped column causes error exit


abtreece:
I have been running `molt verify` against a database I am migrating and am getting the following when comparing the source and target...

```
{"level":"info","time":"2024-03-21T15:05:22-05:00","message":"verification in progress"}
Error: error verifying: error loading type anyelement: unknown typtype
Usage:
  molt verify [flags]

Flags:
      --concurrency int                          Number of tables to process at a time (defaults to number of CPUs).
      --continuous                               Whether verification should continuously run on each shard.
      --continuous-pause-between-runs duration   Amount of time to pause between continuous runs (e.g. 1h, 2m).
  -h, --help                                     Help for verify command.
      --live                                     Enable live mode, which attempts to account for rows that can change in value by retrying them before marking them as inconsistent.
      --live-flush-interval duration             Maximum amount of time to wait before retrying rows (live mode only). (default 1s)
      --live-max-batch-size int                  Maximum number of rows to retry at a time (live mode only). (default 1000)
      --live-retries-max-iterations int          Maximum number of retries before marking rows as inconsistent (live mode only). (default 5)
      --live-retry-initial-backoff duration      Amount of time live verification should initially backoff for before retrying. (default 250ms)
      --live-retry-max-backoff duration          Maximum amount of time a retry attempt should take before retrying again (live mode only). (default 1s)
      --live-retry-multiplier int                Multiplier to apply to backoff duration after each failed row verification (live mode only). (default 2)
      --live-runs-per-second int                 Maximum number of retry attempts per second (live mode only).
      --log-file string                          If set, writes to the log file specified. Otherwise, only writes to stdout.
      --logging string                           Level to log at (maps to zerolog.Level). (default "info")
      --metrics-listen-addr string               Address for the metrics endpoint to listen to. (default "127.0.0.1:3030")
      --row-batch-size int                       Number of source/target rows to scan at a time. (default 20000)
      --rows                                     If true, verify both the schema (columns, types) and row data. If false, verify only the schema. (default true)
      --rows-per-second int                      If set, maximum number of rows to read per second on each shard.
      --schema-filter string                     POSIX regexp filter for schemas to action on. (default ".*")
      --source string                            Connection string of the source database.
      --table-filter string                      POSIX regexp filter for tables to action on. (default ".*")
      --target string                            Connection string of the target database.
      --use-console-writer                       Use the console writer, which has cleaner log output but introduces more latency (defaults to false, which logs as structured JSON).

[...]

PR #189, opened 2024-03-21
verfiy: ignore dropped columns

Jeremyyang920:
This commit adds a attisdropped clause for pg/crdb queries when getting columnns to verify. Previously, this would return and when running LoadDataType in the PGX driver, it would cause an issue as the dropped column could not be resolved. We now omit dropped columns from being returned since we are not verifying them anyways.

Release Note: Fixes a bug for pg/crdb verify that causes an error when there are dropped columns on the table
Fixes: #187

### DIFF
--- a/verify/testdata/datadriven/crdb/dropped_column.ddt
+++ b/verify/testdata/datadriven/crdb/dropped_column.ddt
@@ -1,0 +1,24 @@
+# Test that a dropped column can result in a successful fetch
+
+exec all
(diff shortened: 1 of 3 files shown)


verify/testdata/datadriven/crdb/dropped_column.ddt
```diff
@@ -0,0 +1,24 @@
+# Test that a dropped column can result in a successful fetch still
```
ajstorm, line 1: Nit: you can drop the "still" at the end of this line.

ajstorm (approved):
Nice find!  LGTM

PR #190, opened 2024-03-27
cmd: support --version flag

ZhouXing19:
Release note: None

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -35,7 +35,7 @@ func init() {
 	rootCmd.AddCommand(EscapePasswordCommand())
 	rootCmd.AddCommand(verify.Command())
 	rootCmd.AddCommand(fetch.Command())
-
+	rootCmd.Version = fmt.Sprintf("v%s", moltVersion)
 	// Hide completion options, because irrelevant.
 	rootCmd.CompletionOptions.HiddenDefaultCmd = true
 
--- a/e2e/testdata/pg/simple_fetch_verify.ddt
+++ b/e2e/testdata/pg/simple_fetch_verify.ddt
@@ -36,7 +36,6 @@ id | t
 ----+---
 (0 rows)
 
-
 fetch --source 'postgres://postgres@localhost:5432/defaultdb' --target 'postgres://root@localhost:26257/defaultdb?sslmode=disable' --table-filter 'tbl1' --local-path /tmp/basic --local-path-listen-addr '0.0.0.0:9005' --non-interactive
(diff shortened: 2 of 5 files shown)


ryanluu12345 (approved):
Overall this looks good to me. I like the implementation. However, what is your plan on updating this for releases? Will you be doing something similar to what we do in `crdb-proxy` and automating the version bump and then creating a PR?

You can do this via:
https://github.com/cockroachlabs/crdb-proxy/blob/main/.github/workflows/release.yaml#L171-L206
https://github.com/cockroachlabs/crdb-proxy/blob/main/.github/workflows/release.yaml#L43-L51

e2e/testdata/pg/simple_fetch_verify.ddt
```diff
@@ -36,6 +36,9 @@ id | t
 ----+---
 (0 rows)
 
+version
```
Jeremyyang920, line 39: I don't know if we want this in a test since this will fail after we update version each time.

cmd/root.go
```diff
@@ -10,6 +10,7 @@ import (
 )
 
 const rootCmdUse = "molt"
+const moltVersion = "v0.1.2"
```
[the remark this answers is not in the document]
ZhouXing19: Updated

e2e/testdata/pg/simple_fetch_verify.ddt
ZhouXing19: Removed.

Jeremyyang920:
Theres a typo on the genreated pr
"Bump to ersion-flag"

But not a big deal.

ZhouXing19:
> Theres a typo on the genreated pr
"Bump to ersion-flag"

it's because in the triggered run i didn't specify the version (seems its default value is `version-flag`). With the normal release flow with tag like `vxx.xx.xx` it should work correctly.

----

Updated the release workflow with test passed in : https://github.com/cockroachdb/molt/actions/runs/8454727832

Can i ask for another round of review on the release.yaml? Thanks!

e2e/testdata/pg/simple_fetch_verify.ddt
```diff
@@ -36,7 +36,6 @@ id | t
 ----+---
 (0 rows)
 
-
```
Jeremyyang920, line 39: Curious why this line got removed? But this shouldnt affect test output right?

Jeremyyang920 (approved):

PR #193, opened 2024-03-28
molt-nightly.sh: added long flag support and specific machine configuration options

ryanluu12345:
In order to support machine configuration options, implemented long flag support so that we are not constrained by single letter options. Additionally, updated the arguments to Roachprod so that they are now dynamic based on which ones are non-empty.

Added in better UX to the script by having a usage section and helpers that extract and check arguments.

Resolves: CC-27574
Release Note: None

Test here: https://teamcity.cockroachdb.com/buildConfiguration/Internal_Molt_MoltNightly/14605809?buildTab=log&focusLine=1917&logView=linear&linesState=116


build/teamcity/internal/molt/molt-nightly.sh
```diff
@@ -2,8 +2,25 @@
...
 
+VARIANT="base_aws"
+
 # Get this current directory.
 DIR="$(cd "$(dirname "$0")" && pwd)"
 
-# Run the setup script.
-$DIR/setup-scale-test.sh -n 4 -z us-east-1a
\ No newline at end of file
+# Add new case statements for new variants
```
ryanluu12345, line 10: @Jeremyyang920 this is the section of code that I think will help us to scale to each variant we need. We can key off different env variables and inputs here and just pass them into the `setup-scale-test` which is our main executor script.

This allows us to setup many different tests across different dimensions: scale, cloud provider, specific machine types, source databases, datasets....you name it

build/teamcity/internal/molt/setup-scale-test.sh
```diff
@@ -2,31 +2,120 @@
...
+    [[ ("$1" == *=* && -n ${1#*=}) || ( ! -z "$2" && "$2" != -*)  ]];
+}
+
+extract_argument() {
+  echo "${2:-${1#*=}}"
+}
+
 export ROACHPROD_USER=migrations
 # Read flags
 CLUSTER=migrations-nightly-final
```
Jeremyyang920, line 30: Just noticed this, but if we end up running things in parallel, our cluster names cant be the same so we may need to find a suffix to identify the case running.

build/teamcity/internal/molt/molt-nightly.sh
```diff
@@ -2,8 +2,25 @@
...
+case $VARIANT in
+  base_aws)
+    $DIR/setup-scale-test.sh -n 4 --cloud aws --zones us-east-1a
+    ;;
+
+  scaled_aws)
+    $DIR/setup-scale-test.sh -n 4 --cloud aws --volume-size 1500 --zones us-east-1a --aws-cpu-options 'CoreCount=8,ThreadsPerCore=2' --machine-type m6idn.4xlarge --aws-machine-type-ssd m6idn.4xlarge
+    ;;
```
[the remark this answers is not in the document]
ryanluu12345: Yep, this just illustrates the concept. This is meant to be a setting you as a test author can change to match your scale needs.

build/teamcity/internal/molt/setup-scale-test.sh
ryanluu12345: I think if we do run things in parallel it will be through the entry point of `molt-nightly.sh`. If you look below at the flags, I support a roachprod cluster name where you can dynamically set it. No need to add a prefix, can just add a unique name entirely.

Jeremyyang920: I see. Then should we still keep this around then? or just use the flag name and enforce it at the top level?

Jeremyyang920 (approved):

build/teamcity/internal/molt/setup-scale-test.sh
ryanluu12345: I like the idea of making this explicitly required. Let me make those changes.

PR #194, opened 2024-03-28
schema-creation: enable customized type mapping

ZhouXing19:
User now can specify a overriding type mapping via `--type-map-file` for `molt fetch` for schema creation. It determines the type mapping for a specific column of a specific table, and override the default mapping.

Release note: enable customized type mapping for schema creation with `--type-map-file`.


cmd/fetch/fetch.go
```diff
@@ -368,6 +368,42 @@ or as an automated job. If not set, prompts user for confirmation before perform
 		),
 	)
 
+	cmd.PersistentFlags().StringVar(&cfg.CustomizedTypeMapPath, "type-map-file", "",
```
ryanluu12345, line 371: This set of changes is very exciting! Feature parity with DMS!

fetch/schema_creation.go
```diff
@@ -42,8 +44,14 @@ func (cs columnsWithType) CRDBCreateTableStmt(logger zerolog.Logger) (string, er
 	// pk. If there are more than one pks, we need to create a pk constraint
 	// that group all the selected columns, thus result in different syntax.
 	includePkForEachCol := len(pkList) <= 1
+
 	for _, col := range cs {
-		colDef, err := col.CRDBColDef(includePkForEachCol, logger)
+		mappingRule := overridingTypeMap["*"]
```
ryanluu12345, line 49: Is there any way that the entry for `*` is invalid and we error or panic here?

fetch/schema_creation.go
```diff
@@ -102,27 +110,41 @@ func (t *columnWithType) CRDBColDef(
...
-						t.dataType,
-						t.tableName,
-						t.columnName,
-						sc.ShortDescription,
-					)
-					logger.Err(err)
-					return nil, err
+		// Prioritize the overriding rules for type mapping.
+		if overrideTypeMap != nil {
+			logger.Debug().Msgf("received overlapping map rule for column %s.%s: %s", t.tableName, t.columnName, overrideTypeMap.String())
```
ryanluu12345, line 115: I wonder if this is worthwhile to log out in `INFO` mode since this mapping logging only happens once on setup and contains essential information about what is being mapped.

utils/typeconv/testdata/mapping.ddt
```diff
@@ -0,0 +1,204 @@
+type-mapping
```
ryanluu12345, line 1: I like these datadriven tests because it shows exactly how the mapping will look in the Go struct itself.

utils/typeconv/pg.go
```diff
@@ -0,0 +1,74 @@
+package typeconv
```
ryanluu12345, line 1: I'm wondering why you have a `pg` specific file for these helpers. Does MySQL not need these equivalent methods? I'd imagine there'd be more discrepancy with scale/precision of certain types in another dialect.

.github/e2esampledata/manytypet.csv
```diff
@@ -0,0 +1,187 @@
+id,name,age,height,is_active,birthdate,registration_timestamp,profile_picture,bio,favorite_colors,credit_card_number,email_address,weight,last_login_ip,membership_status,review_ratings,expiry_date
```
Jeremyyang920, line 1: Is there a typo for this file name? manytypes.csv?

fetch/schema_creation.go
ZhouXing19: This should not error as in golang, a map retrieval yields a zero value when the key is not present.

See also https://go.dev/play/p/--RODteJKte

fetch/fetch.go
```diff
@@ -127,6 +128,9 @@ func Fetch(
 	}
 
 	if !cfg.DropAndRecreateNewSchema {
+		if cfg.CustomizedTypeMapPath != "" {
+			logger.Warn().Msgf("not in drop-on-target-and-recreate mode, ignoring customized type mapping")
```
Jeremyyang920, line 132: Hmm. Do we wan't to do this though? What if someone created the schema on their end manually but still want the type map.

fetch/schema_creation.go
```diff
@@ -102,27 +110,41 @@ func (t *columnWithType) CRDBColDef(
...
+			if !strings.EqualFold(t.dataType, overrideTypeMap.SourceType()) {
+				logger.Warn().Msgf("the original type of column %s is %s, while overriding map has source type %s, so igoring the overriding map", t.columnName, t.dataType, overrideTypeMap.SourceType())
+			} else {
+				colType = overrideTypeMap.CRDBType()
+				logger.Info().Msgf("use customized mapping for column %s.%s: %s -> %s", t.tableName, t.columnName, t.dataType, colType.SQLString())
+			}
+		}
+
+		if colType == nil {
+			if t.mysqlMeta == nil {
```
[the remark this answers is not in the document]
ZhouXing19: Yes I meant to do it in another PR so that this one won't be too large.

utils/typeconv/testdata/mapping.ddt
```diff
@@ -0,0 +1,204 @@
...
+        "crdb-type": "string"
+      }
+    ]
+  }
+]
+----
+cannot get the crdb type for "dota": cannot assert the type "dota" as *types.T on crdb
+
+
+type-mapping expect-error
```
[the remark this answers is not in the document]
ZhouXing19: Good point, added.

fetch/fetch.go
ZhouXing19: Oh, i thought the customized mapping is just for schema creation. But in what case would a user want to have a type map if not used for schema creation? For verify? We already allow-listed quite a few comparison between by-default-incomparable types, so without a type map, verify should works for those types.

verify/tableverify/tableverify.go
```diff
@@ -316,5 +316,29 @@ func allowableTypes(a oid.Oid, b oid.Oid) bool {
 	if a == oid.T_int2 && b == oid.T_bool {
 		return true
 	}
+	if a == oid.T_uuid && b == oid.T_varchar {
```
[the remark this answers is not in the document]
ZhouXing19: Added comparison in `verifyRows()` now. Also added test for `json <> text` comparison for verify.

ZhouXing19:
The majority of test of this PR can be found in `e2e/testdata/pg/type_override.ddt`. The type conversion for datums is also taken care of for verify. You may want to check the result of `verify ____` command in this file, they all showed `num_sucess=<all rows>` with no mismatches.

e2e/testdata/pg/mapping/empty.json
```diff
@@ -0,0 +1,4 @@
+[{"table": "manytypes",
```
Jeremyyang920, line 1: Is our json structure defined as array of json?
Also can we format the json files with like a prettyJson so the levels are indented correctly.

fetch/fetch.go
Jeremyyang920: I think im more so thinking of the case, they don't use schema creation and manually did create table which could still happen but they want the type mapping to exist still so we dont say mismatching column.

verify/tableverify/tableverify.go
```diff
@@ -316,5 +317,29 @@ func allowableTypes(a oid.Oid, b oid.Oid) bool {
...
+	}
+	if a == oid.T_text && b == oid.T_jsonb {
+		return true
+	}
+	if a == oid.T_inet && b == oid.T_text {
+		return true
+	}
+	if a == oid.T_text && b == oid.T_inet {
+		return true
+	}
```
Jeremyyang920, line 343: Does this mean that if we ever get a new type override that doesnt exist, we will have to add it here?

cmd/fetch/fetch.go
```diff
@@ -386,7 +386,13 @@ or as an automated job. If not set, prompts user for confirmation before perform
 		),
 	)
 
-	for _, hidden := range []string{testOnlyFlagStr, "allow-cockroach-replication"} {
+	cmd.PersistentFlags().StringVar(&cfg.CustomizedTypeMapPath, "type-map-file", "",
```
ryanluu12345, line 389: By any chance, do you have the contents of an example file you can include as markdown in the README? I think it's worth documenting this feature in the README.

fetch/fetch.go
ryanluu12345: Yeah, I agree with the use case Jeremy is pointing out. Imagine, they already used something like SCT or a schema tool and then manually drop constraints. Then they'll have source and target definitions that may or may not match 1:1, but they want to be able to treat it as the new types on their target.

ryanluu12345: Another consideration here: are we supporting dropped or ignored columns in the future? If for example the target no longer has columns the customer wants from the source.

fetch/fetch_test.go
```diff
@@ -128,7 +136,7 @@ func TestDataDriven(t *testing.T) {
 
 							for _, missingTable := range missingTables {
 								srcConn := conns[0]
-								stmt, err := GetCreateTableStmt(ctx, logger, srcConn, missingTable.DBTable)
+								stmt, err := GetCreateTableStmt(ctx, logger, srcConn, missingTable.DBTable, "")
```
ryanluu12345, line 139: Nit: comment what the `""` is.

pgconv/datum_converter.go
```diff
@@ -106,6 +108,16 @@ func ConvertRowValue(typMap *pgtype.Map, val any, typOID oid.Oid) (tree.Datum, e
 	case oid.T_timetz: // does not exist in pgtype.
 		d, _, err := tree.ParseDTimeTZ(parsectx.ParseContext, val.(string), time.Microsecond)
 		return d, err
+	case pgtype.InetOID:
```
ryanluu12345, line 111: Can we keep a running list somewhere of which mappings we support? Since people right now don't know what a source type can be mapped to. That way, @taroface can make documentation that documents it in a tabular format like: https://www.convert-in.com/docs/mss2pgs/types-mapping.htm

verify/tableverify/tableverify.go
ryanluu12345: Yes, this is what we had to do for STRING to UUID and INT to BOOL. We need to update this and `datum_converter` files too.

ryanluu12345: This part ungates a conversion, otherwise they won't be able to be converted using the datum converter.

ryanluu12345 (commented):
Overall, looks pretty solid with a couple clarification and nits. I think a lot of the follow up here is properly documenting the behavior.

e2e/testdata/pg/mapping/empty.json
ryanluu12345: You can install a formatter or have it run as a CLI invoked by `make`

fetch/schema_creation.go
```diff
@@ -102,27 +110,41 @@ func (t *columnWithType) CRDBColDef(
...
+				// If this is from a PG source.
+				colType = crdbtypes.OidToType[t.typeOid]
+			} else {
+				// If this is from a MySQL source.
+				colType, scs = t.mysqlMeta.ToDefaultCRDBType(t.dataType, t.columnName)
+				for _, sc := range scs {
+					if sc.Blocking {
```
ryanluu12345, line 132: Why do we have a `Blocking` field here? What does this tell us? Like a blocking translation error?

fetch/fetch.go
ZhouXing19: Do you mean only migrate a subset of columns of a table? We don't support that at this moment but DMS does. Can be a TODO

ryanluu12345: Yes, exactly that

cmd/fetch/fetch.go
ZhouXing19: Done, see `Customized Type Mapping` in the readme.

fetch/schema_creation.go
ZhouXing19: The `Blocking` field is for the severity of an error. A not-blocking error will just show in the log but not halt the whole fetch process. This is to follow [the convention of SCT](https://github.com/cockroachlabs/managed-service/blob/d89dff17e119f188d2b863cce0b4471450eb86f6/migration/internal/mysql/types.go#L39).

pgconv/datum_converter.go
ZhouXing19: Tracking in https://cockroachlabs.atlassian.net/browse/CC-28108

verify/tableverify/tableverify.go
ZhouXing19: Yes.

Jeremyyang920 (approved):
Overall the changes lgtm. Just a few minor nits.

utils/typeconv/map_override.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func getTypeFromName(typ string) (*types.T, error) {
+	stmt, err := parser.Parse(fmt.Sprintf("SELECT 1::%s", typ))
+	if err != nil {
+		return nil, err
+	}
+
+	ast := stmt[0].AST.(*tree.Select)
+	selectCaluse := ast.Select.(*tree.SelectClause)
```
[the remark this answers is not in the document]
ZhouXing19: Added assertion.

fetch/schema_creation.go
```diff
@@ -102,27 +113,41 @@ func (t *columnWithType) CRDBColDef(
...
-						t.columnName,
-						sc.ShortDescription,
-					)
-					logger.Err(err)
-					return nil, err
+		// Prioritize the overriding rules for type mapping.
+		if overrideTypeMap != nil {
+			logger.Info().Msgf("received overlapping map rule for column %s.%s: %s", t.tableName, t.columnName, overrideTypeMap.String())
+			if !strings.EqualFold(t.dataType, overrideTypeMap.SourceType()) {
+				logger.Warn().Msgf("the original type of column %s is %s, while overriding map has source type %s, so igoring the overriding map", t.columnName, t.dataType, overrideTypeMap.SourceType())
```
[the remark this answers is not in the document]
ZhouXing19: Hmm, I think we should only force the mapping if the original type matches the overriding rules. 
Say the original type for column `a` is bool, and the overriding rule is `a : int -> string`, I think we should ignore it.
It's also a simpler logic to apply `*: int -> string`. In this case, since `a` is not of `int` type, it won't be affected by this overriding rule.

PR #195, opened 2024-04-01
fetch: fix the empty string interpreted as null for not-null column

ZhouXing19:
Previously, even for column that is not null, the empty string will be exported to the csv as a non-quoted empty entry, resulting in this value being imported/copied as NULL into the target table, and lead to a `not null constraint violated` error.

This patch is to check if the target column is not null, and if the value for it is an empty string, it should be exported as a quoted empty string, so that we can differentiate it with the null case.

Release note (bug fix): fix the bug where empty string interpreted as null for not-null column.


ZhouXing19:
This is a (super) temporary solution for the escalation from the customer. It inspired that we should allow user-specified null for both export and import stage. 

- It would be nice if we have the `FORCE NULL` and `FORCE NOT NULL` implemented for copy:  https://github.com/cockroachdb/cockroach/issues/85575
- We should allow user to specify `nullif` for IMPORT INTO: https://www.cockroachlabs.com/docs/stable/import-into
- We should allow user to specify `null` for COPY FROM: https://www.cockroachlabs.com/docs/stable/copy-from

fetch/datablobstorage/copy_direct.go
```diff
@@ -52,7 +52,9 @@ func (c *copyCRDBDirect) CreateFromReader(
 	}
 
 	c.logger.Debug().Int("batch", iteration).Msgf("csv batch starting")
-	if _, err := conn.PgConn().CopyFrom(ctx, r, dataquery.CopyFrom(table, false /*skipHeader*/)); err != nil {
+	query := dataquery.CopyFrom(table, false /*skipHeader*/)
+	c.logger.Info().Msgf("runnign copy drom: %s", query)
```
Jeremyyang920, line 56:
```suggestion
	c.logger.Info().Msgf("runnign copy from: %s", query)
```
This shouldn't log anything sensitive out right? We redacted the one for import since i had credentials. But copy from should be safe right?

Jeremyyang920: We should also make this debug level to match the import level.

Jeremyyang920 (commented):
Approving for now so we can help the customer, but lets followup shortly with the todo fixes to clean it up.

fetch/datablobstorage/copy_direct.go
ZhouXing19: Yes, `COPY FROM` stmt only contains the table and column names so should be fine.

Jeremyyang920 (approved):

PR #199, opened 2024-04-01
fetch: Fix fetching empty string

ZhouXing19:
The previous fix in #195 is problematic. The empty string should be imported as literallly empty string to cockroachdb, while the fix in #195 imported it as a string with 2 single quotes. 

- In #195, the crdb result:

```
x | y  |      z      
---+----+-------------
 1 | '' | 
 2 | '' | 'some text'
```

- It should be 
```
 x | y |     z     
---+---+-----------
 1 |   | NULL
 2 |   | some text
```

On the CRDB side, for both `IMPORT INTO` and `COPY FROM`, only when the entry is [a double quoted empty string](https://www.cockroachlabs.com/docs/stable/import-into#:~:text=When%20importing%20from%20CSV%20without%20nullif%2C%20empty%20values%20are%20interpreted%20as%20NULL%20by%20default%2C%20unless%20enclosed%20in%20double%20quotes.), it will be recognized as an empty string, rather than null. 

> When importing from CSV without nullif, empty values are interpreted as NULL by default, unless enclosed in double quotes. 

However, the `Wrtie` function of `encoding/csv` doesn't add double quotes to empty string. So this PR imported the `moltcsv` that is based on `encoding/csv` and allow overriding the quoting behavior for empty strings, if and only if they are for notnull columns.

----

This PR in general fixes the edge case where an empty string is wrongly imported as NULL. For example, this is what we had previously:

- PG
```
CREATE TABLE notnulltable (x int primary key, z text);
INSERT INTO notnulltable VALUES (1, null), (2, 'NULL'), (3, '');
```

- CRDB
```
psql 'postgres://root@localhost:26257/defaultdb?sslmode=disable' -c "\pset null 'REALNULL'" -c "SELECT * FROM notnulltable"
----
Null display is "REALNULL".
 x |    z     
---+----------
 1 | REALNULL
 2 | NULL
 3 | REALNULL
(3 rows)
```

with the change, it now gives

- CRDB
```
psql 'postgres://root@localhost:26257/defaultdb?sslmode=disable' -c "\pset null 'REALNULL'" -c "SELECT * FROM notnulltable"
----
Null display is "REALNULL".
 x |    z     
---+----------
 1 | REALNULL
 2 | NULL
 3 | 
(3 rows)
```

Jira ticket: https://cockroachlabs.atlassian.net/browse/CC-27668

Release note: Fix fetching empty string for not null columns.


dbtable/table.go
```diff
@@ -54,8 +54,31 @@ func (tm DBTable) String() string {
...
+
+func (cl ColumnListWithAttr) ColumnNames() []tree.Name {
+	res := make([]tree.Name, 0)
+	for _, c := range cl {
+		res = append(res, c.Name)
+	}
+	return res
+}
+
+func (cl ColumnListWithAttr) NotNullColIdxList() map[int]bool {
```
ryanluu12345, line 71: Nit: given that this actually returns a mapping a not a list, the naming is a bit imprecise. I'd prefer `NotNullColIdxMapping`

fetch/csv_pipe.go
```diff
@@ -82,12 +83,13 @@ func (p *csvPipe) Pipe(tn dbtable.Name) error {
...
 		}
-		if err := p.csvWriter.Write(record); err != nil {
+		emptyMap := make(map[int]bool)
+		if err := p.csvWriter.Write(record, t.Columns.NotNullColIdxList()); err != nil {
 			return err
 		}
 
 		if p.testingKnobs.TriggerCorruptCSVFile {
-			if err := p.csvWriter.Write([]string{"this", "should", "lead", "to", "an", "error"}); err != nil {
+			if err := p.csvWriter.Write([]string{"this", "should", "lead", "to", "an", "error"}, emptyMap); err != nil {
```
ryanluu12345, line 92: Why is this `emptyMap` now needed in the writer?

dbtable/table.go
ZhouXing19: Good point, renamed

fetch/csv_pipe.go
ZhouXing19: It's actually just a placeholder. Changed it into nil in this case.

moltcsv/writer.go
```diff
@@ -0,0 +1,195 @@
+// Copyright 2011 The Go Authors. All rights reserved.
```
ryanluu12345, line 1: Not sure I have the most context on this writer, but it looks like it's doing what it needs to: handling the case where there is an empty string but not null properly. Now I understand why we pass in the mapping here. Leave this to Jeremy for a second look

ZhouXing19:
I updated this PR with the last commit that covers the case with nullable columns as well. Now, regardless of the nullability of a column, the mapping from source to target is:

- "NULL" (a string with 4 chars) -> "NULL"
- "" (an empty string) -> ""
- NULL (a null value) -> NULL

The change might be ugly to you, but I can't think of a better way. Have added in-line comments and updated e2e tests too. Thanks again for being patient with me!

e2e/testdata/pg/null.ddt
```diff
@@ -3,17 +3,19 @@ psql 'postgres://postgres@localhost:5432/defaultdb' -c "DROP TABLE IF EXISTS not
 ----
 
 exec silent
-psql 'postgres://postgres@localhost:5432/defaultdb' -c "INSERT INTO notnulltable VALUES (1, '', null), (2, '', 'some text');"
+psql 'postgres://postgres@localhost:5432/defaultdb' -c "INSERT INTO notnulltable VALUES (1, '', null), (2, '', 'NULL'), (3, '', '');"
 ----
 
 exec
-psql 'postgres://postgres@localhost:5432/defaultdb?sslmode=disable' -c "SELECT * FROM notnulltable"
+psql 'postgres://postgres@localhost:5432/defaultdb?sslmode=disable' -c "\pset null 'REALNULL'" -c "SELECT * FROM notnulltable"
```
Jeremyyang920, line 10: Great use of the setting to differentiate the values!

fetch/dataexport/dataexport.go
```diff
@@ -94,19 +85,37 @@ func scanWithRowIterator(
...
 				fmtFlags = tree.FmtExport | tree.FmtPgwireText
 			case *tree.DString:
-				// TODO(janexing): is there a better way to distinguish an empty string?
-				const quotedEmptyStr = "''"
-				if t.String() == quotedEmptyStr && columnNamesWithNotNullAttr[i].notNull {
-					fmtFlags = tree.FmtSimple
+				fmtFlags = tree.FmtExport | tree.FmtParsableNumerics | tree.FmtPgwireText
+				compEmptyStrRes, err := t.CompareError(comparectx.CompareContext, &emptyStr)
+				if err != nil {
+					return errors.Wrapf(err, "cannot check if the datum %s is empty string", t)
```
Jeremyyang920, line 91: nit: "error checking if the datum %s is an empty string"

fetch/dataexport/dataexport.go
```diff
@@ -94,19 +85,37 @@ func scanWithRowIterator(
...
+				}
+				// If this is an empty string, further check if this is a null.
+				if compEmptyStrRes == 0 {
+					compResForNull, err := t.CompareError(comparectx.CompareContext, tree.DNull)
+					if err != nil {
+						return errors.Wrapf(err, "cannot check if the datum %s is tree.DNull", t)
```
Jeremyyang920, line 97: nit: error checking if the datum %s if of type tree.DNull

Jeremyyang920 (approved):

PR #202, opened 2024-04-03
molt-nightly.sh: add explicit fetch args for each test case

ryanluu12345:
Release Note: None

Test Run:
https://teamcity.cockroachdb.com/buildConfiguration/Internal_Molt_MoltNightly/14690863?hideProblemsFromDependencies=false&hideTestsFromDependencies=false


build/teamcity/internal/molt/molt-nightly.sh
```diff
@@ -10,11 +10,11 @@ DIR="$(cd "$(dirname "$0")" && pwd)"
 # Add new case statements for new variants
 case $VARIANT in
   base_aws)
-    $DIR/setup-scale-test.sh --cluster-name migrations-nightly-base-aws --source "$SOURCE_CONN" -n 4 --cloud aws --zones us-east-1a 
+    $DIR/setup-scale-test.sh --cluster-name migrations-nightly-base-aws --source "$SOURCE_CONN" -n 4 --cloud aws --zones us-east-1a --fetch-args "--non-interactive --bucket-path 's3://molt-test/rluu-test' --table-handling 'drop-on-target-and-recreate' --table-filter 'test_table_large' --export-concurrency 4"
```
Jeremyyang920, line 13: We should use a migrations-ci bucket and not a personal prefix.

Jeremyyang920 (approved):

build/teamcity/internal/molt/molt-nightly.sh
ryanluu12345: Which migrations bucket is public and used for tests like these?

ryanluu12345: And which AWS credentials should be used to access it? (I can update the internal job's env)

build/teamcity/internal/molt/setup-scale-test.sh
```diff
@@ -94,6 +96,11 @@ handle_options() {
       --source)
         SOURCE=$(extract_argument $@)
 
+        shift
+        ;;
+      --fetch-args)
+        FETCH_ARGS=$(extract_argument "$@")
```
ryanluu12345, line 102: We want to keep things quoted here, hence the "". We want to pass the string straight down as the arg

build/teamcity/internal/molt/molt-nightly.sh
ryanluu12345: Let me update this in a follow up change. Gonna merge this for now.

PR #206, opened 2024-04-04
fetch/moltcsv: fix checking for number of field

ZhouXing19:
For the case where there is an empty string entry in a row, we add an extraneous entry at the end of the string slice to be written to the csv pipe.

For example, if the original row in the table is:
```
5,'',6,''
```

The string slice piped into the csv pipe would be
```
5,'',6,'','1+3'
```
The last `1+3` entry is to let the "read" side of the pipe to know that the entry of indexes 1 and 3 are empty strings, and they should be double quoted when being written to the csv *file*.

It means that we need to be more smart when checking if the number of entry in each row are consistent. This PR is to fix it.

Release note (bug fix): fix checking for number of field when exporting rows to csv.


ryanluu12345 (approved):

moltcsv/writer.go
```diff
@@ -193,3 +193,8 @@ func (w *Writer) fieldNeedsQuotes(field string) bool {
 	r1, _ := utf8.DecodeRuneInString(field)
 	return unicode.IsSpace(r1)
 }
+
+const (
+	QuoteColumnIdxesPrefix    = `quote column idxes:`
+	QuoteColumnIdxesSeparator = `+`
```
Jeremyyang920, line 199: No real preference, but does it makes more sense to use a `,` as the separator instead of +? I know it makes no real impact.

ZhouXing19: It's because "," is a common separator of Csv files, i was worried that it might confuse the read side of the csv pipe, so avoided using it to concat the indexes here.

PR #207, opened 2024-04-04
fetch: use `crdb_internal.protect_mvcc_history` to protect timestamp for crdb source

ZhouXing19:
When exporting data from a crdb source, we lock a timestamp and run `SELECT ... AS OF SYSTEM TIME` with this timestamp to iterate rows from a table. However, if the table is of very large size and the SELECT statement runs long, the garbage collection job of the crdb source may clean up the snapshot of data of this particular timestamp. It means that we need a history retention job to protect this timestamp from being garbage collected.

That's the reason we introduce the `crdb_internal.protect_mvcc_history()` builtin function that protect the given timestamp for a given period of time. We also extend this protection every time the number of export file reach certain threshold. Once the export of data is finished, we cancel this history retention job so that this timestamp is free to be cleaned up.

User won't need to configure this.

We also updated the docker image for crdb (both source and target) to a recent 24.1 nightly built, but we should also test on an older version of crdb source to ensure that the version gate works. Will do that in a later PR. 

Release note: use `crdb_internal.protect_mvcc_history` to protect timestamp for crdb source


fetch/dataexport/crdb.go
```diff
@@ -2,29 +2,132 @@ package dataexport
...
+	"github.com/rs/zerolog"
 )
 
 type crdbSource struct {
-	aost     time.Time
+	aost time.Time
+
+	mu struct {
+		syncutil.Mutex
+		hrJobId string
```
Jeremyyang920, line 26: nit: hrJobID

fetch/dataexport/crdb.go
```diff
@@ -2,29 +2,132 @@ package dataexport
...
+const (
+	noneAccessChecked historyRetentionAccessibility = iota
+	oldCRDVVersionAccess
+	historyRetentionAccessible
+)
+
+func checkIfAllowHistoryRetention(
+	ctx context.Context, conn *dbconn.PGConn,
+) (historyRetentionAccessibility, error) {
+	var boolRes bool
```
Jeremyyang920, line 43: nit: We should name this something more specific to what the boolean is doing. Maybe like a hasFeature or something along those lines.

fetch/export_table.go
```diff
@@ -25,11 +26,9 @@ type exportResult struct {
 }
 
 func getWriter(w *Pipe, compressionType compression.Flag) io.WriteCloser {
-	switch compressionType {
-	case compression.GZIP:
+	if compressionType == compression.GZIP {
```
Jeremyyang920, line 29: Whats the reason for changing the switch case?

fetch/fetch.go
```diff
@@ -431,21 +436,87 @@ func fetchTable(
...
+				}
+				break extendHistoryRetentionJobLoop
+			default:
+				if hrJobId != "" {
+					exportedCSVNum := atomic.LoadInt64(&exportedCSVCnt)
+					if exportedCSVNum > 0 && exportedCSVNum/extensionThreshold > multiplier {
+						logger.Debug().Msgf("seen %d csv exported, extending hisrory retention job %s", exportedCSVNum, hrJobId)
+						multiplier = exportedCSVNum / extensionThreshold
+						if err := sqlSrc.ExtendHistoryRetentionJob(ctx, logger, extendHRJobCntPtr, cfg.TestOnly); err != nil {
+							logger.Warn().Msgf("failed to extend history retention job %s, export job might fail due to snapshot expiration", hrJobId)
```
Jeremyyang920, line 511: I'm curious on this approach and what you think the benefits of using the atomic counter are and introducing this ratio check versus using a old fashioned time.After in a go routine.

fetch/dataexport/crdb.go
```diff
@@ -2,29 +2,132 @@ package dataexport
...
+				tsProtectedConfirmed = true
+				break
+			}
+		}
+		if !tsProtectedConfirmed {
+			return nil, errors.Wrapf(err, "failed confirm timestamp %s is protected", ts)
+		}
+		res.SetHistoryRetentionJobId(strconv.FormatInt(jobID, 10))
+	} else {
+		logger.Warn().Msgf("crdb version might not support history retention for specific timestamp. AS OF SYSTEM TIME with timestamp %s might fail for long export", ts)
```
Jeremyyang920, line 112: I think we should specify here that they should increase their zone configs since thats the workaround.

fetch/fetch.go
ZhouXing19: Single routine:
- Pro: easier communication between channels. 
- Con: If it takes long time to export each file, it might not be able to renew the hr job in time. It means there might need more tuning on renew threashold / flush row / flush size and etc. 

Ticker:
- Pro: Can always ensure the job is renewed before expiration.
- Con: More complicated communication between go routines: ticker routine need to know the export has finished or failed to release the protected timestamp, the main routine also need to confirm that the history retention job is cancelled before proceeding to the import stage (otherwise, if the table is small, there might not be enough time for the ticker routine to cancel job while the whole fetch process has finished). 

I'm neutual to both approach to be honest, it's like leaving the trouble to the devs or to the users.

fetch/dataexport/crdb.go
ZhouXing19: Which zone configs...?

fetch/export_table.go
ZhouXing19: My IDE thinks it's better ;)

fetch/fetch.go
Jeremyyang920: For the ticker cons, we know when export finishes because of the the wait group returning. We don't start import until after 
```		
for _, er := range orderedResults {
			e.StartTime = time.Unix(min(e.StartTime.Unix(), er.StartTime.Unix()), 0)
			e.EndTime = time.Unix(max(e.EndTime.Unix(), er.EndTime.Unix()), 0)
			e.NumRows += er.NumRows
			e.Resources = append(e.Resources, er.Resources...)
		}
```
is done calculating. 
So something like 
```
doneCh := make(chan struct{})
		go func() {
			for {
				select {
				case <-time.After(1 * time.Hour):
					// do the heartbeat
				case <-doneCh:
					// wg.Done for this go routine and wg wait before import.
					// cancel job
				}
			}
		}()
		if err = wg.Wait(); err != nil {
			return err
		}
```
might be enough. Since looking at the cancelJob code, we don't do anything special if it fails. It will just expire by itself.

fetch/dataexport/crdb.go
Jeremyyang920: https://www.cockroachlabs.com/docs/stable/show-zone-configurations
gc.ttlseconds

fetch/dataexport/crdb.go
```diff
@@ -2,29 +2,132 @@ package dataexport
...
 )
 
 type crdbSource struct {
-	aost     time.Time
+	aost time.Time
+
+	mu struct {
+		syncutil.Mutex
+		hrJobID string
+	}
```
Jeremyyang920, line 27: is there risk of concurrenct hrJobID access? Since the current implementation is single thread right?

ZhouXing19: Yes, if the export job finishes instantly, the line `hrJobId := sqlSrc.HistoryRetentionJobId()` and `sqlSrc.CancelHistoryRetentionJob()` will cause data race, though it's not strictly concurrent accessing hrJobID. I saw this with [this test](https://github.com/cockroachdb/molt/actions/runs/8623276983) with sha `ecd72cd489a4e428c765b5f220b74e722d149b84`

fetch/fetch.go
ZhouXing19: Sounds good, that's how I implemented in another draft branch, so I can switch to this approach shortly. But just to confirm, are we going to settle with the ticker solution?

Jeremyyang920: I think it might be easier to maintain and understand compared to the file ratio and the need to use an atomic counter. WDYT?

cc @ryanluu12345 I want to hear your input on this one too.

ryanluu12345: I'm trying to think from the perspective of us a few months from now when we have to maintain this. I do agree that the ticker seems more maintainable and easier to reason about. The main benefit I see from the ticker is the benefit you noted about being able to renew the job before expiration. It would be a bad user experience if the renew mechanism doesn't fire in time and a large export fails after 24hrs.

.github/docker-compose.yml
```diff
@@ -17,12 +17,12 @@ services:
       - "4443:4443"
     command: ["-scheme", "http"]
   cockroachdb:
-    image: cockroachdb/cockroach:latest-v23.1
+    image: us-docker.pkg.dev/cockroach-cloud-images/cockroachdb/cockroach:v24.1.0-alpha.5-146-gc43f54cdde5
```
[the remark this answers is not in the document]
Jeremyyang920: I think we may have to wait for a public 24.1 image. But we may now need to run CI against different versions to ensure the non 24.1 versions do the right thing.

fetch/dataexport/crdb.go
```diff
@@ -2,29 +2,132 @@ package dataexport
...
+			if tm.Equal(ts) {
+				logger.Info().Msgf("timestamp is protected for 24h: timestamp: %d, meta_type: %s, pb: %s", utils.MaybeFormatTimestamp(testOnly, tsInt), metaTyp, pbJson)
+				tsProtectedConfirmed = true
+				break
+			}
+		}
+		if !tsProtectedConfirmed {
+			return nil, errors.Wrapf(err, "failed confirm timestamp %s is protected", ts)
+		}
+		res.SetHistoryRetentionJobId(strconv.FormatInt(jobID, 10))
```
Jeremyyang920, line 110: Im thinking more about this block and whether it is fully necessary since its more so a test of whether the built-in did the right thing and that functionality should reside with the DB tests. Wondering if its sufficient that based on contract of what calling the built in should do is enough as long as it returns a non nil error.

ZhouXing19:
Now this PR takes the ticker approach and I have resolved all comments. RFAL, thanks!

cmd/fetch/fetch.go
```diff
@@ -143,6 +145,37 @@ func Command() *cobra.Command {
...
+						_, err = pgxConnForCRDB.Exec(ctx, `RESET CLUSTER SETTING bulkio.import.retry_duration`)
+						if err != nil {
+							logger.Warn().Err(err).Msgf("failed resetting cluster setting bulkio.import.retry_duration")
+						}
+					}()
+				}
+			}
+			cfg.ExportSettings.CRDBPTSExtensionFreq, err = time.ParseDuration(crdbPTSExtensionFreqStr)
+			if err != nil {
+				return errors.Wrapf(err, "failed getting the protected timestamp duration for crdb")
```
Jeremyyang920, line 172:
```suggestion
				return errors.Wrapf(err, "failed to parse the protected timestamp duration for crdb")
```

cmd/fetch/fetch.go
```diff
@@ -300,6 +333,14 @@ func Command() *cobra.Command {
 		100_000,
 		"Number of rows to select at a time for export from the source database.",
 	)
+
+	cmd.PersistentFlags().StringVar(
```
Jeremyyang920, line 337: You can use a duration var for this instead of string

fetch/dataexport/crdb.go
```diff
@@ -2,29 +2,78 @@ package dataexport
...
+	if isAfter241 {
+		newConn, err := conn.Clone(ctx)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to clone a connection to enforce history retention")
+		}
+		defer func() { newConn.Close(ctx) }()
+		var jobID int64
+		pgConn := newConn.(*dbconn.PGConn)
+		unixTime := ts.Unix()
+		query := fmt.Sprintf("SELECT crdb_internal.protect_mvcc_history('%d', '24h'::interval, 'molt fetch data export')", unixTime)
```
Jeremyyang920, line 54: I'm wondering if we set 24h default here or use the supplied variable. That could default to 24h

fetch/dataexport/crdb.go
```diff
@@ -2,29 +2,78 @@ package dataexport
...
+		logger.Debug().Msgf("running history retention job with: %s", query)
+		err = pgConn.QueryRow(ctx, query).Scan(&jobID)
+		if err != nil {
+			return nil, errors.WithHintf(
+				errors.Wrapf(err, "failed to protect timestamp for %s", ts),
+				"does your crdb source db support crdb_internal.protect_mvcc_history() built-in function?",
+			)
+		}
+		logger.Info().Msgf("timestamp is protected for 24h: timestamp: %d", utils.MaybeFormatTimestamp(testOnly, unixTime))
```
Jeremyyang920, line 63: this 24h here may need to be a string format

fetch/dataexport/crdb.go
```diff
@@ -64,6 +113,103 @@ func (c *crdbSourceConn) Export(
...
 	return c.conn.Close(ctx)
 }
+
+func (c *crdbSource) cancelRetentionJob(
+	ctx context.Context, logger zerolog.Logger, testOnly bool,
+) error {
+	jobId := c.hrJobID
+	if jobId == "" {
+		return nil
+	}
```
Jeremyyang920, line 126: Why the assignment to a variable if the pointer receiver has the hr.JobID on it already?

fetch/dataexport/crdb.go
```diff
@@ -64,6 +113,103 @@ func (c *crdbSourceConn) Export(
...
+				}
+			}()
+			if err != nil {
+				return err
+			}
+			if breakLoop {
+				break
+			}
+		}
+		logger.Info().Msgf("exit history retention management go routine")
```
Jeremyyang920, line 211:
```suggestion
		logger.debug().Msgf("exiting history retention management go routine")
```

fetch/dataexport/crdb.go
```diff
@@ -64,6 +113,103 @@ func (c *crdbSourceConn) Export(
...
+				}
+				defer func() {
+					retErr = newConn.Close(ctx)
+					if retErr != nil {
+						logger.Err(err).Msgf("failed closing connection created for history retention check")
+					}
+				}()
+				newPgConn := newConn.(*dbconn.PGConn)
+
+				select {
```
Jeremyyang920, line 184: I think if may be cleaner if you do 
```
looplabel:
  for{
  // setup connections
  select:
 // do the select stuff here
// break loop label as needed or return err.
```

I think this cuts down some of the complexity of the logic.

fetch/dataexport/crdb.go
```diff
@@ -64,6 +113,103 @@ func (c *crdbSourceConn) Export(
...
+				case <-ticker.C:
+					return func() (bool, error) {
+						hrJobID := c.hrJobID
+						if _, err := newPgConn.Exec(ctx, `SELECT crdb_internal.extend_mvcc_history_protection($1)`, hrJobID); err != nil {
+							msg := fmt.Sprintf("failed extending protection interval for job id %s", utils.MaybeFormatHistoryRetentionJobID(testOnly, hrJobID))
+							logger.Warn().Msgf(msg)
+							return true, errors.Wrapf(err, msg)
+						}
+						*extensionCnt++
+						logger.Info().Msgf("has done %d history retention interval extention for job id %s", *extensionCnt, utils.MaybeFormatHistoryRetentionJobID(testOnly, hrJobID))
```
Jeremyyang920, line 198: updated history retention for job id %s. Total updates : %d

I think this might make more sense as a log message.

fetch/dataexport/crdb.go
```diff
@@ -64,6 +113,103 @@ func (c *crdbSourceConn) Export(
...
+				select {
+				// If the export job has finished, we cancel the history retention job and exit the goroutine.
+				case <-exportFinished:
+					return true, c.cancelRetentionJob(ctx, logger, testOnly)
+				// If timeout, we extend the history retention job and enter the next loop.
+				case <-ticker.C:
+					return func() (bool, error) {
+						hrJobID := c.hrJobID
+						if _, err := newPgConn.Exec(ctx, `SELECT crdb_internal.extend_mvcc_history_protection($1)`, hrJobID); err != nil {
+							msg := fmt.Sprintf("failed extending protection interval for job id %s", utils.MaybeFormatHistoryRetentionJobID(testOnly, hrJobID))
```
Jeremyyang920, line 193:
```suggestion
							msg := fmt.Sprintf("failed to extend protection interval for job id %s", utils.MaybeFormatHistoryRetentionJobID(testOnly, hrJobID))
```

fetch/dataexport/crdb.go
```diff
@@ -64,6 +113,103 @@ func (c *crdbSourceConn) Export(
...
+	pgConn := c.conn.(*dbconn.PGConn)
+	cn, err := pgConn.Clone(ctx)
+	if err != nil {
+		return errors.Wrapf(err, "failed to clone a connection to cancel history retention job")
+	}
+	defer cn.Close(ctx)
+	conn := cn.(*dbconn.PGConn)
+
+	if _, err := conn.Exec(ctx, `CANCEL JOB $1`, jobId); err != nil {
+		msg := fmt.Sprintf("failed canceling history retention job with id %s", utils.MaybeFormatHistoryRetentionJobID(testOnly, jobId))
```
Jeremyyang920, line 141:
```suggestion
		msg := fmt.Sprintf("failed to cancel history retention job with id %s", utils.MaybeFormatHistoryRetentionJobID(testOnly, jobId))
```

fetch/dataexport/crdb.go
```diff
@@ -2,29 +2,78 @@ package dataexport
...
+		logger.Warn().Msgf("crdb version might not support history retention for specific timestamp. AS OF SYSTEM TIME with timestamp %s might fail for long export. Try setting a larger value for gc.ttlseconds", ts)
+	}
+	return res, nil
+}
+
+func (c *crdbSource) resetJobId() {
+	c.hrJobID = ""
+}
+
+func (c *crdbSource) SetHistoryRetentionJobId(s string) {
```
Jeremyyang920, line 75: I see this is only used above on line 64. Is there need for this explicit setter if we are dealing with the crdbSource object directly?

fetch/dataexport/crdb.go
```diff
@@ -64,6 +113,103 @@ func (c *crdbSourceConn) Export(
...
+	ctx context.Context, logger zerolog.Logger, testOnly bool,
+) error {
+	jobId := c.hrJobID
+	if jobId == "" {
+		return nil
+	}
+	defer func() {
+		c.resetJobId()
+		logger.Info().Msgf("done resetting history retention job id")
+	}()
```
Jeremyyang920, line 130: Curious if its actually needed to resetJobID() since if we do reset it, chances are we failed somewhere and have to restart which. means we will create a new crdbSource object on the next run.

fetch/fetch.go
```diff
@@ -199,8 +199,12 @@ func Fetch(
 	if err != nil {
 		return err
 	}
+	if tables == nil {
+		logger.Warn().Msgf("no table to be imported, exiting molt fetch")
```
Jeremyyang920, line 203:
```suggestion
		logger.Warn().Msgf("no table(s) to be imported, exiting molt fetch")
```

Jeremyyang920 (commented):
Overall looks good. Sorry its alot of minor comments.

fetch/dataexport/dataexport.go
```diff
@@ -15,12 +16,22 @@ import (
...
 	Conn(ctx context.Context) (SourceConn, error)
 	Close(ctx context.Context) error
+
+	// HistoryRetention means to extend the expiration of a protected timestamp (PTS). A
+	// SELECT ... AS OF SYSTEM TIME command with a PTS will not be invalidated by garbage collection.
+	// A PTS is created when creating the
+	// cockroachdb datasource (i.e. NewCRDBSource()) via the crdb_internal.protect_mvcc_history(<timestamp>, <lifetime>,
+	// <description>).
+	// At time t, if extend the history retention job, the job's expiration timestamp will be updated to {t + <lifetime>}.
+	HistoryRetentionJobManagement(ctx context.Context, logger zerolog.Logger, exportFinished chan struct{}, renewInterval time.Duration, extensionCnt *int64, testOnly bool) (jobManagementFinished *errgroup.Group)
```
ryanluu12345, line 34: I think this is a more general comment about our interfaces throughout the code base. But how should we handle the cases where an interface method only applies to a single dialect only?

Right now, we just make them no-ops, but we could potentially put it on the specific struct and cast it to a specific type. But maybe more complex than it needs to be.

fetch/import_table.go
```diff
@@ -84,15 +84,16 @@ func reportImportTableProgress(
...
-	}
-	defer conn.Close(ctx)
-
 	prevVal := 0.0
 
-	if err := r.Do(func() error {
+	if err := r.Do(func() (retErr error) {
+		// We need to recreate new connection for each retry. Somehow keep using the same connection won't be able to see the
+		// most up-to-date job.
+		conn, err := pgx.ConnectConfig(ctx, pgConn.Config())
```
ryanluu12345, line 92: What's the performance implication of creating a new connection each time we retry? We poll until it's done so we may be taxing the target DB each time.

fetch/import_table_test.go
```diff
@@ -103,15 +129,18 @@ func TestReportImportTableProgress(t *testing.T) {
 		var b bytes.Buffer
 		logger := zerolog.New(&b)
 
+		// Give some buffer time so that the import job can be registered and become queriable via SHOW JOBS.
+		time.Sleep(30 * time.Second)
```
ryanluu12345, line 133: Is there any way we can make this test a bit more robust. This 30 second delay could end up causing flakes in the future if the job gets registered slightly longer than 30 seconds. One way we can do this is to poll in a `goroutine` for 35 to 45 seconds and if the job shows up, you can then exit the goroutine and test.

fetch/import_table_test.go
```diff
@@ -90,6 +90,32 @@ func TestReportImportTableProgress(t *testing.T) {
...
+				_, err = pgConn.Exec(ctx, `RESET CLUSTER SETTING jobs.registry.interval.adopt`)
+				require.NoError(t, err)
+				_, err = pgConn.Exec(ctx, `RESET CLUSTER SETTING jobs.registry.interval.cancel`)
+				require.NoError(t, err)
+				_, err = pgConn.Exec(ctx, `RESET CLUSTER SETTING jobs.metrics.interval.poll`)
+				require.NoError(t, err)
+			}()
+		}
+
+		// Give some buffer time so that the import job can be registered and become queriable via SHOW JOBS.
```
ryanluu12345, line 117: Why do we need this sleep twice?

fetch/dataexport/dataexport.go
ZhouXing19: I personally think no-op is fine, because for other dialects there's no such functionality, so it's literally a no-op. Adding another struct to specify the crdb case can be a bit cumbersome. Also, if in the future, newly supported dialects have similar functionalities, we can implement for them conveniently as well.

fetch/import_table_test.go
ZhouXing19: The first sleep is wait for the cluster settings to take effect; the second sleep is to wait for the import job being registered.

fetch/import_table_test.go
ZhouXing19: >One way we can do this is to poll in a goroutine for 35 to 45 seconds and if the job shows up, you can then exit the goroutine and test. 
 
 Hmm, i don't really understand this. Are you saying have a goroutine keeps retrying `SHOW JOBS` until the import job shows up? But isn't this what `reportImportTableProgress` does?

cmd/fetch/fetch.go
```diff
@@ -143,6 +145,37 @@ func Command() *cobra.Command {
...
+						}
+					}()
+				}
+			}
+			cfg.ExportSettings.CRDBPTSExtensionFreq, err = time.ParseDuration(crdbPTSExtensionFreqStr)
+			if err != nil {
+				return errors.Wrapf(err, "failed getting the protected timestamp duration for crdb")
+			}
+
+			if !cfg.TestOnly && cfg.ExportSettings.CRDBPTSExtensionFreq < 10*time.Second {
```
ryanluu12345, line 175: Nit: may be worthwhile to declare a variable like `minProtectedTimestampExtensionDur`

fetch/dataexport/crdb.go
ZhouXing19: The reason i want to reset the job id is to make this `cancelRetentionJob` idempotent -- once a job is cancelled, we will not cancel it again. 

>  if we do reset it, chances are we failed somewhere and have to restart which means we will create a new crdbSource object on the next run.

Hmm i don't really get this, isn't this the expected behavior? I can't see the relation of this and resetting the job id

fetch/fetch.go
```diff
@@ -431,28 +435,58 @@ func fetchTable(
...
-					return err
+
+		var jobManagementWg *errgroup.Group
+
+		exportTableFunc := func() (ret error) {
+			exportFinishedChan := make(chan struct{})
+			defer func() {
+				close(exportFinishedChan)
+				if jobManagementWg != nil {
+					retErr = errors.CombineErrors(retErr, jobManagementWg.Wait())
```
ryanluu12345, line 446: Hmm if the wg.Wait returns `nil` but there is a `retErr` wouldn't this mean the returned error would be `nil` even though an error happens

fetch/fetch.go
```diff
@@ -431,28 +435,58 @@ func fetchTable(
...
-				if err != nil {
-					return err
+
+		var jobManagementWg *errgroup.Group
+
+		exportTableFunc := func() (ret error) {
+			exportFinishedChan := make(chan struct{})
+			defer func() {
+				close(exportFinishedChan)
+				if jobManagementWg != nil {
```
ryanluu12345, line 445: Why do this part in the `defer`. The channel close makes sense though.

fetch/fetch.go
ZhouXing19: With `wg.Wait()` do you mean by `jobManagementWg.Wait()`, or `wg` in `exportTableFunc()`?

But either case, it should be fine, because `errors.CombineErrors(nil, nil) == nil, errors.Combine(err, nil) == err, errors.Combine(nil, err) = err`

fetch/fetch.go
ZhouXing19: > Why do this part in the defer. 
 because we only cancel the job when the export job finishes, whether it succeeds or fails. so i put it in a defer here for such a general case.

fetch/dataexport/crdb.go
ZhouXing19: I though about that, but the tricky part is that I have 
```
defer func() {
					retErr = newConn.Close(ctx)
					if retErr != nil {
						logger.Err(err).Msgf("failed closing connection created for history retention check")
					}
				}()
```
In each iteration of the for loop. It's because a new connection is created for a new extension attempt. It doesn't make sense to reuse a connection for each renewal since it basically sits idle most of the time waiting for the ticker. 

So the defer is for the closure of each iteration, so i have to wrap it in a function. But, inside a function, `looplabel` is no longer accessible, so `break looplabel` doesn't work.

I modified it a bit to make it cleaner, but unfortunately it might still look quite ugly ><

fetch/import_table.go
ZhouXing19: I just re-tested and noticed reusing the connection works too, so changed it back.

fetch/dataexport/crdb.go
Jeremyyang920: The idempotency logic makes sense. But with the conditions in which we reset which is successful export in which the jobID doesn't matter anymore or there is an error and we are exiting fetch in which case our jobID is lost anyways as it lives with the crdbconn object. 

I guess long way of asking, in what situation would we be cancelling a job multiple times if we have 1 thread only controlling the job?

cmd/fetch/fetch.go
```diff
@@ -300,6 +329,21 @@ func Command() *cobra.Command {
...
+	cmd.PersistentFlags().DurationVar(
+		&cfg.ExportSettings.CRDBPTSExtensionFreq,
+		"crdb-pts-extension-frequency",
+		10*time.Minute,
+		"Duration for the ticker to wait for the next protected timestamp extension. The new expiration timestamp of the protected timestamp will be {time when extension happens} + {crdb-pts-extension-lifetime}",
+	)
+
+	cmd.PersistentFlags().DurationVar(
+		&cfg.ExportSettings.CRDBPTSExtensionLifetime,
+		"crdb-pts-extension-lifetime",
```
Jeremyyang920, line 342: WDYT about crdb-pts-duration. I think that might be less wordy but still convey what we want.

cmd/fetch/fetch.go
```diff
@@ -300,6 +329,21 @@ func Command() *cobra.Command {
 		100_000,
 		"Number of rows to select at a time for export from the source database.",
 	)
+
+	cmd.PersistentFlags().DurationVar(
+		&cfg.ExportSettings.CRDBPTSExtensionFreq,
+		"crdb-pts-extension-frequency",
```
Jeremyyang920, line 335: WDYT about crdb-pts-refresh-interval?

fetch/dataexport/crdb.go
Jeremyyang920: I see your point. Though im curious if its the right thing to do with failing the entire fetch job and exiting the PTS go routine for a single connection failing to close. Im wondering if in this case its a bit extreme.

fetch/dataexport/crdb.go
Jeremyyang920: Oh something that just came to mind too is, the cancelRetentionJob is inside that go routine below, but what if an error from export happens and we are exiting, do we ever call cancleJob in this case to clean up?

ZhouXing19:
> I see your point. Though im curious if its the right thing to do with failing the entire fetch job and exiting the PTS go routine for a single connection failing to close. Im wondering if in this case its a bit extreme.

I see. What about if we failed to close this connection, we just log the error but don't let the function return it. So that the whole fetch job will still proceed. All i want is to have the main fetch routine the job is cancelled before proceeding with the import stage.

ZhouXing19:
> Oh something that just came to mind too is, the cancelRetentionJob is inside that go routine below, but what if an error from export happens and we are exiting, do we ever call cancleJob in this case to clean up?

It has been ensured with 
```
defer func() {
	close(exportFinishedChan)
        ....
}
```
in the definition of `exportTableFunc`. So no matter if export encounters error or not, it will trigger the job cancellation anyways. I just added a test in `fetch/testdata/crdb/history_retention.ddt` to ensure this too, see the last commit.

fetch/dataexport/crdb.go
```diff
@@ -2,29 +2,77 @@ package dataexport
...
+		if err != nil {
+			return nil, errors.WithHintf(
+				errors.Wrapf(err, "failed to protect timestamp for %s", ts),
+				"does your crdb source db support crdb_internal.protect_mvcc_history() built-in function?",
+			)
+		}
+		logger.Info().Msgf("timestamp is protected for %s: timestamp: %d", settings.CRDBPTSExtensionLifetime.String(), utils.MaybeFormatTimestamp(testOnly, unixTime))
+		res.hrJobID = strconv.FormatInt(jobID, 10)
+	} else {
+		logger.Warn().Msgf("crdb version might not support history retention for specific timestamp. AS OF SYSTEM TIME with timestamp %s might fail for long export. Try setting a larger value for gc.ttlseconds", ts)
```
ryanluu12345, line 69: Should this warning/error message link to docs on how to do that? It would be helpful to understand the implications of doing this before doing it.

fetch/dataexport/dataexport.go
ryanluu12345: Yep agreed. This maintains simplicity here.

fetch/fetch.go
ryanluu12345: Understood

fetch/import_table_test.go
ryanluu12345: What changed within the jobs system that requires these delays now? Do we understand the changes to the DB side of things?

fetch/import_table_test.go
ryanluu12345: Yeah, that's true. Report import table progress does that already. So you'd be implementing the same logic to test the thing, which doesn't seem great. Ok let's just keep the sleeps for now, but let's think about how we can improve this logic without blocking for 40 seconds total. Both takes up time and can be flakey.

ryanluu12345 (approved):
LGTM, just a lot of questions about clarification and considerations for the future

cmd/fetch/fetch.go
```diff
@@ -300,6 +329,21 @@ func Command() *cobra.Command {
 		100_000,
 		"Number of rows to select at a time for export from the source database.",
 	)
+
+	cmd.PersistentFlags().DurationVar(
```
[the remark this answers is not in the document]
ZhouXing19: As long as the ticker works and crdb source is accessible, everytime the ticker checks and extends the PTS, the PTS got extended the original lifetime duration. So as long as the refresh interval is small enough, there won't be case that the extension is not in time.

For example, 
- The PTS is protected at t0, with lifetime 24h. It means the PTS will expire `t0 + 24h`.
- Say the refresh interval is 10 mins. It means at `t1 := t0 + 10min`, we will extend the PTS
- The new expiration timestamp for this PTS will now be `t1 + 24h = t0 + 10min + 24h`.
- And at `t2 := t0 + 20min`, we will extend the PTS again, with new expiration TS `t0 + 20min + 24h`.
So if we set the refresh frequent enough, your case won't happen.

ZhouXing19:
> What changed within the jobs system that requires these delays now? Do we understand the changes to the DB side of things?

I'm not sure. Maybe add a TODO to investigate?

fetch/dataexport/crdb.go
ZhouXing19: I changed it that error from `HistoryRetentionJobManagement` will be not blocking, but just logged. Does it make more sense?

fetch/dataexport/crdb.go
```diff
@@ -83,3 +131,95 @@ func (c *crdbSource) ReplicatorCommand(
...
+	pgConn := c.conn.(*dbconn.PGConn)
+	cn, err := pgConn.Clone(ctx)
+	if err != nil {
+		return errors.Wrapf(err, "failed to clone a connection to cancel history retention job")
+	}
+	defer cn.Close(ctx)
+	conn := cn.(*dbconn.PGConn)
+
+	if _, err := conn.Exec(ctx, `CANCEL JOB $1`, c.hrJobID); err != nil {
+		msg := "failed extending protection interval"
```
Jeremyyang920, line 155: This should be failed to cancel instead of failed extending.

fetch/dataexport/crdb.go
Jeremyyang920: That makes sense. I think we would just document the feature saying ensure your extention duration and ticker is some multiple of each other so that if one fails, you still have a failsafe your ticker will refresh again.

fetch/dataexport/crdb.go
```diff
@@ -83,3 +131,95 @@ func (c *crdbSource) ReplicatorCommand(
...
+					return false, nil
+
+				}
+			}()
+			if err != nil {
+				return err
+			}
+			if breakLoop {
+				break
+			}
```
Jeremyyang920, line 219: nit:  you can make this the for loop condition as for breakLoop, but in that case lets name it more specific on the exit condition. Maybe something like `for runHistoryRetention`

PR #208, opened 2024-04-05
scaletesting: added support for running molt verify + observability

ryanluu12345:
Refactored the test code to be more generic to support both MOLT verify and fetch. Also added in a helper method that takes care of waiting for process completion and downloading artifacts.

Resolves: CC-27755
Release Note: None

Test Run: https://teamcity.cockroachdb.com/buildConfiguration/Internal_Molt_MoltNightly/14717026?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true

Latest Test with Verify Log Parsing: https://teamcity.cockroachdb.com/buildConfiguration/Internal_Molt_MoltNightly/14752770?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true

Flags output:
```
      --table-splits int                         Number of shards to break down each table into while doing row-based verification. (default 1)
      --target string                            Connection string of the target database.
      --use-console-writer                       Use the console writer, which has cleaner log output but introduces more latency (defaults to false, which logs as structured JSON).
(⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 molt % 
```

Test Results:
```
[22:13:19 ](https://teamcity.cockroachdb.com/buildConfiguration/Internal_Molt_MoltNightly/14752770?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&showLog=14752770_1296_1296&logFilter=debug&logView=linear)  0
[22:13:19 ](https://teamcity.cockroachdb.com/buildConfiguration/Internal_Molt_MoltNightly/14752770?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&showLog=14752770_1297_1297&logFilter=debug&logView=linear)  
[22:13:22 ](https://teamcity.cockroachdb.com/buildConfiguration/Internal_Molt_MoltNightly/14752770?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&showLog=14752770_1298_1298&logFilter=debug&logView=linear)  {"level":"info","type":"summary","table_schema":"public","table_name":"test_table_large","num_truth_rows":99999999,"num_success":99999999,"num_conditional_success":0,"num_missing":0,"num_mismatch":0,"num_extraneous":0,"num_live_retry":0,"num_column_mismatch":0,"time":"2024-04-08T22:13:12Z","message":"finished row verification on public.test_table_large (shard 1/2)"}
[22:13:22 ](https://teamcity.cockroachdb.com/buildConfiguration/Internal_Molt_MoltNightly/14752770?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&showLog=14752770_1299_1299&logFilter=debug&logView=linear)  {"level":"info","type":"summary","table_schema":"public","table_name":"test_table_large","num_truth_rows":100000001,"num_success":100000001,"num_conditional_success":0,"num_missing":0,"num_mismatch":0,"num_extraneous":0,"num_live_retry":0,"num_column_mismatch":0,"time":"2024-04-08T22:13:13Z","message":"finished row verification on public.test_table_large (shard 2/2)"}
```


scaletesting/scale_test.go
```diff
@@ -209,6 +251,6 @@ func getProcessStatus(cluster, fileName string) (exitCode int, err error) {
 	return status, nil
 }
 
-func didFetchPass(exitCode int) bool {
+func didProcessPass(exitCode int) bool {
 	return exitCode == 0
```
Jeremyyang920, line 255: Something that im thinking about right now is does verify exit code 0 if it finds a mismatch? We still exit gracefully and just log the issue iirc. So we would pass silently

ryanluu12345: Good point. Do you think, instead, we should verify that number of success + number of conditional success == number of rows seen?

Or should we modify the behavior of verify so that it exits with non-zero if any anomalies are seen?

Jeremyyang920: I think we would want to see the num success is matches our num rows. We shouldn't have conditional successes in our data sets.

ryanluu12345: Even simpler implementation. I'll go ahead and make a check for this as well. I'll make sure that `processPassed` AND `num success ==. num rows`.

ryanluu12345: I implemented this in the latest commit.

scaletesting/scale_test.go
```diff
@@ -254,3 +276,92 @@ func getProcessStatus(cluster, fileName string) (exitCode int, err error) {
...
+		originalLog:    "log3",
+	}, {
+		numTruthRows:   12344,
+		numSuccessRows: 12345,
+		originalLog:    "log4",
+	}}
+
+	require.True(t, didVerifyPass(successfulResults))
+	require.False(t, didVerifyPass(failedResults))
+}
```
Jeremyyang920, line 367: Can you move these to an actual test file?

ryanluu12345: Done!

scaletesting/utils_test.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestExtractVerifyResult(t *testing.T) {
+	fmt.Println(os.Getwd())
```
Jeremyyang920, line 12: nit: remove the println in the test

Jeremyyang920 (approved):

scaletesting/utils_test.go
ryanluu12345: Done!

PR #209, opened 2024-04-05
CDC sink integration with fetch

Jeremyyang920:
This PR sets the foundation for the replication step of MOLT `fetch`. The changes introduced in this PR set up all the infrastructure, commands, and helpers that help to run `cdc-sink` as a subprocess. Each of the setup steps needed to fully run `cdc-sink` are codified in the setup and the replicator commands needed are implemented in each of the language dialect structs.

This change additionally adds the following:
- Flag to gate the usage of Cockroach source replication because lack of maturity and details figured out
- Handling signals for terminate/kill and errors that come from the subprocess `cdc-sink` run
- Docker compose updates to ensure that all of the containers can integrate with `cdc-sink` out of the box

Resolves: CC-27683, CC-27867, CC-27868
Release Note: None

**TODOS**
- [x] Rebase
- [ ] Update ReadME

**Test Results**
Automated tests will be in a follow up PR. Since the testing required here is at the integration layer.

CRDB:
```
DEBUG  [Apr 11 16:59:32] staged mutations                              count=1 duration=14.093916ms target="\"_cdc_sink\".\"public\".\"defaultdb_public_employees\""
```

PG:
```
DEBUG  [Apr 11 17:02:52] upserted rows                                 conflicts=0 duration=4.280167ms proposed=1 target="\"molt\".\"public\".\"employees\"" upserted=1
```

PG Error Case:
```
{"level":"info","time":"2024-04-11T17:05:30-07:00","message":"creating publication"}
ERRO[0000] exited                                        error="unknown command \"logical\" for \"cdc-sink\"\n\nDid you mean this?\n\tmylogical\n\tpglogical\n"
{"level":"error","time":"2024-04-11T17:05:30-07:00","message":"error while running CDC sink"}
{"level":"error","error":"os: process already finished","time":"2024-04-11T17:05:30-07:00"}
Error: CDC sink command failed to run
```

MySQL:
```
DEBUG  [Apr 11 17:08:33] upserted rows                                 conflicts=0 duration=4.208292ms proposed=1 target="\"molt\".\"public\".\"employees\"" upserted=1
```


fetch/fetch.go
```diff
@@ -351,6 +354,52 @@ func Fetch(
 		Dur("net_duration_ms", ovrDuration).
 		Str("net_duration", utils.FormatDurationToTimeString(ovrDuration)).
 		Msgf("fetch complete")
+
+	if cfg.CDCSinkPath != "" && len(dbTables.Verified) > 0 {
```
ryanluu12345, line 358: Nit: could just make a helper that has the form `canRunCDCSink`

fetch/dataexport/pg.go
```diff
@@ -141,3 +143,23 @@ func (p *pgSourceConn) Export(
...
+	return exec.Command(
+		bin,
+		"pglogical",
+		"--sourceConn", p.conn.ConnStr(),
+		"--targetConn", target.ConnStr(),
+		"--publicationName", "molt_fetch",
+		"--targetSchema", fmt.Sprintf("%s.%s", db, sc),
+		"--slotName", fmt.Sprint(p.settings.PG.SlotName),
+		"--stagingSchema", fmt.Sprint(db),
+		"--parallelism", "256",
```
ryanluu12345, line 162: Do we want to make these replicator commands configurable via a flag so people can tune the performance params? I guess this can be a follow up.

fetch/fetch.go
```diff
@@ -351,6 +354,52 @@ func Fetch(
...
+
+		cmd, err := sqlSrc.ReplicatorCommand(binName, conns[1], conns[1].Database(), dbTables.Verified[0][1].Schema)
+		if err != nil {
+			return err
+		}
+		// Create the changefeed on the source side
+		if conns[0].IsCockroach() {
+			tblQ := replicator.CreateTableList(tables)
+			q := fmt.Sprintf(`CREATE CHANGEFEED FOR TABLE %s
+			INTO 'webhook-https://localhost:30004/%s/%s?insecure_tls_skip_verify=true'
```
ryanluu12345, line 383: Where does this webhook-https get started?

fetch/replicator/replicator.go
```diff
@@ -0,0 +1,94 @@
...
+func SetupReplicator(
+	ctx context.Context,
+	conns dbconn.OrderedConns,
+	tables []tableverify.Result,
+	targetSchema tree.Name,
+	cdcCursor string,
+	logger zerolog.Logger,
+) error {
+	// If PG, we need to create a publication.
+	// For CRDB, we need to create a changefeed
```
ryanluu12345, line 40: Where does the binary setup on the cockroach node happen here?

fetch/replicator/replicator.go
```diff
@@ -0,0 +1,94 @@
...
+
+			if _, err := connTarget.Exec(ctx, "CREATE DATABASE IF NOT EXISTS _cdc_sink"); err != nil {
+				return err
+			}
+
+			logger.Info().Msg("seting zone ttl zone config")
+			if _, err := connTarget.Exec(ctx, "ALTER DATABASE _cdc_sink CONFIGURE ZONE USING gc.ttlseconds=300"); err != nil {
+				return err
+			}
+		} else { // If its not a CRDB conn then its a PG conn
```
ryanluu12345, line 68: For this else case couldn't it be mysql or oracle too?

fetch/dataexport/pg.go
Jeremyyang920: I talked with bram and for pg/mysql theres not much to tune really. The parallelism is the one thing he ssaid could maybe help. But thing is we will just have flag overload. We could optionally register cdc-sink flags if they want to run the mode. They are also going to clean up alot of the cdc sink flags soon.

fetch/fetch.go
Jeremyyang920: cdcsink start its. It creates a webhook server for the changefeed to send the messages to

fetch/replicator/replicator.go
Jeremyyang920: It doesn't right now. It starts on where you run fetch.

fetch/replicator/replicator.go
Jeremyyang920: We dont have oracle to -> crdb cdcsink right now if im not mistaken or its not in the wiki.
But mysql has no setup required. The pg case is to create the publications and stuff. Mysql just works with nothing else.

fetch/dataexport/pg.go
ryanluu12345: Ok that's good to know. Then maybe a `cdc-sink-parallelism` flag may make sense

fetch/fetch.go
ryanluu12345: Ok thanks for the context, makes sense!

fetch/replicator/replicator.go
ryanluu12345: Yeah that matches what I see on the wiki. I see Cockroach -> Oracle as an option. I think it will still be worthwhile to explicitly call out if it is Postgres, we do this specific setup. This will be less error prone if we need to add setup for other dialects later.

Jeremyyang920: If you read the block of code above it, its nested inside a `conns[0].(*dbconn.PGConn)` already. So inside it already has to be a pgConn. Then its in the else block of if `conns[0].(*dbconn.PGConn).IsCockroach()` so the only other possible conn is Pgconn.

Other dialects can't even enter the main if statement.

fetch/fetch.go
Jeremyyang920: If the check is used elsewhere sure, but making a helper for a single check used once imo isint that worth and just makes readability harder needing to scroll down to see what canRunCDCSink does and then go back up.

fetch/replicator/replicator.go
ryanluu12345: Ok this makes sense!

fetch/fetch.go
```diff
@@ -351,9 +383,89 @@ func Fetch(
...
+			return err
+		}
+
+		cmd, err := sqlSrc.ReplicatorCommand(binName, conns[1], conns[1].Database(), dbTables.Verified[0][1].Schema)
+		if err != nil {
+			return err
+		}
+
+		cmdErrChan := make(chan error)
+		go func() {
```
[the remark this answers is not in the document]
Jeremyyang920: This looks correct. CRDB mode just requires that you start cdcsink first and then create the changefeed after which this does with the cmd.Start

ryanluu12345:
@Jeremyyang920 , what ticket does this map to btw?

fetch/fetch.go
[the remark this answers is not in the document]
ryanluu12345: Yep, I think this change makes sense and makes running workflow the same for all the dialects. We get the added benefit of being able to add test-only code in between the `Start` and `Wait` commands.

Jeremyyang920:
> @Jeremyyang920 , what ticket does this map to btw?

Theres an new epic https://cockroachlabs.atlassian.net/browse/CC-27547
But i never really created any tasks yet. You can just create like a mysql, pg, and crdb individual task

ryanluu12345:
> > @Jeremyyang920 , what ticket does this map to btw?
> 
> Theres an new epic https://cockroachlabs.atlassian.net/browse/CC-27547 But i never really created any tasks yet. You can just create like a mysql, pg, and crdb individual task

@Jeremyyang920 thank you, I went ahead and added the relevant tasks to the epic. I'll knock them out one by one.

fetch/replicator/replicator.go
```diff
@@ -0,0 +1,98 @@
...
+			if _, err := conn.Exec(ctx, "DROP PUBLICATION IF EXISTS molt_fetch"); err != nil {
```
ryanluu12345, line 73: This `DROP` and `CREATE` can likely be in a transaction to make sure these both happen together or not at all.

fetch/dataexport/pg.go
```diff
@@ -141,3 +143,23 @@ func (p *pgSourceConn) Export(
...
+) (*exec.Cmd, error) {
+	if p.settings.PG.SlotName == "" {
+		return nil, errors.AssertionFailedf("slot_name must be set")
+	}
+	return exec.Command(
+		bin,
+		"pglogical",
+		"--sourceConn", p.conn.ConnStr(),
+		"--targetConn", target.ConnStr(),
+		"--publicationName", "molt_fetch",
```
ryanluu12345, line 158: @Jeremyyang920 , should this be configurable or is it almost always `molt_fetch`?

Jeremyyang920: Im indifferent. Since we are the ones creating it, it makes sense to name it what we created it as. DMS doesnt ask you for one either.

fetch/dataexport/crdb.go
```diff
@@ -67,3 +69,17 @@ func (c *crdbSourceConn) Export(
...
 	return c.conn.Close(ctx)
 }
+
+func (c *crdbSource) ReplicatorCommand(
+	bin string, target dbconn.Conn, db tree.Name, sc tree.Name,
+) (*exec.Cmd, error) {
+	return exec.Command(
+		bin,
+		"start",
+		"--bindAddr", "0.0.0.0:30004",
```
ZhouXing19, line 79: Not opposing what we've done here, but I'm surprised that C2C replication doesn't require a snapshot id for cdcsink. I'm reading https://github.com/cockroachdb/cdc-sink/wiki/C2C but also can't find any reference to that. If so, how can we ensure the replication starts from the ASOT timestamp?

fetch/dataexport/pg.go
```diff
@@ -141,3 +143,23 @@ func (p *pgSourceConn) Export(
...
+		"--targetSchema", fmt.Sprintf("%s.%s", db, sc),
+		"--slotName", fmt.Sprint(p.settings.PG.SlotName),
+		"--stagingSchema", fmt.Sprint(db),
```
ZhouXing19, line 161: nit: why target schema includes the sc, but we don't specify it for staging schema?

ZhouXing19 (approved):
LGTM!

fetch/dataexport/pg.go
ryanluu12345: Ok got it. Then we will just do it on behalf of the user as is already done here.

fetch/dataexport/crdb.go
ryanluu12345: This is a great question that I think @Jeremyyang920 would have more context on. However, from what I understand, this looks to be what is responsible for making sure it starts from the correct point:
```
if conns[0].IsCockroach() {
				tblQ := replicator.CreateTableList(tables)
				logger.Info().Msg("creating changefeed on the source CRDB database")
				q := fmt.Sprintf(`CREATE CHANGEFEED FOR TABLE %s
				INTO 'webhook-https://localhost:30004/%s/%s?insecure_tls_skip_verify=true'
				WITH updated, resolved='1s', min_checkpoint_frequency='1s', initial_scan='no', cursor='%s',
					 webhook_sink_config='{"Flush":{"Bytes":1048576,"Frequency":"1s"}}'`, tblQ, conns[1].Database(), dbTables.Verified[0][1].Schema, sqlSrc.CDCCursor())
				if _, err := conns[0].(*dbconn.PGConn).Exec(ctx, q); err != nil {
					cmdErrChan <- err
					return
				}
			}
```
For context, we first start the CDC command as a backgrounded task. Then we create the changefeed, keying off the cursor: `sqlSrc.CDCCursor()`, which CRDB should give us. I'd imagine instead of using AOST, it uses this cursor to know where to replicate from. @Jeremyyang920 to chime in more when he's back

fetch/dataexport/pg.go
ryanluu12345: Note these docs from CDC sink: https://github.com/cockroachdb/cdc-sink/wiki/PGLogical
```
--stagingSchema atom             a SQL database schema to store metadata in (default _cdc_sink.public)
--targetSchema atom              the SQL database schema in the target cluster to update
```
I think it may just be the naming on cdc-sink side is a bit imprecise in what it requires. The `targetSchema` needs the database and the schema. Whereas the staging schema only needs the db name.

fetch/dataexport/crdb.go
ZhouXing19: Yes, i think that's the case. Thanks for explanation!

PR #210, opened 2024-04-09
fetch: fix bug for 'null' string for a json/jsonb column

ZhouXing19:
Release note: fix bug for 'null' string for a json/jsonb column


e2e/testdata/pg/json_null.ddt
```diff
@@ -0,0 +1,228 @@
+exec silent
+psql 'postgres://postgres@localhost:5432/defaultdb' -c "DROP TABLE IF EXISTS nulljson; CREATE TABLE nulljson (id SERIAL PRIMARY KEY, json_data json NOT NULL DEFAULT '{}'::json); "
+----
+
+exec silent
+psql 'postgres://postgres@localhost:5432/defaultdb' -c "INSERT INTO nulljson VALUES (1, 'null'); "
+----
+
+exec expect-error
+psql 'postgres://postgres@localhost:5432/defaultdb' -c "INSERT INTO nulljson VALUES (2, 'NULL'); "
```
Jeremyyang920, line 10: Is this just a pg semantics thing where the casing matters? Inserting null vs NULL results in different behavior?

ZhouXing19: yes, pg only accept `null`, case sensitive.

pgconv/datum_converter.go
```diff
@@ -160,15 +161,25 @@ func convertNumeric(val pgtype.Numeric) (*tree.DDecimal, error) {
...
+) (tree.Datums, error) {
 	ret := make(tree.Datums, len(vals))
 	if len(vals) != len(typOIDs) {
 		return nil, errors.AssertionFailedf("val length != oid length: %v vs %v", vals, typOIDs)
 	}
 	for i := range vals {
 		var err error
-		if ret[i], err = ConvertRowValue(typMap, vals[i], typOIDs[i]); err != nil {
-			return nil, err
+		if (typOIDs[i] == pgtype.JSONOID || typOIDs[i] == pgtype.JSONBOID) && bytes.Equal(rawVals[i], []byte("null")) {
```
Jeremyyang920, line 173: is []byte{"null") the only case that we need to check for?

Jeremyyang920 (approved):
Just some small questions but overall lgtm! Great finds.

pgconv/datum_converter.go
ZhouXing19: Yes i think so, since pg only accept 'null' as the valid input for a json column

PR #215, opened 2024-04-11
update image

ZhouXing19:


ZhouXing19:
Alright we need to set the cluster settings to bring the ci time down.

PR #219, opened 2024-04-15
docker: consolidated docker compose setup to the docker directory

ryanluu12345:
Previously, we were maintaining two sets of database configurations for our manual and CI testing workflows. This led to duplication and drift between the two environments. We now consolidate so that it's simpler to run any type of testing and have the docker compose configs transparent.

Resolves: CC-27869
Release Note: None

**Testing**
The CI testing results here will prove out that things work.

Local Testing
Largely the same with the exception of the connection string needed.
```
--source "mysql://user:password@localhost/molt"
```
Instead of `root:password`

**Impact of consolidation**
On a fresh pull from Docker repos, this takes 40 seconds in latest CI run, versus 34 seconds in a previous run. This is an acceptable amount of delay caused by consolidating.

**TODO**

- [x] Conditional seeding of data


ryanluu12345:
Ahh I see what I need to dig into: the delays caused by data seeding are probably leading to this issue. I should probably do conditional seeding. Let me look into how I can do that.

Makefile
```diff
@@ -23,3 +23,9 @@ run_e2e_tests:
 
 rewrite_e2e_tests:
 	go test -timeout 600s -run TestDataDriven github.com/cockroachdb/molt/e2e -e2e-enabled --rewrite
+
+run_compose_local:
```
ZhouXing19, line 27: Just to confirm if i understand it correctly: 
- `docker-compose ./docker/docker-compose.yml` is now a set of light-weight container -- they are not bound to local volumns, nor have init scripts to run. So they are fast to set up and tear down. 

- `{run|down}_compose_local` is based on the `docker-compose ./docker/docker-compose.yml` but with init scripts and enhanced with observability.

so the former are "bare bone" containers to suit the tests, but the later are a more comprehensive, wrapped containers. 

I like this implementation, but what is `run_compose_local` is used for? For local manual tests only?

ZhouXing19 (approved):

Makefile
ryanluu12345: Yep the `./docker/docker-compose.yml` is lightweight and the bare minimum needed to run CI. It makes it so there are minimal changes for E2E and unit testing.

The `compose_local` setup is meant to be a local debugging and testing environment that allows you to run fetch on a pre-seeded dataset. So a use case is: if you want to test a new feature e2e without writing automated tests. I routinely do this when I'm building out new features. So yes, local manual tests on a dataset we use often.

ZhouXing19: Cool, thanks for explaining!

PR #221, opened 2024-04-17
cdc_sink_test: added first cut of CDC sink integration testing

ryanluu12345:
This change supports CDC sink integration testing for PostgreSQL, MySQL, and CockroachDB as sources. The test runs all the setup on the source and target tables, does the initial data load via fetch and then spins up the CDC sink routine and verifies that a rows is properly replicated on the target. Additionally, added infrastructural and Github workflow changes to support this testing in our CI environment: mainly to pull the correct CDC sink latest version.

Resolves: CC-27870
Release Note: None


dbconn/pg.go
```diff
@@ -39,11 +42,17 @@ func ConnectPGConfig(ctx context.Context, id ID, cfg *pgx.ConnConfig) (*PGConn,
 	if err := conn.QueryRow(ctx, "SELECT version()").Scan(&version); err != nil {
 		return nil, err
 	}
+	// TODO: review Jane's diff and make more robust.
```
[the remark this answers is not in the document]
ryanluu12345: @ZhouXing19 is the last instance of `defaultdb` or database name guaranteed to correspond to the database. I think this may lead to more issues down the line actually if the user has a string where they pass in postgres options and it's the same name as their database somehow.

Should we just limit the affected database names to `defaultdb` or the test database name? So slight change to your diff and replace the `defaultdb`

Side note: really wishing there were a way to turn a pg conn string config back to the string instead of relying on the original one.

fetch/cdc_sink_test.go
```diff
@@ -0,0 +1,250 @@
...
+	"github.com/cockroachdb/molt/utils"
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	createSourceTableStmt        = `CREATE TABLE cdcsinktbl1 (id INT PRIMARY KEY, t TEXT);`
+	insertInitialSourceTableStmt = `INSERT INTO cdcsinktbl1 VALUES (1, 'aaa'), (2, 'bb b'), (3, 'ééé'), (4, '🫡🫡🫡'), (5, '娜娜'), (6, 'Лукас'), (7, 'ルカス');`
+	insertCDCSourceTableStmt     = `INSERT INTO cdcsinktbl1 VALUES (8, 'cdc sink added this');`
+	numRetries                   = 8
```
Jeremyyang920, line 26: I don't mind the number here but why 8 specifically?

fetch/cdc_sink_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+		if _, err := testutils.ExecConnQuery(ctx, insertInitialSourceTableStmt, conns[0]); err != nil {
+			return err
+		}
+
+		// Need a slight wait between creating the table and starting fetch
+		// in order to address a timing issue for AOST for CRDB, when
+		// the AOST can get trimmed and it misses the CREATE/INSERT table here.
+		if conns[0].IsCockroach() {
+			time.Sleep(time.Second)
```
Jeremyyang920, line 48: Is 1 second truly enough here and is there a btter way to check this so we dont flake.

fetch/cdc_sink_test.go
```diff
@@ -0,0 +1,250 @@
...
+		return nil
+	}
+
+	var defaultPollingFunc = func(conns dbconn.OrderedConns, sentinelFileName string, sigint chan os.Signal) (err error) {
+		defer func() {
+			t.Logf("sending termination signal to fetch")
+			sigint <- syscall.SIGTERM
+		}()
+
+		t.Logf("polling for the row to be written to the target")
```
Jeremyyang920, line 60: Is this log line in the right order? It says polling but the next one is writing the row. Should the polling log come after?

fetch/cdc_sink_test.go
ryanluu12345: No specific reason. It was an arbitrary one. I was also playing around with retries.

fetch/cdc_sink_test.go
ryanluu12345: It's something that @ZhouXing19 and I looked at together. We think it's because if it's within the same second it gets trimmed. So the logic here is that if we delay at least a second, we shouldn't trim down.

Not sure if there is a better way since by the time we receive the marshalled value from the DB in the struct, it's already truncated. @ZhouXing19 for any comments here.

fetch/cdc_sink_test.go
ryanluu12345: Yeah, you're correct here. I'll move it down a few lines

fetch/cdc_sink_test.go
ZhouXing19: I think one second should be solid enough as the molt AOST takes the now() timestamp and rounds it to the second precision.

ryanluu12345: @Jeremyyang920 per Jane's comment, I'm going to go ahead and just do one second here. We shall see the runs in Ci. But I think the reasoning is sound.

fetch/cdc_sink_test.go
```diff
@@ -0,0 +1,250 @@
...
+		sourceConnClone, err := conns[0].Clone(context.Background())
+		require.NoError(t, err)
+
+		t.Logf("writing row to the source side of the database")
+		if _, err := testutils.ExecConnQuery(ctx, insertCDCSourceTableStmt, sourceConnClone); err != nil {
+			require.NoError(t, err)
+		}
+
+		result := ""
+		for i := 0; i < numRetries; i++ {
```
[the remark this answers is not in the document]
ryanluu12345: Done! Good suggestion.

Jeremyyang920 (approved):

PR #223, opened 2024-04-19
fetch/schema-creation: consider dropped column for c2c case

ZhouXing19:
Previously, dropped column from a crdb source are wrongly included when we fetching the columns when creating a new schema for target. Now we exclude those dropped columns.

Release note (bug fix): fixed bug with dropped column from a crdb source when creating new schema


Jeremyyang920 (approved):
This is good for now, but curious if we need to also check how mysql performs with a dropped column.

ZhouXing19:
> but curious if we need to also check how mysql performs with a dropped column.

Added tests for mysql and pg too, it seems to be working fine.

PR #224, opened 2024-04-24
fetch: omit continuation token logging in the event that the export did not finish yet

ryanluu12345:
Previously, we were outputting continuation token details at any point of the initial data load for fetch. This can provide confusing behavior because we are not able to continue from the token if the export fails or partially completes. Now, we ensure that we fully complete the export before we allow continuations.

**Important note**
I'm going for an all or nothing approach here. If even one table does not have their data fully exported, we can't continue anyways. That token is invalid because the token will continue from the relevant file for every table.

Resolves: CC-27460
Release Note: Logging for the continuation token is now omitted in the case that the export of data does not successfully complete.

**Tests**
Failed at CDC sink -> no token
```
{"level":"debug","command":"/Users/ryanluu/go/src/github.com/cockroachdb/molt/cdc-sink start --bindAddr 0.0.0.0:30004 --tlsSelfSigned --disableAuthentication --targetConn postgres://root@localhost:26258/molt?sslmode=disable -v","time":"2024-04-23T17:44:21-07:00","message":"executing CDC sink command"}
{"level":"info","time":"2024-04-23T17:44:21-07:00","message":"creating changefeed on the source CRDB database"}
{"level":"error","time":"2024-04-23T17:44:21-07:00","message":"error while running CDC sink"}
{"level":"info","time":"2024-04-23T17:44:21-07:00"}
```

Failed at Export -> no token
```
{"level":"info","type":"summary","num_tables":1,"cdc_cursor":"2024-04-24T00:58:59Z","time":"2024-04-23T17:58:59-07:00","message":"starting fetch"}
{"level":"warn","time":"2024-04-23T17:58:59-07:00","message":"clearing all continuation tokens because running in clear continuation mode"}
{"level":"info","time":"2024-04-23T17:59:00-07:00","message":"data extraction phase starting"}
Error: failed at export
```

Failed at import -> token shown:
```
y retention job id"}
{"level":"debug","hr_job":"962887600707338241","time":"2024-04-23T17:59:36-07:00","message":"exiting history retention management go routine"}
{"level":"info","time":"2024-04-23T17:59:36-07:00","message":"history retention job has finished"}
{"level":"info","fetch_id":"d2614250-45db-4988-af3b-be073aa975ee","time":"2024-04-23T17:59:36-07:00","message":"continue from this fetch ID"}
Error: failed after export
```

**Follow Up**
Created this ticket to track the work needed to do a `do what you can approach` for the way we do exports. This PR reminded us that right now we have an `all or nothing approach` to exports.
https://cockroachlabs.atlassian.net/browse/CC-28032


Jeremyyang920 (approved):

fetch/fetch.go
```diff
@@ -244,13 +244,18 @@ func Fetch(
 		}
 	}()
 
-	madeItPastInitialDataLoad := false
+	finishedExport := false
+	finishedDataLoad := false
```
Jeremyyang920, line 248: nit: You can in line declare using var since default for bool is already false.

PR #225, opened 2024-04-24
fetch: fixed data race for finishedExport

ryanluu12345:
finishedExport is set inside of concurrent goroutines. Because of this, we must atomatically set and read this value. Therefore, we change it to an atomic.Bool.

Release Note: None

Testing
```
{"level":"info","time":"2024-04-24T10:48:11-07:00","message":"timestamp is protected for 24h0m0s: timestamp: 1713980891"}
{"level":"info","type":"summary","num_tables":1,"cdc_cursor":"2024-04-24T17:48:11Z","time":"2024-04-24T10:48:11-07:00","message":"starting fetch"}
{"level":"warn","time":"2024-04-24T10:48:11-07:00","message":"clearing all continuation tokens because running in clear continuation mode"}
Error: failed at export
```


ryanluu12345:
> I'm wondering if it is correct to have a single `finishedExport` bool var to determine the global completion of export for all tables.
> 
> Say we have 3 tables exporting in parallel. Table 1 and 2 finished, atomically set `finishedExport = true`. And, while table 3 hasn't finished exporting, we somehow terminated the process of fetch and hence enter the defer, which check if `finishedExport == ture`. In our case, in reality, export hasn't finished for all table, but the defer would see `finishedExport == true`, which iiuc is not correct.
> 
> I think a better way that also avoid datarace is to have a `statusMarker := make([]bool, numTable)`, and for each goroutine, just mark the corresponding entry in this list to true (`i.e. statusMarker[i] = true`). And in the defer, iterate the whole list to see if all entries are true.

This would be the correct behavior if we wait on all the statuses, but given that we use an error group, it's already an `all or nothing` right? So if even one fails and didn't make it past export, we would return inside the `g.Do` and then cancel the other goroutines.

fetch/fetch.go
```diff
@@ -389,7 +390,18 @@ func Fetch(
 		}
 	}()
 
-	if err := g.Wait(); err != nil {
+	err = g.Wait()
+	// Check if export finished for each table.
+	tableToExportStatus.Range(func(key, value any) bool {
```
ZhouXing19, line 395: nit: since finishedExport is read in the defer only, i think we can just replace `finishedExport` with `tableToExportStatus`, and move this check with `Range()` in the defer. This would also handle the case where we enter the defer before `g.Wait()` returns.

ZhouXing19 (approved):
LGTM!

fetch/fetch.go
ryanluu12345: Done!

PR #226, opened 2024-04-25
dataexport: add replicator flag parsing logic and helpers

ryanluu12345:
Added in utilities to parse the replicator flags and assemble them into the correct mappings so that they can be combined, overrided, etc. with the default settings.

Drilled in the utilities to the replicator command method so that it can merge the overrided flags.

Resolves: CC-27981
Release Note: Replicator flags are now supported which allows users to configure the replicator command to their use case.


fetch/dataexport/crdb.go
```diff
@@ -120,17 +120,26 @@ func (c *crdbSourceConn) Close(ctx context.Context) error {
...
+	bin string, target dbconn.Conn, db tree.Name, sc tree.Name, cdcSinkArgs string,
 ) (*exec.Cmd, error) {
-	return exec.Command(
+	defaultArgs := map[string]string{
+		"-v": "",
+	}
+	additionalFlagList, err := getFlagList(defaultArgs, cdcSinkArgs)
+	if err != nil {
+		return nil, err
+	}
```
Jeremyyang920, line 131: Instead of redclaring defaultArgs in each dialect, I think we should create the base defaultArgs in dataexport.go and then that will be initialized by default for use in the other dialects. Similar pattern almost to the redaction map logic.

ryanluu12345: Yep, I implemented it that way based on your suggestion. Def less duplication. I see what you mean

fetch/dataexport/mysql.go
```diff
@@ -103,23 +103,30 @@ func (m *mysqlConn) Close(ctx context.Context) error {
...
+	additionalFlagList, err := getFlagList(DefaultReplicatorFlags, cdcSinkArgs)
+	if err != nil {
+		return nil, err
+	}
+
+	cmdArgs := []string{
 		bin,
 		"mylogical",
-		"--sourceConn", mysqlurl.CfgToConnStr(mysqlConfig, true)+"?sslmode=disable",
+		"--sourceConn", mysqlurl.CfgToConnStr(mysqlConfig, true) + "?sslmode=disable",
```
Jeremyyang920, line 121: Oh hmm. I just thought about something that may require some testing. If we have a mysql db instance that requires ssl already. The cfg.ToConnstr already has the ssl stuff in it. We might not be able to do the sslmod=disable.

Jeremyyang920: I think the correct behavior is just passing the conn string as is since its the same we we needed to connect for fetch.

ryanluu12345: So are you saying just take out the string addition for `sslmode=disable`? The sslmode=disable can be passed by the caller. Wouldn't it be redundant right now if someone passed in the `sslmode=disable` manually?

ryanluu12345: I just tested it without that SSL line:
```
ERROR  [Apr 25 16:50:02] exited                                        detail="invalid sslmode: \"\"\ngithub.com/cockroachdb/cdc-sink/internal/source/mylogical.(*Config).Preflight\n\t/home/runner/work/cdc-sink/cdc-sink/internal/source/mylogical/config.go:180\ngithub.com/cockroachdb/cdc-sink/internal/source/mylogical.ProvideEagerConfig\n\t/home/runner/work/cdc-sink/cdc-sink/internal/source/mylogical/provider.go:116\ngithub.com/cockroachdb/cdc-sink/internal/source/mylogical.Start\n\t/home/runner/work/cdc-sink/cdc-sink/internal/source/mylogical/wire_gen.go:39\ngithub.com/cockroachdb/cdc-sink/internal/cmd/mylogical.Command.func1\n\t/home/runner/work/cdc-sink/cdc-sink/internal/cmd/mylogical/mylogical.go:35\ngithub.com/cockroachdb/cdc-sink/internal/util/stdlogical.New.func1\n\t/home/runner/work/cdc-sink/cdc-sink/internal/util/stdlogical/stdlogical.go:112\ngithub.com/spf13/cobra.(*Command).execute\n\t/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:983\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1115\ngithub.com/spf13/cobra.(*Command).Execute\n\t/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1039\ngithub.com/spf13/cobra.(*Command).ExecuteContext\n\t/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1032\nmain.main\n\t/home/runner/work/cdc-sink/cdc-sink/main.go:141\nruntime.main\n\t/opt/hostedtoolcache/go/1.22.1/x64/src/runtime/proc.go:271\nruntime.goexit\n\t/opt/hostedtoolcache/go/1.22.1/x64/src/runtime/asm_arm64.s:1222" error="invalid sslmode: \"\""
```
We have to specify it in the conn string, it appears.

Jeremyyang920: Yea it would be. But im more so thinking the case, they passed in sslmode stuff and then we tack on a sslmode=disable.

ryanluu12345: I tested it with passing in sslmode from the fetch command, and now I see this error:
```
Use "molt fetch [command] --help" for more information about a command.

Error 1193 (HY000): Unknown system variable 'sslmode'
exit status 1
```
This is now failing the first DB connection request to load connections for fetch. I wonder if it's how we do the parsing.

ryanluu12345: This is the resulting conn string, which looks right to me though... unless sslmode is not supported as a param in this scheme: `user:password@tcp(localhost:3306)/molt?sslmode=disable`

ryanluu12345: Actually, I think this may be a good solution to this problem:
```
	sourceConn := mysqlurl.CfgToConnStr(mysqlConfig, true)
	if mysqlConfig.TLS == nil {
		sourceConn += "?sslmode=disable"
	}

	cmdArgs := []string{
		bin,
		"mylogical",
		"--sourceConn", mysqlurl.CfgToConnStr(mysqlConfig, true),
		"--targetConn", target.ConnStr(),
		"--defaultGTIDSet", c.gtid,
		"--targetSchema", fmt.Sprintf("%s.%s", db, sc),
		"--stagingSchema", fmt.Sprint(db),
		"-v",
	}
```

ryanluu12345: Yep, this works when no SSL mode is specified. I'll do some testing to see if SSL mode = another mode works.

ryanluu12345: @Jeremyyang920 does this solution work for you?

Jeremyyang920: Oh yea in user:password@tcp(localhost:3306)/molt?sslmode=disable more, the argument is TLS and not sslmode. We created our version of the parser originally so we could parse conn strings much more easily like in. a PG way. Idk why mysql made things so difficult.

The solution looks fine i think since cdc-sink takes in the pg style string.

ryanluu12345: Ok sounds good. @Jeremyyang920 , could I get an approval here then since the solution works for you?

fetch/dataexport/crdb.go
```diff
@@ -120,17 +120,23 @@ func (c *crdbSourceConn) Close(ctx context.Context) error {
 }
 
 func (c *crdbSource) ReplicatorCommand(
-	bin string, target dbconn.Conn, db tree.Name, sc tree.Name,
+	bin string, target dbconn.Conn, db tree.Name, sc tree.Name, cdcSinkArgs string,
```
Jeremyyang920, line 123: can you do a check through the repo and find where we use cdcSink instead of replicator? Lets keep things consistent across the board. Even though i konw the binary is still cdc-sink.

Jeremyyang920 (approved):
LGTM once you do the naming check in the comment.

fetch/dataexport/crdb.go
ryanluu12345: Done!

PR #227, opened 2024-04-25
cmd/fetch: add logic to warn user about potential inconsistency when running MySQL with parallelism and concurrency

ryanluu12345:
MySQL has limitations with consistency when working with cross-transaction consistency. The purpose of this change is to enforce sane defaults when running to MySQL: running with no inter-table and intra-table parallelism, while allowing for concurrency if folks want. However in that case, we alert them of documentation that will allow them to prepare their environment so consistency could be achieved even with parallelism.

Resolves: CC-28114
Release Note: MySQL now supports a consistent default mode that does not leverage any parallelism. There is also new user text that alerts the user of the case when fetch with MySQL may not be consistent and shows documentation.


ryanluu12345:
@taroface , just wanted to note that this is a feature that will need heavy documentation since it concerns data consistency for using MOLT fetch with MySQL. We documented some of this during our team meeting today. We'll create a separate follow up DOCS ticket for you. Right now, I just want to get some feedback on the copy here. Pretty much I want to convey that we set sane defaults that allow fetch to be consistent when using MySQL. In the case that users override flags, we want to note that it could potentially be consistent and to check our docs for more details of how they can remediate that or see what we can guarantee.

cmd/fetch/fetch.go
```diff
@@ -446,3 +457,15 @@ or as an automated job. If not set, prompts user for confirmation before perform
 
 	return cmd
 }
+
+func handleMySQLConcurrencyFlags(logger zerolog.Logger, sourceConn dbconn.Conn, cfg *fetch.Config) {
+	if _, ok := sourceConn.(*dbconn.MySQLConn); ok {
+		if !cfg.IsExportConcurrencySet && !cfg.IsTableConcurrencySet {
+			logger.Warn().Msg("defaulting export concurrency and table concurrency to 1 in order to guarantee data consistency for MySQL")
```
taroface, line 464: Is this actually setting a default, or just setting both to 1? Not sure if the nuance matters.

```suggestion
			logger.Warn().Msg("defaulting export concurrency and table concurrency to 1 in order to guarantee data consistency for MySQL")
```

taroface (commented):
2 initial comments/questions ... I probably want to work on these suggestions a bit more

cmd/fetch/fetch.go
ZhouXing19: > Is this actually setting a default, or just setting both to 1?

It's both -- the default for both export and table concurrency are 1, if the dialect is MySQL.

ZhouXing19 (approved):
LGTM, thanks!

cmd/fetch/fetch.go
ryanluu12345: Yep, what Jane mentioned. We want to disallow any parallelism at all otherwise we could be inconsistent

cmd/fetch/fetch.go
```diff
@@ -446,3 +457,15 @@ or as an automated job. If not set, prompts user for confirmation before perform
...
+			cfg.Shards = 1
+			cfg.TableConcurrency = 1
+		} else if (cfg.IsTableConcurrencySet || cfg.IsExportConcurrencySet) && (cfg.TableConcurrency > 1 || cfg.Shards > 1) {
+			logger.Warn().Msgf("table concurrency or export concurrency specified and either is greater than 1, this can lead to inconsistency. Read the docs for details on how to run in this mode consistently: %s", mysqlDocsLink)
```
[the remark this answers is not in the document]
taroface: Edited my suggestion - does this make sense?

It would apply to all identical blocks in this PR, ofc.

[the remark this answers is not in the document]
ryanluu12345: Loooks good to me!

cmd/fetch/fetch.go
ryanluu12345: If I'm not mistaken @taroface , your suggestion is exactly the same as the original right?

cmd/fetch/fetch.go
```diff
@@ -446,3 +457,15 @@ or as an automated job. If not set, prompts user for confirmation before perform
...
 }
+
+func handleMySQLConcurrencyFlags(logger zerolog.Logger, sourceConn dbconn.Conn, cfg *fetch.Config) {
+	if _, ok := sourceConn.(*dbconn.MySQLConn); ok {
+		if !cfg.IsExportConcurrencySet && !cfg.IsTableConcurrencySet {
+			logger.Warn().Msg("defaulting export concurrency and table concurrency to 1 in order to guarantee data consistency for MySQL")
+			cfg.Shards = 1
+			cfg.TableConcurrency = 1
+		} else if (cfg.IsTableConcurrencySet || cfg.IsExportConcurrencySet) && (cfg.TableConcurrency > 1 || cfg.Shards > 1) {
+			logger.Warn().Msgf("table or export concurrency is greater than 1. This can lead to inconsistency when migrating MySQL data. For details on how to maintain consistency when using --table-concurrency and --export-concurrency: %s", mysqlDocsLink)
```
taroface, line 468:
```suggestion
			logger.Warn().Msgf("table concurrency or export concurrency is greater than 1. This can lead to inconsistency when migrating MySQL data. For details on maintaining consistency when using --table-concurrency and --export-concurrency: %s", mysqlDocsLink)
```

taroface (approved):
I thought of some further tweaks to clarify/make readable. Otherwise this LGTM once those are in (and if you agree)!

cmd/fetch/fetch.go
ryanluu12345: Addressed! We should be good!

PR #230, opened 2024-05-01
schema-creation: support multi-region tables

ZhouXing19:
We now support multiregion tables for the crdb-crdb case.

It is required for a successful move for a `REGIONAL BY ROW` table that the support regions of the source db should be the subset of the ones of the target table.

Clause for locality, such as `LOCALITY REGIONAL BY TABLE` and `LOCALITY REGIONAL BY ROW` are appended to the create table statement. For `REGIONAL BY ROW` (RBR) tables, we don't explicitly create the `crdb_region` implicit column, but just let `LOCALITY REGIONAL BY ROW` to invoke the creation of this column implicitly.

Also specified the schema for tables, meaning 2 tables with the same names but from different schema should not interfere with each other.

Release note: support cockroachdb multi-region tables for fetch.


ryanluu12345:
I think it would be helpful to describe the exact way in which you're handling multi-region tables in the PR description, so people can know what to expect. An example would be: "We now support multi-region tables now by ....."

fetch/schema_creation.go
```diff
@@ -207,28 +216,34 @@ FROM (
         NOT a.attnotnull AS nullable,
     CASE
         WHEN a.attname IN (
-            SELECT column_name
-            FROM information_schema.constraint_column_usage
+            SELECT ccu.column_name
```
ryanluu12345, line 219: It may be worthwhile to add a comment to this query to identify the key parts of it:
- Selecting column names for non-hidden columns
- Selecting the region locality details
- etc.

fetch/schema_creation.go
```diff
@@ -392,8 +440,9 @@ ORDER BY
 	return res, nil
 }
 
+// TODO(janexing): could fail with mysql.
```
ryanluu12345, line 443: Why could this fail with MySQL?

fetch/schema_creation.go
```diff
@@ -418,32 +467,173 @@ func GetTypeConvMap(
...
+	secondaryRegion bool
+}
+
+func maybeCheckTableLocality(
+	ctx context.Context, logger zerolog.Logger, conn dbconn.Conn, table *dbtable.DBTable,
+) error {
+	if !conn.IsCockroach() {
+		return nil
+	}
+	newConn, err := conn.Clone(ctx)
```
ryanluu12345, line 487: Are you cloning this connection because it can use the connection struct concurrently?

fetch/schema_creation.go
ZhouXing19: Oh this is extraneous comments I used to remind myself, removing.

fetch/schema_creation.go
ZhouXing19: I added some comments, but I think it's already pretty straightforward with the alias or name of result from the SELECT statements.

ZhouXing19:
Added more explanatory text in commit message and pr description.

e2e/testdata/pg/type_override.ddt
```diff
@@ -94,39 +94,7 @@ fetch --type-map-file='./testdata/pg/mapping/manytypes.json' --source 'postgres:
...
-{"level":"warn","message":"the original type of column id is integer, while overriding map has source type inet, so igoring the overriding map"}
-{"level":"info","message":"received overlapping map rule for column manytypes.name: {inet:{family:StringFamilyStringFamily width:0 precision:0 locale:\"\" visible_type:0 oid:25 time_precision_is_set:false}}"}
-{"level":"warn","message":"the original type of column name is character varying(100), while overriding map has source type inet, so igoring the overriding map"}
-{"level":"info","message":"received overlapping map rule for column manytypes.age: {inet:{family:StringFamilyStringFamily width:0 precision:0 locale:\"\" visible_type:0 oid:25 time_precision_is_set:false}}"}
-{"level":"warn","message":"the original type of column age is integer, while overriding map has source type inet, so igoring the overriding map"}
-{"level":"info","message":"received overlapping map rule for column manytypes.height: {inet:{family:StringFamilyStringFamily width:0 precision:0 locale:\"\" visible_type:0 oid:25 time_precision_is_set:false}}"}
```
Jeremyyang920, line 103: Was all of the red here and below meant to be rewritten? Or did we just change the log level

fetch/schema_creation.go
```diff
@@ -47,6 +47,12 @@ func (cs columnsWithType) CRDBCreateTableStmt(
 	includePkForEachCol := len(pkList) <= 1
 
 	for _, col := range cs {
+		if table.LocalityExpr == multiRegionRBRLocality && col.columnName == multiRegionHiddenColName {
+			// We don't explicitly create `crdb_region` table for RBR tables but let it be created
+			// automatically via the `LOCALITY REGIONAL BY ROW` clause.
+			logger.Info().Msgf("skipping creating hidden default column %s for table %s has locality %s", multiRegionHiddenColName, table.String(), table.LocalityExpr)
```
Jeremyyang920, line 53:
```suggestion
			logger.Info().Msgf("skipping creation of the hidden default column %s for table %s with locality %s", multiRegionHiddenColName, table.String(), table.LocalityExpr)
```
I think that makes the log a bit cleaner.

fetch/schema_creation.go
```diff
@@ -418,32 +468,158 @@ func GetTypeConvMap(
...
+func MaybeCheckMultiRegionLocality(
+	ctx context.Context, logger zerolog.Logger, conns dbconn.OrderedConns, table *dbtable.DBTable,
+) (checked bool, err error) {
+	srcConn := conns[0]
+	tgtConn := conns[1]
+	if !srcConn.IsCockroach() {
+		return false, nil
+	}
+	if !tgtConn.IsCockroach() {
+		return false, errors.AssertionFailedf("source is cockroachdb while target is not cocktoachdb")
```
Jeremyyang920, line 506:
```suggestion
		return false, errors.AssertionFailedf("source is cockroachdb while target is not cockroachdb")
```

e2e/testdata/pg/type_override.ddt
ZhouXing19: Actually, good call! I forgot to change the table name to include the schema in `manytypes.json` so the overriding json was ignored. I now updated it to `public.manytypes`

fetch/schema_creation.go
```diff
@@ -418,32 +468,158 @@ func GetTypeConvMap(
...
 	if err != nil {
 		return "", errors.Wrapf(err, "failed get columns for target table: %s", table.String())
 	}
 
 	var res string
+
 	for _, col := range newCols {
 		if col.udtDefinition != "" {
+			if col.columnName == multiRegionHiddenColName && table.LocalityExpr == multiRegionRBRLocality {
+				logger.Info().Msgf("skipping creating enum for hidden default column %s for table %s has locality %s", multiRegionHiddenColName, table.String(), table.LocalityExpr)
```
Jeremyyang920, line 609:
```suggestion
				logger.Info().Msgf("skipping creation of the enum for hidden default column %s, for table %s, with locality %s", multiRegionHiddenColName, table.String(), table.LocalityExpr)
```

Jeremyyang920 (approved):
Great find and fix!

ryanluu12345 (approved):

PR #231, opened 2024-05-01
fetch: adapt MySQL connection string parsing logic to consider TLS parameters

ryanluu12345:
(See linked ticket for details on the investigation and resolution)

Previously, the custom ssl flags that are used (i.e. sslmode, sslcert) were leading to errors upon interacting with the database, because they are invalid system variables (aka query parameters for the MySQL DSN). The issue is traced back to how we do not discard the TLS parameters after we use them to configure a custom TLS object in the code.

This change primarily makes the following changes so that we can support TLS for MySQL:
1. Store a new field in the MySQLConn object that keeps track of the TLS parameters
2. Delete the TLS parameters from the original configuration's parameter map so that it doesn't show up in the resultant connection string
3. Adapt the replicator command for MySQL to consider the TLS parameters before it passes along the string to the replicator

Resolves: CC-28145
Release Note: None

**Testing**
1. Setup an RDS instance that requires SSL from all connections
2. Ran fetch with `--ongoing` replication specified and a connection string that includes `sslmode=sslmode=verify-ca&sslrootcert=global-bundle.pem
```
{"level":"info","type":"summary","fetch_id":"e939d787-c94a-4924-a7eb-ff6b20313fd4","num_tables":1,"tables":["public.employees"],"cdc_cursor":"60dd9e2f-074a-11ef-b98f-12a76f81d567:1-44","net_duration_ms":4991.374083,"net_duration":"000h 00m 04s","time":"2024-04-30T19:14:59-07:00","message":"fetch complete"}
{"level":"info","time":"2024-04-30T19:14:59-07:00","message":"starting replicator"}
INFO   [Apr 30 19:14:59] cdc-sink starting                             -buildmode=exe -compiler=gc -ldflags="-s -w" CGO_ENABLED=0 GOARCH=arm64 GOOS=darwin vcs=git vcs.modified=true vcs.revision=beb976aefb34e09649fd6a945938dc49f3badf14 vcs.time="2024-04-03T18:30:30Z"
INFO   [Apr 30 19:15:00] Version info: 8.0.35                         
[2024/04/30 19:15:00] [info] binlogsyncer.go:156 create BinlogSyncer with config {10 mysql rluu-mysql-test.cpa6lrp2ahsc.us-east-1.rds.amazonaws.com 3306 admin    false false 0x140000bc1a0 false UTC false 0 0s 0s 0 false false 0 <nil> 0x14000500f00}
[2024/04/30 19:15:00] [info] binlogsyncer.go:410 begin to sync binlog from GTID set 60dd9e2f-074a-11ef-b98f-12a76f81d567:38-39
```


dbconn/mysql.go
```diff
@@ -22,13 +24,33 @@ func ConnectMySQL(ctx context.Context, id ID, connStr string) (*MySQLConn, error
...
+	}
+
+	return tlsMap
+}
+
+func processTLSParam(cfg *mysql.Config, tlsMap map[string]string, paramName string) {
+	if v, ok := cfg.Params[paramName]; ok {
+		tlsMap[paramName] = v
+		delete(cfg.Params, paramName)
+	}
```
Jeremyyang920, line 53: I see what you are trying to do here, but I think case, I this this is an insance where splitting this out makes it worse from a readability standpoint of having to jump to another function. This isint used anywhere else either and is only 3 lines so I think it just belongs in the loop above.

dbconn/mysql.go
```diff
@@ -22,13 +24,33 @@ func ConnectMySQL(ctx context.Context, id ID, connStr string) (*MySQLConn, error
...
 		return nil, err
 	}
 	m := pgtype.NewMap()
-	return &MySQLConn{id: id, url: u, DB: db, typeMap: m, database: tree.Name(cfg.DBName)}, nil
+	return &MySQLConn{id: id, url: u, DB: db, typeMap: m, database: tree.Name(cfg.DBName), tlsMap: tlsMap}, nil
+}
+
+func handleTLSParams(cfg *mysql.Config) map[string]string {
+	tlsParams := []string{"sslmode", "sslrootcert", "sslcert", "sslkey"}
+	tlsMap := map[string]string{}
```
Jeremyyang920, line 40: nit: make(map[strint]string{},4) since we know thats the max number of keys from tlsParams. We don't have to do a map resize even if small.

fetch/dataexport/mysql.go
```diff
@@ -110,6 +110,22 @@ func (c *mysqlSource) ReplicatorCommand(
...
+	mysqlConn, ok := c.conn.(*dbconn.MySQLConn)
+	if !ok {
+		return nil, errors.New("failed to cast to a MySQL connection")
+	}
+
+	// Add TLS parameters back to the configuration to pass to CDC sink.
+	if mysqlConfig.Params == nil {
+		// In the case that TLS map is also nil or an empty map
+		// no extra parameters will be added to the connection string anyways.
+		mysqlConfig.Params = mysqlConn.TLSMap()
```
Jeremyyang920, line 122: What is the need to do this?

dbconn/mysql.go
ryanluu12345: Yeah, makes sense. I've gone ahead and merged them back together.

fetch/dataexport/mysql.go
ryanluu12345: When we do the parse to config from the connection string and there are no parameters set, the `Params` field can be nil. Therefore, we want to set it to something valid (i.e. the TLSMap)

dbconn/mysql.go
ryanluu12345: Makes sense. Updated

ryanluu12345:
@Jeremyyang920 tested again and it looks good to me!

fetch/dataexport/mysql.go
Jeremyyang920: Does it break if params is nil? Those should all be optional anyways.

Jeremyyang920 (approved):

fetch/dataexport/mysql.go
ryanluu12345: It panics, yes

PR #233, opened 2024-05-02
molt-nightly.sh: support multiple version testing for scale tests

ryanluu12345:
Previously, we only tested on a hardcoded version of CRDB. Now, we support testing for the latest alpha version, which will help to catch regressions, along with testing for LTS versions that we specify we support on our CRDB website.

Resolves: [CC-27411](https://cockroachlabs.atlassian.net/browse/CC-27411)
Release Note: None

Test Run: https://teamcity.cockroachdb.com/buildConfiguration/Internal_Molt_MoltNightly/15091578?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true


build/teamcity/internal/molt/test-variants.sh
```diff
@@ -0,0 +1,31 @@
...
+# Add new case statements for new variants
+case $VARIANT in
+  base_aws)
+    $DIR/setup-scale-test.sh --cluster-name migrations-nightly-base-aws --cluster-version "$CLUSTER_VERSION" --source "$SOURCE_CONN" -n 4 --cloud aws --zones us-east-1a --machine-type m6idn.2xlarge --aws-machine-type-ssd m6idn.2xlarge \
+    --fetch-args "--non-interactive --bucket-path 's3://migrations-fetch-ci-test/scale-test' --table-handling 'drop-on-target-and-recreate' --table-filter 'test_table_large' --export-concurrency 4" \
+    --verify-args "--table-splits 2"
+    ;;
+
+  scaled_aws)
+    $DIR/setup-scale-test.sh --cluster-name migrations-nightly-scaled-aws --cluster-version "$CLUSTER_VERSION" --source "$SOURCE_CONN" -n 4 --cloud aws --volume-size 1500 --zones us-east-1a --aws-cpu-options 'CoreCount=8,ThreadsPerCore=2' --machine-type m6idn.4xlarge --aws-machine-type-ssd m6idn.4xlarge \
```
ZhouXing19, line 20: Not blocking but just out of curiosity -- does the number of nodes (indicated by the `-n` flag) is supposed to be >= export concurrency? Or maybe it doesn't matter?

Jeremyyang920: The -n only has implications for the IMPORT part since it distributes the files to each available node. Concurrency only matters on the actual machine you run fetch on as that drives cpu usage.

ZhouXing19: I see, thanks!

Jeremyyang920, line 20: Are we running both variants for all versions?

ryanluu12345: No, just the scaled version. Take a look at `molt-nightly.sh`. That's where I invoke only the scaled test. The reason is that it runs much faster.

Jeremyyang920 (approved):

PR #234, opened 2024-05-02
fetch: added behavior to exit fetch if columns mismatch on the target

ryanluu12345:
Previously, there was an edge case where fetch could end up failing if there was a mismatching column definition and the mismatching column had a not null constraint. This behavior alerts the user after a potentially long export happens that the fetch cannot continue. Instead, we let the user know early that there is a mismatching column.

Resolves: CC-27398
Release Note: None

**Test**
```
go run . fetch \
  --source 'mysql://user:password@localhost/molt' \
  --target 'postgres://root@localhost:26257/molt?sslmode=disable' \
  --table-handling 'truncate-if-exists' \
  --bucket-path 's3://molt-test/your-path' \
  --ongoing-replication \
  --table-filter "mismatch"  --cleanup --non-interactive
...
{"level":"error","mismatch_tag":"source_not_null","failable_mismatch":false,"reason":"column name NOT NULL mismatch: true vs false","time":"2024-05-14T13:19:36-07:00","message":"not migrating column from public.mismatch as it mismatches on the target side"}
{"level":"error","mismatch_tag":"column_type","failable_mismatch":true,"reason":"column type mismatch on address: int4 vs inet","time":"2024-05-14T13:19:36-07:00","message":"not migrating column from public.mismatch as it mismatches on the target side"}
{"level":"error","mismatch_tag":"target_not_null","failable_mismatch":true,"reason":"column age NOT NULL mismatch: false vs true","time":"2024-05-14T13:19:36-07:00","message":"not migrating column from public.mismatch as it mismatches on the target side"}
{"level":"error","mismatch_tag":"target_extra_column","failable_mismatch":false,"reason":"extraneous column extra found","time":"2024-05-14T13:19:36-07:00","message":"not migrating column from public.mismatch as it mismatches on the target side"}
{"level":"error","mismatch_tag":"target_missing_column","failable_mismatch":true,"reason":"missing column missing_on_target","time":"2024-05-14T13:19:36-07:00","message":"not migrating column from public.mismatch as it mismatches on the target side"}
Error: exiting early due to mismatching column definitions that must be manually resolved. For details, see logs containing "mismatch_tag".
```


ryanluu12345:
Open questions here:
- What happens if the differences are just because the target has a constraint the source does not?
- What happens if the types are not coercible at all?
- What is the better behavior here? 1. Failing at the import and potentially having the ability to continue after fixing target schema. 2. Failing before fetch even runs an instructing the user to fix mismatches in schema (if source has not null and target doesn’t, this will fail)

The more I think about this change, the more edge cases I think this introduces. It may be too aggressive of an approach since any difference in column definition, on any table, will render fetch unable to start.

However, having specific cases where fetch exits on column mismatch may lead to incomplete data migration or failure at a later stage (import) instead of at the beginning.

ZhouXing19:
> What happens if the differences are just because the target has a constraint the source does not?

If they are using the built-in schema creation they should not hit this case. The target can only miss constraints from the source. If it's because the user created the schema themselves, i think it's up to them to fix the schema (drop constraints on target or etc.)

> What is the better behavior here? 1. Failing at the import and potentially having the ability to continue after fixing target schema. 2. Failing before fetch even runs an instructing the user to fix mismatches in schema (if source has not null and target doesn’t, this will fail)

I lean towards 2.

ryanluu12345:
Discussed offline with the team. Here are some notes and path forward:
- We want to move the verify results error logic handling before any of the `fetchTable` and concurrent logic runs
- Remove the verify table results check inside of the `fetchTable` function
- Research and compile list of all the constraint and type mismatches that can occur: null/not null, wrong type entirely, etc.

**Naive Solution**
- If any mismatch, just exit early
- If they want to keep going even though there is a mismatch, provide a flag to override

**Nuanced Solution**
- If any mismatch, tag the mismatch reason: specific constraint issues, type mismatch, etc.
- Create a mismatch reason mapping that tells fetch when it should exit
- Research how to tag verify mismatch failures

ryanluu12345:
After discussing with the team, I think we are going with nuanced solution because we can enumerate the types of verify mismatches we could run into. Here are the associated mappings:
1. Truth is missing a primary key: exit early
2. Extraneous column found: can still continue (or maybe we want this to exit right now?)
3. Nullable on source but not null on target: exit early
4. Column type mismatch (not in allowable types): exit early
5. Missing column: exit early
6. PRIMARY KEY does not match source of truth (columns and types must match): exit early
7. Collation for primary key does not match: exit early? could also continue if we update collations right?
8. Unique constraint on target: can still continue (because can be dropped during migration and reapplied)
9. Check constraint on target: can still continue (because can be dropped during migration and reapplied)
10. FK constraint: can still continue (because can be dropped during migration and reapplied)

fetch/fetch.go
```diff
@@ -536,6 +545,40 @@ func Fetch(
...
+			logger.Error().
+				Str("mismatch_tag", col.Tag).
+				Bool("failable_mismatch", isFailableMismatch).
+				Str("reason", col.Info).
+				Msgf("not migrating column from %s.%s as it mismatches on the target side", col.Name.Schema, col.Name.Table)
+		}
+	}
+
+	if hasFailableMismatch {
+		return errors.Newf(`exiting early due to column definition mismatches. See above logs with "mismatch_tag" field for details.`)
```
ryanluu12345, line 572: @taroface , wondering if I could have some copy help here?

fetch/fetch.go
```diff
@@ -536,6 +545,40 @@ func Fetch(
...
+	}
+
+	if hasNonFailableMismatch {
```
ryanluu12345, line 575: @taroface here too

fetch/fetch.go
taroface, line 572:
```suggestion
		return errors.Newf(`exiting early because mismatching column types could not be automatically converted. For details, see the logs that contain "mismatch_tag".`)
```

taroface (commented):
Left 2 suggestions

fetch/fetch.go
ryanluu12345: Addressed this!

fetch/fetch.go
```diff
@@ -536,6 +545,40 @@ func Fetch(
...
+		logger.Warn().Msg("only non-failable mismatches were found, fetch continuing")
```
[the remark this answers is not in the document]
ryanluu12345: Addressed this

fetch/fetch.go
```diff
@@ -554,6 +563,40 @@ func Fetch(
...
+				Msgf("not migrating column from %s.%s as it mismatches on the target side", col.Name.Schema, col.Name.Table)
+		}
+	}
+
+	if hasFailableMismatch {
+		return errors.Newf(`exiting early due to mismatching column definitions that must be manually resolved. For details, see logs containing "mismatch_tag".`)
+	}
+
+	if hasNonFailableMismatch {
+		logger.Warn().Msg(`found only mismatching column definitions that do not block data migration. For details, see logs containing "mismatch_tag". Fetch continuing`)
```
taroface, line 594:
```suggestion
		logger.Warn().Msg(`found only column mismatches that do not block migration. For details, see logs with "mismatch_tag". Fetch continuing. `)
```

e2e/testdata/pg/type_override.ddt
```diff
@@ -181,38 +181,14 @@ verify --source 'postgres://postgres@localhost:5432/defaultdb' --target 'postgre
...
-{"level":"info","message":"verifying common tables"}
-{"level":"info","message":"establishing snapshot"}
-{"level":"info","type":"summary","num_tables":1,"cdc_cursor":"0/19E3610","message":"starting fetch"}
-{"level":"warn","message":"clearing all continuation tokens because running in clear continuation mode"}
-{"level":"warn","reason":"column type mismatch on is_active: bool vs inet","message":"not migrating column {public manytypes} as it mismatches"}
-{"level":"info","message":"data extraction phase starting"}
-{"level":"info","type":"summary","num_rows":100,"export_duration_ms":1000,"export_duration":"000h 00m 01s","message":"data extraction from source complete"}
-{"level":"info","message":"cleaning up resources created during fetch run"}
-{"level":"info","type":"summary","fetch_id":"0000000000","num_tables":1,"tables":["public.manytypes"],"cdc_cursor":"0/19E3610","net_duration_ms":1000,"net_duration":"000h 00m 01s","message":"fetch complete"}
+Error: exiting early due to mismatching column definitions that must be manually resolved. For details, see logs containing "mismatch_tag".
```
taroface, line 184:
```suggestion
Error: exiting early due to mismatching column definitions that must be manually resolved. For details, see logs with "mismatch_tag".
```

taroface (approved):
LGTM; i suggested some ways to shorten these further.

Jeremyyang920:
> After discussing with the team, I think we are going with nuanced solution because we can enumerate the types of verify mismatches we could run into. Here are the associated mappings:
> 
> 1. Truth is missing a primary key: exit early
> 2. Extraneous column found on target: can still continue
> 3. Nullable on source but not null on target: exit early
> 4. Nullable on target but not null on source: can still continue (all rows should be properly migrated)
> 5. Column type mismatch (not in allowable types): exit early
> 6. Missing column: exit early
> 7. PRIMARY KEY does not match source of truth (columns and types must match): exit early
> 8. Collation for primary key does not match: exit early? could also continue if we update collations right?
> 9. Unique constraint on target: can still continue (because can be dropped during migration and reapplied)
> 10. Check constraint on target: can still continue (because can be dropped during migration and reapplied)
> 11. FK constraint: can still continue (because can be dropped during migration and reapplied)

Are we going to have each of these documented?

fetch/fetch.go
Jeremyyang920, line 594: Do we have an example of this in a DDT? And what exactly do we mean in column definitions that do not block. I think we may want to be more clear here?

taroface:
> > After discussing with the team, I think we are going with nuanced solution because we can enumerate the types of verify mismatches we could run into. Here are the associated mappings:
> > 
> > 1. Truth is missing a primary key: exit early
> > 2. Extraneous column found on target: can still continue
> > 3. Nullable on source but not null on target: exit early
> > 4. Nullable on target but not null on source: can still continue (all rows should be properly migrated)
> > 5. Column type mismatch (not in allowable types): exit early
> > 6. Missing column: exit early
> > 7. PRIMARY KEY does not match source of truth (columns and types must match): exit early
> > 8. Collation for primary key does not match: exit early? could also continue if we update collations right?
> > 9. Unique constraint on target: can still continue (because can be dropped during migration and reapplied)
> > 10. Check constraint on target: can still continue (because can be dropped during migration and reapplied)
> > 11. FK constraint: can still continue (because can be dropped during migration and reapplied)
> 
> Are we going to have each of these documented?

Yes, @ryanluu12345 made a ticket earlier today: https://cockroachlabs.atlassian.net/browse/DOC-10284

Is documenting this a GA requirement?

Jeremyyang920:
> > > After discussing with the team, I think we are going with nuanced solution because we can enumerate the types of verify mismatches we could run into. Here are the associated mappings:
> > > 
> > > 1. Truth is missing a primary key: exit early
> > > 2. Extraneous column found on target: can still continue
> > > 3. Nullable on source but not null on target: exit early
> > > 4. Nullable on target but not null on source: can still continue (all rows should be properly migrated)
> > > 5. Column type mismatch (not in allowable types): exit early
> > > 6. Missing column: exit early
> > > 7. PRIMARY KEY does not match source of truth (columns and types must match): exit early
> > > 8. Collation for primary key does not match: exit early? could also continue if we update collations right?
> > > 9. Unique constraint on target: can still continue (because can be dropped during migration and reapplied)
> > > 10. Check constraint on target: can still continue (because can be dropped during migration and reapplied)
> > > 11. FK constraint: can still continue (because can be dropped during migration and reapplied)
> > 
> > 
> > Are we going to have each of these documented?
> 
> Yes, @ryanluu12345 made a ticket earlier today: https://cockroachlabs.atlassian.net/browse/DOC-10284
> 
> Is documenting this a GA requirement?

I don't think its GA requirement since we have the logs to support it. Could be a followup. I think getting the main gist of everything else is probably more important.

ryanluu12345:
> > > > After discussing with the team, I think we are going with nuanced solution because we can enumerate the types of verify mismatches we could run into. Here are the associated mappings:
> > > > 
> > > > 1. Truth is missing a primary key: exit early
> > > > 2. Extraneous column found on target: can still continue
> > > > 3. Nullable on source but not null on target: exit early
> > > > 4. Nullable on target but not null on source: can still continue (all rows should be properly migrated)
> > > > 5. Column type mismatch (not in allowable types): exit early
> > > > 6. Missing column: exit early
> > > > 7. PRIMARY KEY does not match source of truth (columns and types must match): exit early
> > > > 8. Collation for primary key does not match: exit early? could also continue if we update collations right?
> > > > 9. Unique constraint on target: can still continue (because can be dropped during migration and reapplied)
> > > > 10. Check constraint on target: can still continue (because can be dropped during migration and reapplied)
> > > > 11. FK constraint: can still continue (because can be dropped during migration and reapplied)
> > > 
> > > 
> > > Are we going to have each of these documented?
> > 
> > 
> > Yes, @ryanluu12345 made a ticket earlier today: https://cockroachlabs.atlassian.net/browse/DOC-10284
> > Is documenting this a GA requirement?
> 
> I don't think its GA requirement since we have the logs to support it. Could be a followup. I think getting the main gist of everything else is probably more important.

Agreed with Jeremy. I went ahead and documented the behavior in the ticket. Given how verify mismatch currently works, some of the more detailed cases are not currently included in there. However, because for unique, check or FK constraints we can still continue, the net behavior is as expected. Let me note in the ticket.

fetch/fetch.go
ryanluu12345: The column definitions that do not block are specified in the logging. Basically the extraneous column or other constraints on the target. 

It's redundant to specify here. I think people should just look at the `mismatch_tag` logging.

e2e/testdata/pg/non_failing_column_mismatch.ddt
```diff
@@ -0,0 +1,63 @@
...
+----+---+-------
+(0 rows)
+
+fetch --source 'postgres://postgres@localhost:5432/defaultdb' --target 'postgres://root@localhost:26257/defaultdb?sslmode=disable' --table-filter 'mismatch_tbl1' --local-path /tmp/basic --local-path-listen-addr '0.0.0.0:9005' --non-interactive
+----
+{"level":"info","message":"default compression to GZIP"}
+{"level":"info","message":"checking database details"}
+{"level":"info","source_table":"public.mismatch_tbl1","target_table":"public.mismatch_tbl1","message":"found matching table"}
+{"level":"info","message":"verifying common tables"}
+{"level":"error","mismatch_tag":"target_extra_column","failable_mismatch":false,"reason":"extraneous column extra found","message":"not migrating column from public.mismatch_tbl1 as it mismatches on the target side"}
```
ryanluu12345, line 45: @Jeremyyang920 , this is what the non-failing mismatch logging looks like in ddt format.

taroface: In that case though maybe "do not block migration" shouldn't be printed after all, since the error says that the data actually wasn't migrated. We could just remove that part for simplicity.

ryanluu12345: Makes sense, since they don't make their way to the target anyways. What do you think is an appropriate edit here @taroface ?

fetch/fetch.go
```diff
@@ -554,6 +563,40 @@ func Fetch(
...
+				Msgf("not migrating column from %s.%s as it mismatches on the target side", col.Name.Schema, col.Name.Table)
+		}
+	}
+
+	if hasFailableMismatch {
+		return errors.Newf(`exiting early due to mismatching column definitions that must be manually resolved. For details, see logs with "mismatch_tag".`)
+	}
+
+	if hasNonFailableMismatch {
+		logger.Warn().Msg(`found only column mismatches that do not block migration. For details, see logs with "mismatch_tag". Fetch continuing.`)
```
taroface, line 594: I feel like this is simplest, unless it is important to say that "only" non-failing mismatches were found.

```suggestion
		logger.Warn().Msg(`found column mismatches. For details, see logs with "mismatch_tag". Fetch continuing.`)
```

e2e/testdata/pg/non_failing_column_mismatch.ddt
taroface: Actually, I misread that error. It seems that all data is still migrated, just not extraneous source columns, correct? If so, I'm ok with my current suggestion.

ryanluu12345: Yes, that's correct. All the other columns still migrate over. So it doesn't block the overall migration job.

taroface (approved):
Copy lgtm

ZhouXing19 (approved):

PR #236, opened 2024-05-03
schema-creation: log generic error only

ZhouXing19:
Release note: None


fetch/schema_creation.go
```diff
@@ -171,7 +171,7 @@ func (t *columnWithType) CRDBColDef(
 			logger.Warn().Msgf("type [%s] for column %s: %s", t.dataType, t.Name(), sc.ShortDescription)
 		}
 		if err != nil {
-			log.Err(err)
+			log.Err(errors.Newf("failed getting crdb type for schema creation"))
 			return nil, err
```
Jeremyyang920, line 175: I think what you can do is just `return nil, errors.Newf("failed getting crdb type for schema creation")`

Jeremyyang920: We would have already seen all the other failed logs above, so here we just return the generic error and exit using that. Since if you just log, err does not get overwritten and you returnt he last instance of err.

ZhouXing19: I see, done

Jeremyyang920 (approved):

PR #238, opened 2024-05-06
[Daft] fetch: add lock for mysql concurrent export 

ZhouXing19:
DNM


ZhouXing19:
This PR is RFAL -- added lock related flag to ask explicitly specify the options to lock tables or not.

ZhouXing19:
Discussed offline -- we will keep this open and merged it upon request by customer.

cmd/fetch/fetch.go
```diff
@@ -440,6 +443,17 @@ or as an automated job. If not set, prompts user for confirmation before perform
 		"Whether docker is used for the database backend, which will change the behavior of certain URL pathing (i.e. changefeed URL)",
 	)
 
+	cmd.PersistentFlags().Var(
```
ryanluu12345, line 446: Is this a flag that we want to publicize? Or should it only be on a `need to know basis` after they consult with our team? Kind of like how we have the `allow-cockroach-replication` flag for an as needed or testing basis.

cmd/fetch/fetch.go
```diff
@@ -471,14 +485,29 @@ If a rule has specified column "*", all columns with matching source type will a
...
+func handleMySQLConcurrencyFlags(
+	logger zerolog.Logger, sourceConn dbconn.Conn, cfg *fetch.Config,
+) error {
 	if _, ok := sourceConn.(*dbconn.MySQLConn); ok {
 		if !cfg.IsExportConcurrencySet && !cfg.IsTableConcurrencySet {
 			logger.Warn().Msg("defaulting export concurrency and table concurrency to 1 in order to guarantee data consistency for MySQL")
 			cfg.Shards = 1
 			cfg.TableConcurrency = 1
 		} else if (cfg.IsTableConcurrencySet || cfg.IsExportConcurrencySet) && (cfg.TableConcurrency > 1 || cfg.Shards > 1) {
+			if cfg.LockTableMode == fetch.NoneLockTables {
```
ryanluu12345, line 497: Ok understand the behavior of your other PR now. So we only allow concurrency to be above 1 in the case that lock table mode is specified.

e2e/testdata/crdb-multi-region/wrong_regions.ddt
```diff
@@ -30,58 +30,5 @@ database |  region   | primary | secondary |    zones
 
 fetch --source 'postgres://root@localhost:26257/newdb?sslmode=disable' --target 'postgres://root@localhost:26267/newdb?sslmode=disable' --table-filter 'newtable' --local-path /tmp/basic --local-path-listen-addr '0.0.0.0:9005' --non-interactive --table-handling 'drop-on-target-and-recreate' --cleanup=true --table-concurrency=1
 ----
-----
 Error: failed checking multi-region consistency between source and target databases: source crdb contains region us-east-2 but target crdb doesn't contain
```
ryanluu12345, line 33: How come the help text is no longer printed out? Do we hide it now?

e2e/testdata/pg/type_override.ddt
```diff
@@ -217,61 +217,7 @@ is_active
 
 fetch --type-map-file='./testdata/pg/mapping/doesnt_exist.json' --source 'postgres://postgres@localhost:5432/defaultdb' --target 'postgres://root@localhost:26257/defaultdb?sslmode=disable' --table-filter 'manytypes' --direct-copy --table-handling 'drop-on-target-and-recreate'  --cleanup=true --non-interactive
 ----
-----
 Error: failed to get the overriding type map for schema creation: failed to read json file ./testdata/pg/mapping/doesnt_exist.json for type mapping: open ./testdata/pg/mapping/doesnt_exist.json: no such file or directory
-Usage:
```
ryanluu12345, line 222: Same question here about removed help text? If we are removing help text from E2E tests, should we do this in a separate commit + PR? This affects all tests

fetch/dataexport/dataexport.go
```diff
@@ -55,6 +56,24 @@ type Settings struct {
 
 	CRDBPTSExtensionFreq     time.Duration
 	CRDBPTSExtensionLifetime time.Duration
+
+	mySQLExportCfg mySQLExportConfig
+}
+
+type mySQLExportConfig struct {
```
ryanluu12345, line 63: Worth storing which lock table mode it is in?

fetch/fetch.go
```diff
@@ -269,6 +272,41 @@ func Fetch(
...
+			loggerWithTableConcurrency := logger.With().Int("table-concurrency", cfg.TableConcurrency).Logger().With().Int("export-concurrency", cfg.Shards).Logger()
+			unlockTablesFunc, err = dataexport.LockTablesForMySQL(ctx, conns[0], loggerWithTableConcurrency, tables)
+			if err != nil {
+				return errors.Wrapf(err, "failed to lock tables for concurrent table export")
+			}
+			tablesLocked.Store(true)
+		}
+	}
+
+	defer func() {
```
ryanluu12345, line 304: This defer is only really relevant if the `cfg.LockTableMode == LockTables` right? Why not put this after `tablesLocked.Store(true)`. Since at that point, you are guaranteed to have both a non-nil `unlockTablesFunc` and the `tablesLocked` will be true then. 

In what other case could this logic be triggered? In the case where it doesn't enter the conditional, retErr will just be whatever value the error was in the function anyways.

fetch/fetch.go
```diff
@@ -269,6 +272,41 @@ func Fetch(
 		logger.Warn().Msgf("no table(s) to be imported, exiting molt fetch")
 		return nil
 	}
+
+	exportSettings := cfg.ExportSettings
```
ryanluu12345, line 276: Do we want to do a check for if tables are already locked? I guess broader question is: if tables are already locked, what should we do here? Attempt to lock? Exit and let them know things are already locked, or?

e2e/testdata/crdb-multi-region/wrong_regions.ddt
ryanluu12345: Disregard this, I recognize this from the PR you just merged in.

PR #242, opened 2024-05-08
fetch: fix bug for history retention and decimal

ZhouXing19:
jira:
- https://cockroachlabs.atlassian.net/browse/CC-28309
- https://cockroachlabs.atlassian.net/browse/CC-28310

Release note: fix bug for history retention and decimal


fetch/dataexport/dataexport.go
```diff
@@ -80,6 +81,10 @@ func scanWithRowIterator(
 	table rowiterator.ScanTable,
 ) error {
 	cw := moltcsv.NewWriter(writer)
+	_, ok := c.(*dbconn.PGConn)
+	if ok {
+		pgxdecimal.Register(c.TypeMap())
+	}
```
Jeremyyang920, line 87: Do we need to register the type map twice? Since we register it pg.go as well.

ZhouXing19: Removed the one in NewPgSource. We have to declare it in the most immediate usage.

Jeremyyang920 (approved):
Another rabbit hole but LGTM

ZhouXing19:
> Another rabbit hole but LGTM

🐰

PR #243, opened 2024-05-09
cmd: disallow setting table/export-concurrency for mysql over 1

ZhouXing19:
We strictly only allow single txn export for mysql source. No parallization is allowed.

For more context, concurrent export for mysql require locking the tables to guarantee consistency. Per the conversation offline, we only support no-concurrent mode for mysql source, and will pursue the concurrent solution only after being explicit asked by a customer.

Release note: strictly only allow setting table/export-concurrency for mysql to 1.


fetch/fetch.go
```diff
@@ -137,7 +144,18 @@ func Fetch(
 		cfg.FlushSize = blobStore.DefaultFlushBatchSize()
 	}
 	if cfg.TableConcurrency == 0 {
-		cfg.TableConcurrency = 4
+		if isMySQL {
+			cfg.TableConcurrency = 1
+		} else {
+			cfg.TableConcurrency = 4
+		}
+	}
```
Jeremyyang920, line 152: Do we need to do this since we do the change in the helper above?

ZhouXing19: Do you mean `handleMySQLConcurrencyFlags`? I think this is still needed since for unit tests, `Fetch` is not called via command line, so just add a sanity check here.

ryanluu12345:
Is it worth explaining why we are completely disallowing this now? Even though before we warned? I think the release notes or a comment in the code would be helpful for some context.

fetch/fetch.go
ryanluu12345: That's a good point. It's at a different layer, so you still need to be able to set through tests.

ZhouXing19:
> Is it worth explaining why we are completely disallowing this now? Even though before we warned? I think the release notes or a comment in the code would be helpful for some context.

Updated the PR description

ZhouXing19:
RFAL, thanks!

cmd/fetch/fetch.go
```diff
@@ -471,14 +470,17 @@ If a rule has specified column "*", all columns with matching source type will a
...
+	logger zerolog.Logger, sourceConn dbconn.Conn, cfg *fetch.Config,
+) error {
 	if _, ok := sourceConn.(*dbconn.MySQLConn); ok {
 		if !cfg.IsExportConcurrencySet && !cfg.IsTableConcurrencySet {
 			logger.Warn().Msg("defaulting export concurrency and table concurrency to 1 in order to guarantee data consistency for MySQL")
 			cfg.Shards = 1
 			cfg.TableConcurrency = 1
 		} else if (cfg.IsTableConcurrencySet || cfg.IsExportConcurrencySet) && (cfg.TableConcurrency > 1 || cfg.Shards > 1) {
-			logger.Warn().Msgf("table concurrency or export concurrency is greater than 1. This can lead to inconsistency when migrating MySQL data. For details on maintaining consistency when using --table-concurrency and --export-concurrency: %s", mysqlDocsLink)
+			return fetch.NoConcurrencyAllowedForMySQLErr
```
ryanluu12345, line 482: So we are not even documenting the multiple options they have to remain consistent? This just restricts the inconsistent mode from even happening?

cmd/fetch/fetch_test.go
```diff
@@ -39,6 +40,19 @@ func TestHandleMySQLConcurrencyFlags(t *testing.T) {
...
+			args: args{
+				sourceConn: &dbconn.MySQLConn{},
+				cfg: &fetch.Config{
+					TableConcurrency:       1,
+					Shards:                 1,
+					IsTableConcurrencySet:  true,
+					IsExportConcurrencySet: true,
+				},
+			},
+			expectedLoggerOutput: ``,
```
ryanluu12345, line 54: Nit, could probably make this just `""`

cmd/fetch/fetch_test.go
```diff
@@ -49,9 +63,7 @@ func TestHandleMySQLConcurrencyFlags(t *testing.T) {
 					IsTableConcurrencySet: true,
 				},
 			},
-			expectedLoggerOutput: `{"level":"warn","message":"table concurrency or export concurrency is greater than 1. This can lead to inconsistency when migrating MySQL data. For details on maintaining consistency when using --table-concurrency and --export-concurrency: https://www.cockroachlabs.com/docs/stable/molt-fetch"}
-`,
-			expectedConfig: &fetch.Config{TableConcurrency: 1, Shards: 4},
+			expectedErrStr: fetch.NoConcurrencyAllowedForMySQLErr.Error(),
```
ryanluu12345, line 66: Clean!

fetch/fetch.go
```diff
@@ -87,6 +87,11 @@ type Config struct {
 	ReplicatorFlags string
 }
 
+// TODO: update with mysql docs links for these cases.
+const mysqlDocsLink = "https://www.cockroachlabs.com/docs/stable/molt-fetch"
+
+var NoConcurrencyAllowedForMySQLErr = errors.AssertionFailedf("for mysql source export, only accept 1 for table concurrency and shard concurrency. See also: %s", mysqlDocsLink)
```
ryanluu12345, line 93: What do you intend for these documents to say? We no longer support concurrency greater than 1 for MySQL now. So is it even worth writing out the different options?

cmd/fetch/fetch.go
ZhouXing19: I noted in the doc request (the google doc) that if the user would like to try concurrent mode, contact the migrations team.

ryanluu12345 (approved):
Code looks good. Just unsure what the docs will now look like. Let's discuss during team weekly

fetch/fetch.go
ZhouXing19: IIUC we will elaborate more in the doc.

ZhouXing19:
Merging for now, if there's follow-up from the team weekly I'll update the jira ticket for the doc.

cmd/fetch/fetch_test.go
ZhouXing19: it's actually the same as "".

PR #244, opened 2024-05-09
fetch: use shared ctx for parallel table fetch attempts

ZhouXing19:
To enable concurrent fetch with multiple tables, we lauch new goroutine for each table. The context used for each goroutine doesn't interfere each other, meaning if one goroutine fails, the others will still continue. That can result
in hanging even when fetching one table sees an error.

This commit is to ensure that all goroutines now use the child context generated by the errgroup. When one goroutine returns an error, the child context is automatically cancelled for all other goroutines, thus we will immidiately exit the whole fetch process and emit the error.

## Test

The test is to tell such a story:

We try to fetch 2 tables concurrently. Table 1's import is forced to fail, and table 2 runs a long query at either sharding or import stage. We expect that when table 1 is forced to fail, the long running query in table 2's goroutine will exit immediately with a `context cancelled` error.

Release note: None


fetch/testdata/pg/continuation-edge-cases.ddt
```diff
@@ -1,30 +1,30 @@
 # Testing for duplicate key errors.
 exec all
-CREATE TABLE tbl1(id INT PRIMARY KEY, t TEXT)
+CREATE TABLE tbl7(id INT PRIMARY KEY, t TEXT)
```
ZhouXing19, line 3: Changes for several ddt files are merely updating the table names so that they don't overlap each other.

Somehow with testing knobs added, different tests seem to interfere with each other, though we clean the database for each test.

fetch/fetch.go
```diff
@@ -375,8 +375,8 @@ func Fetch(
 	}
 
 	workCh := make(chan tableverify.Result)
-	g, _ := errgroup.WithContext(ctx)
-	for i := 0; i < cfg.TableConcurrency; i++ {
+	g, tableConcurrentCtx := errgroup.WithContext(ctx)
+	for i := 0; i < min(cfg.TableConcurrency, len(tables)); i++ {
```
ryanluu12345, line 379: May be worth adding a comment that specifies what this behavior guards against: a blocking channel read, if I'm not mistaken

dbconn/dbconn.go
```diff
@@ -35,6 +35,16 @@ func RegisterTelemetry(conns OrderedConns) error {
 	return nil
 }
 
+type DBConnDialect string
+
+const (
```
ryanluu12345, line 40: What is the motivation of bundling this change in this PR? In my opinion, this renaming and reusability change could go in a separate dependency PR.

fetch/import_table.go
```diff
@@ -84,15 +85,17 @@ func reportImportTableProgress(
...
-	conn, err := pgx.ConnectConfig(ctx, pgConn.Config())
-	if err != nil {
-		return err
-	}
-	defer conn.Close(ctx)
-
 	prevVal := 0.0
 
 	if err := r.Do(func() error {
+		if ctxErr := ctx.Err(); ctxErr != nil {
```
ryanluu12345, line 91: Could you explain a bit why you're moving this in side the retry logic? Is it because the connection could be reused and lead to a `conn busy` error?

fetch/testdata/pg/continuation-edge-cases.ddt
ryanluu12345: Hmm, I think we need to revisit our database clean logic. I remember running into this issue during some other testing, but I don't think we properly connect to the database and clean it. It may just be cleaning `defaultdb` every time.

dbconn/dbconn.go
ryanluu12345: Not a huge deal. I see why it's here. You can just keep in this PR.

verify/tableverify/tableverify.go
```diff
@@ -332,7 +332,7 @@ func allowableTypes(a oid.Oid, b oid.Oid) bool {
 	if a == oid.T_uuid && b == oid.T_varchar {
 		return true
 	}
-	if a == oid.T_int2 && b == oid.T_bool {
+	if a == oid.T_bool && b == oid.T_int2 {
```
ryanluu12345, line 335: Should we be able to support both when source and target are int2 and we try to convert with bool>

fetch/fetch.go
```diff
@@ -691,11 +691,11 @@ func fetchTable(
 
 			jobManagementWg = sqlSrc.HistoryRetentionJobManagement(ctx, logger, exportFinishedChan, extensionFreq, extensionCntPtr, cfg.TestOnly, &testingKnobs)
 
-			wg, _ := errgroup.WithContext(ctx)
+			wg, shardsCtx := errgroup.WithContext(ctx)
```
ryanluu12345, line 694: Do we run into the same hanging behavior here if a shard export fails?

fetch/testdata/pg/continuation-edge-cases.ddt
ZhouXing19: all tables are created within `defaultdb` so cleaning `defaultdb` should actually solve it. I think it might the clean db code might just have returned without waiting for the db got completely drained. I'm not sure, but it doesn't seem to be a hard blocker and would like to defer it till a future PR

fetch/import_table.go
ZhouXing19: It's because this will lead to a panic within the pgx lib if the context had been cancelled (in the case where the other goroutine failed). So we precheck to avoid a panic case.

verify/tableverify/tableverify.go
ZhouXing19: We are supporting both direction. See line above. Originally this is redundant to one of the existing cases.

fetch/fetch.go
ZhouXing19: Yes

fetch/fetch.go
ZhouXing19: Added comment inline.

fetch/import_table.go
ryanluu12345: Got it, to avoid panicking. Yeah I agree to be more proactive here.

fetch/import_table.go
```diff
@@ -248,6 +289,7 @@ func importWithBisect(
 		// Otherwise do nothing. If the len is more than 1, we have
 		// not bottomed out yet on a leaf node.
 		if err != nil && len(curr) > 1 {
+			// TODO(jane): should we check if this is a context cancelled error and early exit?
```
ryanluu12345, line 292: Is this still applicable? If context is cancelled wouldn't it just cancel the goroutine?

fetch/testdata/pg/constraint-exceptions.ddt
```diff
@@ -63,10 +63,25 @@ INSERT INTO dependent VALUES (1, 0, 'aaa')
...
+query target
+SELECT table_name, message, sql_state, file_name, stage FROM _molt_fetch_exceptions WHERE table_name='dependent'
+----
+[target]:
+table_name	message	sql_state	file_name	stage
+tag: SELECT 0
+
+fetch useCopy expect-error table-filter=dependent
+----
+ERROR: insert on table "dependent" violates foreign key constraint "dependent_fid_fkey" (SQLSTATE 23503)
```
ryanluu12345, line 83: It seems that your comment on line 68 contracts what we see here. We see the error message for the `dependent` table, even though you call out that `dependent` will terminate immediately before the error is hit. Is this a bug?

fetch/testdata/pg/continuation-edge-cases.ddt
ryanluu12345: Yeah I don't think so, let's just track it in a post GA ticket and dig in later.

verify/tableverify/tableverify.go
ryanluu12345: Ok got it, I expanded the diff and saw it. Thanks for clarifying. Makes sense.

ryanluu12345 (commented):
Generally looks good to me, but I had that one question about the discrepancy between your test comment and the output

fetch/import_table.go
ZhouXing19: No, it will make `_, err := conn.Exec(ctx, importQuery)` returns an error `context canceled`.

fetch/testdata/pg/constraint-exceptions.ddt
ZhouXing19: The difference between the fetch I commented on line 68 and this one is that 
- The former one fetch *both* tables `dependency` and `dependent`. 
- The latter one fetch *only* table `dependent`.

In the former attempt, the goroutine for `dependent` is interrupted early *because* `dependency` fails at the truncate stage. 
In the latter attempt, since we only fetch `dependent`, its goroutine is not interrupted, thus it proceeds till it hits its own failure, which is at the import stage. 

Does it make sense?

fetch/import_table.go
ryanluu12345: What is the implication of not checking the context cancellation here? How likely is this?

ZhouXing19: For example, we're fetching 2 tables in parallel. Table a and b both entered the import stage. 
Table a failed the import, canceling the shared context. Table b will error here, but it will continue this bisect here, trying to narrow down which is the culprit file. But the import failed not for corrupted file, but because of table a. 
So table b won't immediate exit until it finishes the bisect.

Jeremyyang920: Is the ctx passed to importTable the one that is created from the intial waitgroup?

Jeremyyang920:
Im thinking about some parallel cases here and it might require us to sit down a think more, but if we have 2 tables both in import phase and since it shares the same context all the way, our continuation token error message for table 2 will be kinda whack if it just says context cancelled.

fetch/import_table.go
ZhouXing19: Yes, it will be `tableConcurrentCtx` created for each table.

ZhouXing19:
> Im thinking about some parallel cases here and it might require us to sit down a think more, but if we have 2 tables both in import phase and since it shares the same context all the way, our continuation token error message for table 2 will be kinda whack if it just says context cancelled.

Yes, that's the expected behavior. Unfortunately I can't think of a handy way to distinguish that table 2's ctx is canceled due to table 1.

Jeremyyang920:
> > Im thinking about some parallel cases here and it might require us to sit down a think more, but if we have 2 tables both in import phase and since it shares the same context all the way, our continuation token error message for table 2 will be kinda whack if it just says context cancelled.
> 
> Yes, that's the expected behavior with current code change. Unfortunately I can't think of a handy way to distinguish that table 2's ctx is canceled due to table 1.

I think it would have to involve each go routine having its own context but we can think through more of expected behaviors once andy is back.

Jeremyyang920 (approved):
Overall LGTM, lets have a discussion with Andy on good UX for this once he is back and im back from pto.

ZhouXing19:
@Jeremyyang920 @ryanluu12345 Could you give another look at the test `fetch/testdata/pg/continuation.ddt` that I modified? I had to do some weird tweak to make sure it still works.

fetch/testdata/pg/constraint-exceptions.ddt
```diff
@@ -63,10 +63,25 @@ INSERT INTO dependent VALUES (1, 0, 'aaa')
 ----
 [source] INSERT 0 1
 
+# We now run fetch on both `dependent` and `dependency` tables *concurrently*. The fetch attempt for
```
ryanluu12345, line 66: Test now looks good!

ryanluu12345 (approved):
Test output is now fixed for continuation cases. The tests now look deterministic. Awesome!

PR #246, opened 2024-05-13
e2e: hide verbose usage info for error message

ZhouXing19:
Release note: None


ZhouXing19:
@ryanluu12345 This PR might help with the unhide --non-interactive one

ryanluu12345 (approved):

e2e/datadriven_test.go
```diff
@@ -139,3 +139,8 @@ func redactLogs(s string) string {
 	const fileServerStartingPattern = `(?m)^[^\n]*starting file server[^\n]*\n?`
 	return regexp.MustCompile(fileServerStartingPattern).ReplaceAllString(res, "")
 }
+
+func extractErrorOnlyFromMoltRes(s string) string {
```
ryanluu12345, line 143: I think this is a change that will set us up for easier maintenance of the e2e tests, especially when we change flags. Thanks!

e2e/datadriven_test.go
```diff
@@ -139,3 +139,8 @@ func redactLogs(s string) string {
...
+	const errPattern = `Error: [^\n]+\n`
```
ryanluu12345, line 144: What happens in the case where there are multi-line errors? A case I can think of is if the underlying error has multiple lines (`\n` is a valid substring inside). I think it's worthwhile to add a test for the case where that happens.

ryanluu12345:
> @ryanluu12345 This PR might help with the unhide --non-interactive one

Ahh just read this part. You may need to rebase since I merged that `--non-interactive` change already

PR #247, opened 2024-05-13
verify: added tags for each type of mismatch

ryanluu12345:
Added tags for each type of mismatch so that it's easier to filter on mismatch types that lead to error versus ones that do not. Adds logic to split up cases where the not null constraint is on the target versus the source.

Resolves: CC-27398
Release Note: None


verify/tableverify/tableverify.go
```diff
@@ -107,6 +109,11 @@ func verifyTable(
 		delete(truthMappedCols, targetCol.Name)
 
 		if sourceCol.NotNull != targetCol.NotNull {
+			tag := inconsistency.MismatchTargetNotNull
+			if sourceCol.NotNull {
+				tag = inconsistency.MismatchSourceNotNull
```
ZhouXing19, line 114: Hmm, I thought source notnull should be fine?

ryanluu12345:
The intention of this change is to set up the tagging, so that in my follow up PR, I can write helpers that can easily determine if a mismatch warrants a failable mismatch versus ones that don't.

```
func IsFailableMismatch(mismatch MismatchDef) {
     return mismatch == MismatchPK || mismatch == MismatchTargetNotNull .....
}
```

verify/tableverify/tableverify.go
ryanluu12345: Source not null should still be reported. But it won't be used in my logic to determine if it's failable.

ryanluu12345: Actually let me write out that logic rn.

ryanluu12345: Updated PR with that logic for helpers.

ryanluu12345: I guess the real question is: do we also want to just report a mismatch only when the target is stricter than the source? So in this case, only the target has not null.

ZhouXing19: I think we can settle with only reporting `target stricter than source` case for now. It's because if we consider the other way around as well, I think there will be a lot more cases and may lead to confusing logging. However, source stricter than target is essentially "benign" for fetch as it won't introduce additional data inconsistency. Finally, we already have the query to collection "dropped constraint" when creating schema, which is meant to notify the user to add back the constraints when they *finished* molt fetch, and we can easily leverage that for the user-customized schema as well.

ryanluu12345: Sound reasoning. I think we'll go with this, since it actually does lead to a mismatch that we should action on. I think the question is: since this also affects verify, should we report this mismatch regardless of which side it occurs on?

I went ahead and just changed the logic to do the target not null logic for now.

ryanluu12345: Spoke offline, and we decided we should report this mismatch because it allows verify to let the user know about the mismatch. Fetch decides what it wants to do with the mismatch, however.

ZhouXing19 (approved):

PR #248, opened 2024-05-14
fetch: use txn-bound connection for mysql export 

ZhouXing19:
Since mysql doesn't support time traveling, we have to use one single connection, start a txn, and run all the export within this txn. Concurrency is disallowed for mysql source, meaning we will have to keep this connection open for all tables until they all finished export.

Release note: None


fetch/dataexport/mysql.go
```diff
@@ -26,13 +25,20 @@ type mysqlSource struct {
...
 
 func NewMySQLSource(
 	ctx context.Context, settings Settings, conn *dbconn.MySQLConn,
 ) (*mysqlSource, error) {
 	var source string
 	var start, end int
+	_, err := conn.ExecContext(ctx, "START TRANSACTION WITH CONSISTENT SNAPSHOT")
+	if err != nil {
+		return nil, errors.WithHint(errors.Wrap(err, "failed starting transaction with consistent snapshot"),
+			"is your mysql source based on InnoDB engine? "+
```
Jeremyyang920, line 39:
```suggestion
			"Does your MySQL database support the InnoDB engine? "+
```

fetch/testdata/mysql/basic.ddt
```diff
@@ -20,13 +20,35 @@ INSERT INTO tbl2 VALUES (11, 'aaa'), (22, 'bb b'), (33, 'ééé'), (44, '🫡
...
+id	t
+tag: SELECT 0
+
 fetch
 ----
 
 query all
 SELECT * FROM tbl1
 ----
-[source] 0 rows affected
```
ryanluu12345, line 29: Why is this output different now after fetch?

fetch/testdata/mysql/exceptions.ddt
```diff
@@ -13,7 +13,14 @@ INSERT INTO tbl1 VALUES (11, 'aaa'), (22, 'bb b'), (33, 'ééé'), (44, '🫡
 query all
 SELECT * FROM tbl1
 ----
-[source] 0 rows affected
```
ryanluu12345, line 16: Same question for why this is now different?

fetch/dataexport/mysql.go
```diff
@@ -26,13 +25,20 @@ type mysqlSource struct {
...
 	conn     dbconn.Conn
+	txnBound bool
 }
 
 func NewMySQLSource(
 	ctx context.Context, settings Settings, conn *dbconn.MySQLConn,
 ) (*mysqlSource, error) {
 	var source string
 	var start, end int
+	_, err := conn.ExecContext(ctx, "START TRANSACTION WITH CONSISTENT SNAPSHOT")
```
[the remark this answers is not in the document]
ryanluu12345: @Jeremyyang920 @ZhouXing19 I think we should consider supporting scale testing for MySQL: https://cockroachlabs.atlassian.net/browse/CC-28391

fetch/dataexport/mysql.go
```diff
@@ -98,8 +100,10 @@ func (m *mysqlConn) Export(
 	})
 }
 
+// For mysql, we never close the individual connections. We close the mysqlSrc when we finish the
+// source connection pool (mysqlSrc) only.
 func (m *mysqlConn) Close(ctx context.Context) error {
```
ryanluu12345, line 105: Why not make this call the `mysqlSource` close then? Or does this not have access to it overall?

fetch/export_table.go
```diff
@@ -87,15 +87,17 @@ func exportTable(
 			return nil
 		}()
 
-		if exportErr != nil {
-			exportErr = errors.Wrapf(exportErr, "error from exporting stage")
-		}
+		closeConnErr := func() error {
```
ryanluu12345, line 90: Ok I follow the logic here. The  `.Close` from above is a no-op, so that we only close after it's finished.

testutils/conn.go
```diff
@@ -182,17 +182,33 @@ func QueryConnCommand(t *testing.T, d *datadriven.TestData, conns dbconn.Ordered
...
+			rows, err := conn.QueryContext(ctx, d.Input)
 			if err != nil {
 				sb.WriteString(fmt.Sprintf("[%s] error: %s\n", conn.ID(), err.Error()))
 				continue
 			}
-			r, err := tag.RowsAffected()
-			if err != nil {
-				sb.WriteString(fmt.Sprintf("[%s] error getting rows affected: %s\n", conn.ID(), err.Error()))
-				continue
+			sb.WriteString(fmt.Sprintf("[%s]:\n", conn.ID()))
```
ryanluu12345, line 190: Could you please add a comment on what's the significance of this specific string format? Like why it needs to be formatted as such.

fetch/dataexport/mysql.go
```diff
@@ -55,31 +62,26 @@ func (m *mysqlSource) CDCCursor() string {
 }
 
 func (m *mysqlSource) Close(ctx context.Context) error {
+	if m.txnBound {
+		conn := m.conn.(*dbconn.MySQLConn)
+		_, err := conn.Exec("COMMIT")
```
ryanluu12345, line 67: Per the retries conversation we talked about today, it's probably worth retrying this commit statement. In the case there is a network blip or unavailability of the database temporarily during this step, we could leave a long running transaction uncommitted on the source

testutils/conn.go
```diff
@@ -182,17 +182,33 @@ func QueryConnCommand(t *testing.T, d *datadriven.TestData, conns dbconn.Ordered
...
+			typs, err := rows.ColumnTypes()
+			require.NoError(t, err)
+			// Since we are passing in "valPtrs", golang doesn't like if we use
```
ryanluu12345, line 193: Can you please explain a bit more what this below block of code does? Not clear from an initial read. A comment would suffice

ryanluu12345 (approved):
General approach LGTM. Just had some requests to add comments and also explain a bit of the testing discrepancy I'm noticing in the DDTs.

fetch/dataexport/mysql.go
ZhouXing19: We call this function after we finish export for ONE table. We don't close the connection (or sqlSrc) for MySQL at this point as we may still need to reuse it if there's any other table to be exported.

fetch/testdata/mysql/basic.ddt
ZhouXing19: It's because previously for `query` in datadriven test in mysql, we were using `exec`, which is wrong. In other words, we never print the query results for a mysql query.

testutils/conn.go
ZhouXing19: Added comment inline. For reference, I'm replicating the logic that Oliver had in `ScanRowDynamicTypes()`

fetch/dataexport/mysql.go
ZhouXing19: Done.

testutils/conn.go
ZhouXing19: Previously it was calling Exec and was just wrong.

Jeremyyang920 (approved):
LGTM!

e2e/testdata/mysql/bool.ddt
```diff
@@ -111,6 +112,7 @@ fetch --source 'mysql://user:password@0.0.0.0:3306/defaultdb' --target 'postgres
 {"level":"info","table":"public.bool_table","type":"summary","net_duration_ms":1000,"net_duration":"000h 00m 01s","import_duration_ms":1000,"import_duration":"000h 00m 01s","export_duration_ms":1000,"export_duration":"000h 00m 01s","num_rows":4,"cdc_cursor":"0/19E3610","message":"data import on target for table complete"}
 {"level":"info","table":"public.bool_table","message":"cleaning up resources created during fetch run"}
 {"level":"info","type":"summary","fetch_id":"0000000000","num_tables":1,"tables":["public.bool_table"],"cdc_cursor":"0/19E3610","net_duration_ms":1000,"net_duration":"000h 00m 01s","message":"fetch complete"}
```
ryanluu12345, line 114: Do we have testing where we do multiple select queries and then close it after the batch of select queries?

e2e/testdata/mysql/bool.ddt
```diff
@@ -111,6 +112,7 @@ fetch --source 'mysql://user:password@0.0.0.0:3306/defaultdb' --target 'postgres
...
+{"level":"info","message":"closed export source"}
```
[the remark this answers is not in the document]
ZhouXing19: yeah

fetch/dataexport/mysql.go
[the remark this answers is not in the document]
ryanluu12345: Yep I think that that will be an important documentation point. I wonder if it's something we should log out too. But makes sense, as long as customers can set the connection max lifetime to how long they expect to migrate for * safety factor, we should be gtg

fetch/dataexport/mysql.go
```diff
@@ -55,31 +63,41 @@ func (m *mysqlSource) CDCCursor() string {
 }
 
 func (m *mysqlSource) Close(ctx context.Context) error {
+	if m.txnBound {
```
ryanluu12345, line 66: Great, thanks so much for adding the retry. This will be great!

fetch/dataexport/mysql.go
ryanluu12345: Yep, got it. Thanks for clarifying

fetch/testdata/mysql/basic.ddt
ryanluu12345: Ok talked offline. So for broader vis, `query` in the ddt mapped to `exec` in the Go code, which was wrong. @ZhouXing19 fixed this.

testutils/conn.go
ryanluu12345: Would be worthwhile to also note that we use this logic in `ScanRowDynamicTypes` as well. But makes sense to me with the comment.

testutils/conn.go
ryanluu12345: Cool, thanks

ryanluu12345 (approved):
LGTM, let's ship it. I'll test the latest via my scale testing

e2e/testdata/mysql/bool.ddt
ZhouXing19: We have gated table-concurrency and export-concurrency = 1 so I think we should be good.

PR #249, opened 2024-05-14
fetch: fix life cycle for history retention job

ZhouXing19:
**NB: no new tests are added in this PR.** It's because the history retention job is extended in the background, while the duration of the whole fetch process varies based on lots of factors, thus it's hard to get deterministic output. 

---- 

This PR involves couples of fixes. 

1. Bug fix for history retention job lifecycle. 
Previously, we cancel the job when we finish exporting a table. This is wrong when there are multiple tables to be moved -- we need to keep the timestamp will ALL tables finished export. 

2. Added more detailed info to logging

3. Added retry logic for history retention.


fetch/fetch.go
```diff
@@ -452,6 +528,9 @@ func Fetch(
 	go func() {
 		defer close(workCh)
 		for _, table := range tables {
+			if table.Table != "tbl1" {
+				time.Sleep(500 * time.Millisecond)
+			}
```
Jeremyyang920, line 533: Hmm should we make this an actual testing knob? what if someone really does have a table named tb1. Not that its a big deal since its 500ms.

ZhouXing19: Oh yeah it's draft code I forgot to remove lol

cmd/fetch/fetch.go
```diff
@@ -176,7 +176,7 @@ func Command() *cobra.Command {
 						}
 					}()
 
-					const minProtectedTimestampExtensionDur = 10 * time.Second
+					const minProtectedTimestampExtensionDur = 1 * time.Second
```
ryanluu12345, line 179: What is the rationale for changing to a smaller min extension duration?

e2e/testdata/crdb-multi-region/schema_creation.ddt
```diff
@@ -78,18 +78,18 @@ fetch --source 'postgres://root@localhost:26257/defaultdb' --target 'postgres://
...
-{"level":"info","message":"cleaning up resources created during fetch run"}
-{"level":"info","message":"data extraction phase starting"}
-{"level":"info","type":"summary","num_rows":3,"export_duration_ms":1000,"export_duration":"000h 00m 01s","message":"data extraction from source complete"}
-{"level":"info","message":"starting data import on target"}
-{"level":"info","message":"imported 3 rows for batch for files 1 to 3"}
-{"level":"info","type":"summary","net_duration_ms":1000,"net_duration":"000h 00m 01s","import_duration_ms":1000,"import_duration":"000h 00m 01s","export_duration_ms":1000,"export_duration":"000h 00m 01s","num_rows":3,"cdc_cursor":"0/19E3610","message":"data import on target for table complete"}
-{"level":"info","message":"cleaning up resources created during fetch run"}
+{"level":"info","table-concurrency-idx":"-1","table":"newc.newtable","message":"data extraction phase starting"}
```
ryanluu12345, line 81: How do you envision this `table-concurrency-idx` to be used for debugging?

fetch/csv_pipe.go
```diff
@@ -81,7 +81,7 @@ func (p *csvPipe) Pipe(t dbtable.VerifiedTable) error {
 				Str("table", tn.SafeString()).
 				Int("num_rows", p.numRows).
 				Int("shard", p.shardNum).
-				Msgf("row import status")
+				Msgf("row write to csv pipe status")
```
ryanluu12345, line 84: Thanks for fixing this. This is one of the most confusing log lines in our codebase

fetch/dataexport/crdb.go
```diff
@@ -159,18 +167,22 @@ func (c *crdbSource) cancelRetentionJob(
 	conn := cn.(*dbconn.PGConn)
 
 	if _, err := conn.Exec(ctx, `CANCEL JOB $1`, c.hrJobID); err != nil {
-		msg := "failed extending protection interval"
+		msg := "failed canceling protection interval"
```
ryanluu12345, line 170: Super nit: `failed to cancel protection interval`

e2e/testdata/crdb-multi-region/schema_creation.ddt
ZhouXing19: hmm, indeed it might not be very useful. Removing.

cmd/fetch/fetch.go
ZhouXing19: Ah draft code again, reverting.

fetch/fetch.go
```diff
@@ -401,9 +412,74 @@ func Fetch(
...
 	}
 
+	var jobManagementWg *errgroup.Group
+	exportFinishedChan := make(chan error)
+	if hrJobID := sqlSrc.HistoryRetentionJobID(); hrJobID != "" {
+		// Check if all table finished extension every 5 seconds. If finished, cancel job.
+		go func() {
+			hrJobID := hrJobID
+			hrLogger := logger.With().Str("hr_job_export_checker", utils.MaybeFormatHistoryRetentionJobID(cfg.TestOnly, hrJobID)).Logger()
+			ticker := time.NewTicker(5 * time.Second)
```
ryanluu12345, line 422: Should `5 * time.Second` be refactored into a variable since this is defined here and checked later in the code block.

fetch/fetch.go
```diff
@@ -432,10 +506,10 @@ func Fetch(
 				// 2. When the fetch ID is passed in and exception log is not nil, which means it is a table we want to continue from.
 				// This means we want to skip if we are trying to continue but there is no entry that specifies where to continue from.
 				if (cfg.FetchID != "" && relevantExceptionLog != nil) || (cfg.FetchID == "") {
-					madePastExport, err := fetchTable(ctx, cfg, logger, conns, blobStore, sqlSrc, table, tableShards, relevantExceptionLog, isClearContinuationTokenMode, testingKnobs)
-					tableToExportStatus.Store(table.SafeString(), madePastExport)
-					if err != nil {
-						return err
+					// We pass the tableToExportStatus to fetchTable so that we can check if all
```
ryanluu12345, line 509: Ahh, this is the comment I was looking for. So instead of returning the export status now, we just update the sync map concurrently right? And that's why when we check if export finished, we can just see if it finished for all tables. That makes more sense.

Does the output of the test that verifies this show that we log out the continuation token only when we expect? If there is not test, wondering if you can manually check by forcing an error after export completed and also before export completed to see if the logging occurs.

ZhouXing19: > Does the output of the test that verifies this show that we log out the continuation token only when we expect? If there is not test, wondering if you can manually check by forcing an error after export completed and also before export completed to see if the logging occurs.

Yeah, I still saw logs like 
```
{"level":"info","time":"2024-05-15T15:04:30-04:00","message":"table public.new_order failed export, omitting continuation token"}
```
when fetching the tpcc db from drt cluster. (though I hoped it would pass and never have to see it ><)

ZhouXing19: See the last commit

e2e/testdata/pg/continous_token.ddt
```diff
@@ -0,0 +1,44 @@
+exec silent
```
ryanluu12345, line 1: Nit on the file name: `continuation_token.ddt`

e2e/testdata/pg/continous_token.ddt
```diff
@@ -0,0 +1,44 @@
...
+{"level":"info","table":"public.tbl1","type":"summary","num_rows":100000,"export_duration_ms":1000,"export_duration":"000h 00m 01s","message":"data extraction from source complete"}
+{"level":"info","table":"public.tbl1","message":"starting data import on target"}
+{"level":"info","table":"public.tbl2","message":"imported 1 rows for batch for files 1 to 1"}
+{"level":"info","table":"public.tbl2","type":"summary","net_duration_ms":1000,"net_duration":"000h 00m 01s","import_duration_ms":1000,"import_duration":"000h 00m 01s","export_duration_ms":1000,"export_duration":"000h 00m 01s","num_rows":1,"cdc_cursor":"0/19E3610","message":"data import on target for table complete"}
+{"level":"info","table":"public.tbl1","table":"public.tbl1","file_name":"part_00000001.tar.gz","message":"creating or updating token for addsstable [/Table/104/1/1/0,/Table/104/1/24999/0/NULL): checking for key collisions: ingested key collides with an existing one: /Table/104/1/1/0; "}
+{"level":"info","table":"public.tbl1","table":"public.tbl1","continuation_token":"711438b9-ff46-4e29-a312-02a1e091938d","message":"created continuation token"}
```
ryanluu12345, line 42: Hmm this is not deterministic right? Should this be maybe format continuation token? So kind of like how you make `fetch_id` === `00000000`.

fetch/export_table.go
```diff
@@ -101,7 +119,7 @@ func exportTable(
 			if err := func() error {
 				resource, err := datasource.CreateFromReader(ctx, fRW, table, itNum, importFileExt, numRowsCh, testingKnobs, shard.ShardNum)
 				if err != nil {
-					return err
+					return errors.Wrapf(err, "error creating CSVs from the data export process")
```
ryanluu12345, line 122: Nit: it's not always CSVs since it can also be compressed CSVs.

fetch/fetch.go
ryanluu12345: Perfect. Left a comment about making the continuation token deterministic. You likely need a `MaybeFormat` util there.

ryanluu12345 (approved):
Overall lgtm, just a few nits. Good work on this!

PR #250, opened 2024-05-15
verify: updated the getPrimaryKey so that it retrieves the sort order of PKs

ryanluu12345:
In order to improve performance for the sharding, we want to avoid full table scans by considering the order of the primary keys. Add helper logic to extract details about the primary key sort order. Update helpers for the get shards to use this information to build a query with the proper sort order.

Resolves: CC-27398
Release Note: None

**Local Testing**
Logged out the exact queries that are being run:
```
is min  true SELECT o_w_id, o_c_id, o_d_id, o_id, o_carrier_id, o_all_local FROM public.orders ORDER BY o_w_id ASC, o_c_id DESC, o_d_id ASC, o_id DESC, o_carrier_id DESC, o_all_local DESC LIMIT 1
is min  false SELECT o_w_id, o_c_id, o_d_id, o_id, o_carrier_id, o_all_local FROM public.orders ORDER BY o_w_id DESC, o_c_id ASC, o_d_id DESC, o_id ASC, o_carrier_id ASC, o_all_local ASC LIMIT 1
```

Verified that the explain verbose doesn't show FULL for table scan anymore:
```
root@localhost:26257/molt> EXPLAIN (VERBOSE) SELECT o_w_id, o_c_id, o_d_id, o_id, o_carrier_id, 
                        -> o_all_local FROM public.orders ORDER BY o_w_id ASC, o_c_id DESC,     
                        -> o_d_id ASC, o_id DESC, o_carrier_id DESC, o_all_local DESC LIMIT 1;  
                                    info
----------------------------------------------------------------------------
  distribution: local
  vectorized: true

  • scan
    columns: (o_w_id, o_c_id, o_d_id, o_id, o_carrier_id, o_all_local)
    estimated row count: 1 (20% of the table; stats collected 6 hours ago)
    table: orders@order_pkey
    spans: LIMITED SCAN
    limit: 1
(9 rows)


```

TODO

- [x] Write test for get primary key and cols
- [x] Write test for query that is constructed based on pk cols


verify/shard_table.go
```diff
@@ -195,16 +197,34 @@ func buildSelectForSplitPG(tbl tableverify.Result, isMin bool) *tree.Select {
...
 		orderClause := &tree.Order{Expr: tree.NewUnresolvedName(string(pkCol))}
-		if !isMin {
+
+		if orderStr == "ASC" {
+			orderClause.Direction = tree.Ascending
+		} else {
 			orderClause.Direction = tree.Descending
 		}
+
+		// Flip the sort order if we want the maximum (this is "reverse order")
```
ryanluu12345, line 210: @ZhouXing19 , open question here. I did an `EXPLAIN(VERBOSE)` and it looked like it worked fine. I was able to flip the order of each PK field and it avoided the full table scan.

verify/tableverify/query.go
```diff
@@ -169,32 +170,42 @@ func getColumnsForTables(
...
+	var pkSortOrder []string
 
 	switch conn := conn.(type) {
 	case *dbconn.PGConn:
+		// TODO: figure out the query for postgres
+		// TODO: figure out the query for cockroach
 		rows, err := conn.Query(
 			ctx,
 			`
 SELECT
```
ryanluu12345, line 203: This query is still incorrect. The example that I saw was the following:
```
  orders     | CREATE TABLE public.orders (
             |     o_id INT8 NOT NULL,
             |     o_d_id INT8 NOT NULL,
             |     o_w_id INT8 NOT NULL,
             |     o_c_id INT8 NULL,
             |     o_entry_d TIMESTAMP NULL,
             |     o_carrier_id INT8 NULL,
             |     o_ol_cnt INT8 NULL,
             |     o_all_local INT8 NULL,
             |     CONSTRAINT "primary" PRIMARY KEY (o_w_id ASC, o_d_id ASC, o_id ASC)
             | )
```
This led to the following representation (indexes map 1 to 1):
```
[o_w_id o_d_id o_id]
[ASC DESC DESC]
```
Technically, this should all be ascending. There is something wrong with our bit manipulation to get the ASC or DESC order.

verify/tableverify/query.go
```diff
@@ -211,21 +222,25 @@ ORDER BY
...
 			ret = append(ret, c)
+			pkSortOrder = append(pkSortOrder, s)
 		}
 		if rows.Err() != nil {
-			return ret, errors.Wrap(err, "error collecting primary key")
+			return ret, pkSortOrder, errors.Wrap(err, "error collecting primary key")
 		}
 		rows.Close()
 	case *dbconn.MySQLConn:
+		// TODO: figure out the query for MySQL
```
ryanluu12345, line 243: Query for MySQL is finicky too. It shows that it is always `ASC` when in reality we define it as `DESC`. I'm not exactly sure how the INDEX type being BTREE makes a difference here to let us know what the sort order is.

ryanluu12345:
Open questions here:
1. What are your thoughts on this general approach to drill in the sort order information and utilize it?
2. What are ways we can figure out the precise query / queries for the sort order? There's nothing too clear in the docs or forum answers.
3. In the case where we do `drop_on_target_and_recreate`, should we be doing the `verify` before or after? In the case that the table already exists on the target , the PKs might be created in different order the next time. Although, the more I think about it, we may not care since we only want to ensure the data on the source side is in the sorted way we care about.

verify/tableverify/query.go
ryanluu12345: I was able to confirm this when I changed my sort order for PK for `o_id` to ASC on the source as well

verify/tableverify/query.go
```diff
@@ -169,32 +170,49 @@ func getColumnsForTables(
...
 			ctx,
 			`
 SELECT
-    a.attname AS column_name
+    a.attname AS column_name,
+	CASE
+		WHEN (i.indoption[array_position(i.indkey, a.attnum) - 1] & 1) = 1 THEN 'DESC'
+		ELSE 'ASC'
+    END AS sort_order
 FROM
```
ZhouXing19, line 216: nit: would be nice if we can have unit test just for this function. The test can cover various combination of ordering with various number of pk columns. I have tested this query manually with some random schema and it works for all test cases, but still wish we can be more prudent with more unit tests.

ZhouXing19 (commented):
existing code with crdb works for me -- pending discussion on the mysql case.

verify/tableverify/query.go
```diff
@@ -169,32 +170,49 @@ func getColumnsForTables(
...
+		// CRDB implementation: https://github.com/cockroachdb/cockroach/pull/42343/files
+		// PG implementation: https://github.com/postgres/postgres/blob/b6423e92abfadaa1ed9642319872aa1654403cd6/src/include/catalog/pg_index.h#L70-L76
 		rows, err := conn.Query(
 			ctx,
 			`
 SELECT
-    a.attname AS column_name
+    a.attname AS column_name,
+	CASE
+		WHEN (i.indoption[array_position(i.indkey, a.attnum) - 1] & 1) = 1 THEN 'DESC'
```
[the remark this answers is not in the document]
ryanluu12345: Messaged @rafiss offline for a review on this logic. I think he has the best context on `indoption` since he implemented it here: https://github.com/cockroachdb/cockroach/pull/42343/files

verify/tableverify/query.go
ryanluu12345: @ZhouXing19 I completely agree with you. I'll write a test for this. I actually left a comment above that function haha

ryanluu12345: Thoughts on any other cases besides these?
```
// 1. Multiple asc in a row, then desc
// 2. Interleaved asc and desc
// 3. Only asc
// 4. Only desc
```

ZhouXing19: yeah, and ideally mixed ordinal position of the columns, such as 
```
CREATE TABLE t (c1 int, c2 int, c3 int, PRIMARY KEY (c1, c3 ASC, c2 DESC))
```
Here the oridinal positions are `c1, c3, c2`, in contrast to the order of column in table definition `c1, c2, c3`

verify/tableverify/query.go
```diff
@@ -169,32 +170,49 @@ func getColumnsForTables(
 	return ret, nil
 }
 
-func getPrimaryKeysForTables(
+func getPrimaryKeysAndSortOrdersForTables(
 	ctx context.Context, conns dbconn.OrderedConns, tbls [2]dbtable.DBTable,
-) ([2][]tree.Name, error) {
+) ([2][]tree.Name, [2][]string, error) {
```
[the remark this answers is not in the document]
ryanluu12345: @ZhouXing19 , wondering if I could get a review on the latest commit. Latest commit should show this.

verify/tableverify/query.go
ryanluu12345: Noted. I'll add this to my testing later.

verify/tableverify/query.go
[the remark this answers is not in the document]
ryanluu12345: @rafiss reviewed it offline and says it looks good to him. The only thing he asked is if we write test cases on real data here since the `indoption` fields can be tricky. Will do.

verify/tableverify/query.go
ryanluu12345: @Jeremyyang920 , @ZhouXing19 and I discussed offline. I think for now, we are just going to leave this as a todo for MySQL and document.

verify/tableverify/tableverify.go
```diff
@@ -174,34 +174,36 @@ func verifyTable(
 	var collationMismatch bool
 	if currPKSame {
 		for i, col := range targetPKCols {
-			if targetPKCols[i] != truthPKCols[i] {
+			// TODO: figure out if we want to do a stricter check and make
```
ryanluu12345, line 177: @ZhouXing19 , I think for now, probably not. But we may want to configure this in the future right?

I don't think we should block the verify or migration if the sort orders don't match rn.

ZhouXing19: yeah i don't think it should matter. The order for PK iiuc should only matter for performance.

verify/tableverify/query.go
ryanluu12345: Added it in my test cases!

verify/tableverify/tableverify.go
ryanluu12345: Ok sounds good. Took out comment. We'll keep as is unless we want more strict later.

verify/tableverify/query.go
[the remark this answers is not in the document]
ryanluu12345: TFTR!

verify/tableverify/query.go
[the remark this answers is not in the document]
ryanluu12345: All test cases have been written in `query_test` now. Resolving.

verify/tableverify/query.go
ryanluu12345: Left a comment with a ticket link for tracking.

verify/tableverify/query.go
ryanluu12345: Fixed the query.

ZhouXing19 (approved):
Amazing! Thanks for working on this, and also adding the comprehensive tests!

PR #251, opened 2024-05-15
fetch: updated all references of cdc-sink to replicator

ryanluu12345:
cdc-sink recently changed all their artifact names to replicator (the new and improved name). Because of this change, the fetch repository can now replace all references to cdc-sink and use replicator instead.

Resolves: CC-28362
Release Note: CDC sink references are now replaced with references to replicator. This affects users of the ongoing replication feature since they will now need a replicator binary.


fetch/replicator/replicator.go
```diff
@@ -14,19 +14,15 @@ import (
 )
 
 func FindReplicatorBinary(start string) (string, error) {
-	// TODO: Change once cdc-sink binaries are renamed to
-	// replicator. Or we modify our release script to change the
-	// name we bundle in our ZIP.
-	p := filepath.Join(start, "cdc-sink")
-	if _, err := os.Stat(p); err == nil {
+	p := filepath.Join(start, "replicator")
+	if fi, err := os.Stat(p); err == nil && !fi.IsDir() {
```
ryanluu12345, line 18: Ran into a bug where it was trying to run the directory `molt/fetch/replicator` instead of the replicator binary. I want to make sure that it's actually a file before I try running it.

Jeremyyang920 (approved):

PR #253, opened 2024-05-16
logging: default log to a file

ZhouXing19:
Release note: None


cmd/fetch/fetch.go
```diff
@@ -253,8 +255,8 @@ func Command() *cobra.Command {
 	cmd.PersistentFlags().StringVar(
 		&logFile,
 		"log-file",
-		"",
-		"If set, writes to the log file specified. Otherwise, only writes to stdout.",
+		fmt.Sprintf("fetch-%s.log", time.Now().UTC().Format(time.RFC3339)),
+		"Default write it to a file with format \"fetch-{datetime}.log\". If is set with \"stdout\", write to stdout only",
```
Jeremyyang920, line 259: So is the behavior now always print to stdout and log a file by default?

ZhouXing19: yeah, the new default is {stdout + to a file}. Open to revision here

ryanluu12345:
I think this is a broader question, but do we want to add an item to support log rotation? I think the way we do it right now to a single file won't be sustainable especially for **very large migrations**

ZhouXing19:
> I think this is a broader question, but do we want to add an item to support log rotation? I think the way we do it right now to a single file won't be sustainable especially for very large migrations

This is a valid issue. I think we can make it a follow-up. The DB's log is a good example to follow.

cmd/fetch/fetch.go
Jeremyyang920: I think that should be fine. Just like ryan pointed out we might have really large log files but that can be a future thing.

```suggestion
		fmt.Sprintf("fetch-%s.log", time.Now().UTC().Format(time.RFC3339)),
		"By default, writes to a file named \"fetch-{datetime}.log\". If set as \"stdout\", writes will only go to stdout",
```
```suggestion
		fmt.Sprintf("fetch-%s.log", time.Now().UTC().Format(time.RFC3339)),
		"Default write it to a file with format \"fetch-{datetime}.log\". If is set with \"stdout\", write to stdout only",
```

Jeremyyang920: Oh and I think we dont need to escape the quotes here since its just output text.

ZhouXing19:
Issue to track log rotation: https://cockroachlabs.atlassian.net/browse/CC-28385

ZhouXing19:
RFAL, thanks!

fetch/fetch.go
```diff
@@ -479,13 +477,17 @@ func Fetch(
...
+		i := i
 		g.Go(func() error {
 			for {
 				table, ok := <-workCh
 				if !ok {
 					return nil
 				}
 
+				loggerWithConcurrencyIdx := logger.With().Str("table-concurrency-idx", fmt.Sprintf("%d", utils.MaybeFormatIndex(cfg.TestOnly, i))).Logger()
+				loggerWithConcurrencyIdx.Info().Msgf("start fetching table %s", table.SafeString())
```
Jeremyyang920, line 489:
```suggestion
				loggerWithConcurrencyIdx.Info().Msgf("starting fetch for table %s", table.SafeString())
```

moltlogger/logger.go
```diff
@@ -49,15 +50,18 @@ func RegisterLoggerFlags(cmd *cobra.Command) {
...
+func Logger(fileName string, testOnly bool) (zerolog.Logger, error) {
 	var writer io.Writer = os.Stdout
 	if loggerConfigInst.useConsoleWriter {
 		writer = zerolog.NewConsoleWriter(func(w *zerolog.ConsoleWriter) {
 			w.TimeFormat = time.RFC3339
 		})
 	}
 
-	if fileName != "" {
+	if fileName != "stdout" && !testOnly {
```
Jeremyyang920, line 61: We should use strings.equalfolds here for case insensitivity.

Jeremyyang920 (approved):
LGTM with just the 2 small comments.

moltlogger/logger.go
```diff
@@ -49,15 +51,18 @@ func RegisterLoggerFlags(cmd *cobra.Command) {
...
+func Logger(fileName string, testOnly bool) (zerolog.Logger, error) {
 	var writer io.Writer = os.Stdout
 	if loggerConfigInst.useConsoleWriter {
 		writer = zerolog.NewConsoleWriter(func(w *zerolog.ConsoleWriter) {
 			w.TimeFormat = time.RFC3339
 		})
 	}
 
-	if fileName != "" {
+	if strings.EqualFold(fileName, "stdout") && !testOnly {
```
ryanluu12345, line 62: @ZhouXing19 I just noticed this bug while testing on the DRT cluster. Previously, this `fileName` check intends to create a multi writer in the case that there is a file name specified by the user. However, with the logic change here, we don't do anything if the user passes in a file name. This renders the `--log-file` override flag useless. 

I think the correct behavior is to:
1. If stdout is default, then create the file name for them
2. If stdout is not default, and not empty, then just use that filename

So what this logic ends up changing to is:
```
if !testOnly {
   if fileName == "stdout" {
			fileName = fmt.Sprintf("fetch-%s.log", time.Now().UTC().Format(time.RFC3339))
   }
}
```

moltlogger/logger.go
```diff
@@ -49,15 +51,18 @@ func RegisterLoggerFlags(cmd *cobra.Command) {
...
+		if fileName == "" {
```
ryanluu12345, line 63: Also if I'm not mistaken, this code block is unreachable right? The reason I'm saying this is: we only enter this block if fileName == "stdout". So there is no way for `fileName` to be empty haha.

ryanluu12345: I can take care of this fix

moltlogger/logger.go
ryanluu12345: Nevermind, it's even simpler. The fix that we need here is this:
```
if !strings.EqualFold(fileName, "stdout") && !testOnly {
```

ryanluu12345: Fixed here: https://github.com/cockroachdb/molt/pull/257

PR #255, opened 2024-05-16
fetch and verify: added a flag to show or hide connection details logging

ryanluu12345:
Previously, molt could leak connection strings in the logging in the case wrong strings are passed in. However, with the latest change, the default state is more secure and gives a general error letting the user know that connecting to the database failed. In the case the user overrides the flag, then the full connection details can be shown, allowing the user to debug.

Resolves: CC-28264
Release Note: None

**Testing (verified for both fetch and verify)**
Before
```
unable to parse url: postgres://root:%6@localhost:26257/molt?sslmode=disable: parse "postgres://root:%6@localhost:26257/molt?sslmode=disable": invalid URL escape "%6"
exit status 1
```

After (in case of flag default not set)
```
failed to connect to source or target database
```

After (in case of flag enabled and confirmed)
```
Setting the --show-connection-logging flag can potentially log the full connection string. Continue (y/n)?
y
Error: unable to parse url: postgres://root:%6@localhost:26257/molt?sslmode=disable: parse "postgres://root:%6@localhost:26257/molt?sslmode=disable": invalid URL escape "%6"
```

After (in case of flag enabled and not confirmed)
```
Setting the --show-connection-logging flag can potentially log the full connection string. Continue (y/n)?
n
Error: accepting risk of connection string being logged was not confirmed, exiting early
```


fetch/fetch.go
```diff
@@ -577,7 +581,14 @@ func Fetch(
 		sigintChan := make(chan os.Signal, 1)
 		cmdErrChan := make(chan error)
 		go func() {
-			logger.Debug().Str("command", cmd.String()).Msgf("executing replicator command")
+			// Hard to determine where to redact since the user could technically provide
+			// a string that has an invalid password or unescaped values. This can throw
+			// off any reg-exing logic we throw in since we don't know what the password is.
+			// It's safe to just allow logging if the user overrides the flag.
```
yecs1999, line 587: So are we still redacting if the flag is set to false? this comment makes it seem it's difficult to redact this command. Or are we just not showing any logs at all in this case?

yecs1999 (changes requested):

fetch/fetch.go
ryanluu12345: Yeah in this case we do not redact. We just don't show the log unless you previously confirmed with the flag.

ryanluu12345: No way to hit this code path unless the user confirms they're ok with it. Otherwise in default behavior, there is no chance this gets logged.

yecs1999: Okay, thanks for confirming.

yecs1999 (approved):
LGTM, I want to see what the other security folks think though

ZhouXing19 (commented):
LGTM. But wondering if we can add hint to a connection error that they can use this new flag to debug if the connection string they provide is incorrect?

moltlogger/logger.go
```diff
@@ -43,6 +45,13 @@ func RegisterLoggerFlags(cmd *cobra.Command) {
 		loggerConfigInst.useConsoleWriter,
 		"Use the console writer, which has cleaner log output but introduces more latency (defaults to false, which logs as structured JSON).",
 	)
+	cmd.PersistentFlags().BoolVar(
+		&loggerConfigInst.showConnectionLogging,
+		"show-connection-logging",
+		false,
+		`Whether or not connection logging gets shown. False by default to prevent connection string leaks.
```
ryanluu12345, line 52: @taroface mainly wanted your feedback on copy here?

cmd/internal/cmdutil/dbconn.go
```diff
@@ -41,3 +45,25 @@ func LoadDBConns(ctx context.Context) (dbconn.OrderedConns, error) {
 	}
 	return dbconn.OrderedConns{source, target}, nil
 }
+
+func MaybeHandleConnectionLogging(
```
ryanluu12345, line 49: @taroface , could I have your take on the copy here?

cmd/fetch/fetch.go
```diff
@@ -143,10 +143,13 @@ func Command() *cobra.Command {
 				logger.Info().Msgf("user set compression to %s", cfg.Compression.String())
 			}
 
+			// Get the connection logging flag.
+			cfg.ShowConnectionLogging = moltlogger.GetShowConnectionLogging()
 			conns, err := cmdutil.LoadDBConns(ctx)
-			if err != nil {
-				return err
+			if retErr := cmdutil.MaybeHandleConnectionLogging(logger, cfg.ShowConnectionLogging, err); err != nil {
```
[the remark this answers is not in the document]
ryanluu12345: Good catch, yes it should be.

moltlogger/logger.go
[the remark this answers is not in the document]
ryanluu12345: Done!

cmd/fetch/fetch.go
```diff
@@ -143,10 +143,13 @@ func Command() *cobra.Command {
 				logger.Info().Msgf("user set compression to %s", cfg.Compression.String())
 			}
 
+			// Get the connection logging flag.
+			cfg.ShowConnectionLogging = moltlogger.GetShowConnectionLogging()
 			conns, err := cmdutil.LoadDBConns(ctx)
-			if err != nil {
-				return err
+			if retErr := cmdutil.MaybeHandleConnectionLogging(logger, cfg.ShowConnectionLogging, err); retErr != nil {
```
Jeremyyang920, line 149: Is this the only spot we load and check the db conns?

moltlogger/logger.go
```diff
@@ -43,6 +45,14 @@ func RegisterLoggerFlags(cmd *cobra.Command) {
...
```
Jeremyyang920, line 52: Are we hiding this flag?

cmd/internal/cmdutil/dbconn.go
```diff
@@ -41,3 +45,25 @@ func LoadDBConns(ctx context.Context) (dbconn.OrderedConns, error) {
...
+	logger zerolog.Logger, showConnectionLogging bool, originalErr error,
+) error {
+	if showConnectionLogging {
+		fmt.Println(`Setting the --show-connection-logging flag can potentially log the full connection string. Continue (y/n)?`)
+		var confirmation string
+		fmt.Scanln(&confirmation)
+
+		if !strings.EqualFold(confirmation, "y") {
+			return errors.New("accepting risk of connection string being logged was not confirmed, exiting early")
```
taroface, line 58:
```suggestion
			return errors.New("risk of logging connection string was not confirmed; exiting early")
```

taroface (commented):
Some suggestions

cmd/fetch/fetch.go
ryanluu12345: From what I see, yes:
image

It's just these three spots.

moltlogger/logger.go
ryanluu12345: @Jeremyyang920 , I guess it depends on if we want to document this. Should we add this flag to documentation? If not, we can just hide it.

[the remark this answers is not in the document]
ryanluu12345: Shown == output, correct.

cmd/internal/cmdutil/dbconn.go
```diff
@@ -41,3 +45,25 @@ func LoadDBConns(ctx context.Context) (dbconn.OrderedConns, error) {
 	}
 	return dbconn.OrderedConns{source, target}, nil
 }
+
+func MaybeHandleConnectionLogging(
+	logger zerolog.Logger, showConnectionLogging bool, originalErr error,
+) error {
+	if showConnectionLogging {
+		fmt.Println(`Setting the --show-connection-logging flag can potentially log the full connection string. Continue (y/n)?`)
```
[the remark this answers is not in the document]
ryanluu12345: Given that this is not logging but rather a prompt for actions, I will stay consistent with other prompts for actions, which are full sentences with capital first letter.

cmd/internal/cmdutil/dbconn.go
ryanluu12345: Noted.

moltlogger/logger.go
```diff
@@ -43,6 +45,14 @@ func RegisterLoggerFlags(cmd *cobra.Command) {
 		loggerConfigInst.useConsoleWriter,
 		"Use the console writer, which has cleaner log output but introduces more latency (defaults to false, which logs as structured JSON).",
 	)
+	cmd.PersistentFlags().BoolVar(
+		&loggerConfigInst.showConnectionLogging,
+		"show-connection-logging",
+		false,
+		`Whether or not connection logging gets shown. False by default to prevent connection string leaks. 
+Can be set to true to get more detail about connection failures or specific replicator commands, which contain connection string details.
```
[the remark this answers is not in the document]
ryanluu12345: Done.

moltlogger/logger.go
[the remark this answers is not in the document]
ryanluu12345: Done.

ryanluu12345: @taroface , just to be safe, let's hide this flag and not document it. In the case customers request this gets documented, we can do that later on.

taroface (approved):

Jeremyyang920 (approved):

ryanluu12345:
> LGTM. But wondering if we can add hint to a connection error that they can use this new flag to debug if the connection string they provide is incorrect?

I think based on discussion above, we are going to hide this entirely. If a customer runs into issues using this to connect then we can let them know the existence of it, or make it clear to the field teams. We want to provide a secure default upfront. We'll see if this is a common problem as we see usage. In the case that it is, we can add the docs and unhide immediately.

PR #258, opened 2024-05-17
fetch: fix timestamp passed to protect timestamp builtin.

ZhouXing19:
Previously, we pass the timestamp in second precision. This is wrong as it requires a nanosecond precision.

Release note: None


fetch/dataexport/crdb.go
```diff
@@ -58,7 +58,7 @@ func NewCRDBSource(
 		defer func() { newConn.Close(ctx) }()
 		var jobID int64
 		pgConn := newConn.(*dbconn.PGConn)
-		unixTime := ts.Unix()
```
ryanluu12345, line 61: Haha yeah I can see how we fail to protect if we have a much smaller number here than `UnixNano`

ryanluu12345 (approved):

PR #259, opened 2024-05-17
molt-nightly.sh: added a new test case that points to the employees workload in RDS for MySQL

ryanluu12345:
Previously, we did not have MySQL scale testing. We now introduce the foundation for supporting MySQL scale tests. The first iteration points at the employees table which has 200 million rows.

Resolves: CC-28391
Release Note: None

Test Run: https://teamcity.cockroachdb.com/buildConfiguration/Internal_Molt_MoltNightly/15311382?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true


ryanluu12345:
This PR is a good example of the level of effort to add new scale tests. Really simple. The steps are as follows:
1. Create a new database in RDS
2. Put the DB connection string as an env secret in the molt-nightly configuration: https://teamcity.cockroachdb.com/admin/editBuildParams.html?id=buildType:Internal_Molt_MoltNightly
3. Pick a schema and workload and generate data to load in (the most time consuming part)
4. Create a new test case here that references that source and runs with the relevant args and RP config options

Jeremyyang920 (approved):

